PR #31, opened 2022-08-02
fix iframe will take away the focus

DiamondThree:


lawvs:
Hey, i make a simple `LazyIframe` implement, what do you think?

```tsx

const LazyIframe = ({
  src,
  delay = 3000,
  fallback = null,
}: {
  src: string;
  delay?: number;
  fallback?: ReactElement;
}) => {
  const [show, setShow] = useState(false);

  const onLoad = () => {
    setTimeout(() => {
      // Prevent iframe from scrolling parent container
      // Remove the delay after the issue is resolved
      // See W3C https://github.com/w3c/csswg-drafts/issues/7134
      // See https://forum.figma.com/t/prevent-figmas-embed-code-from-automatically-scrolling-to-it-on-page-load/26029/6
      setShow(true);
    }, delay);
  };

  if (!show && fallback) {
    return fallback;
  }

  return (
    <iframe
      src={src}
      style={{ display: show ? "block" : "none" }}
      onLoad={onLoad}
    />
  );
};

// Usage

```

lawvs (approved):

libs/components/editor-blocks/src/components/source-view/SourceView.tsx
```diff
@@ -52,32 +91,76 @@ const SourceViewContainer = styled('div')<{
...
+            <div
+                onMouseDown={e => {
+                    e.preventDefault();
+                    e.stopPropagation();
+                }}
+                style={{ display: show ? 'block' : 'none', height: '100%' }}
+            >
+                <iframe src={src} onLoad={onLoad} />
+            </div>
+        </>
```
lawvs, line 138: It looks like you forgot the `fallback`

PR #82, opened 2022-08-04
fix(component): children line style in whiteboard

Austaras:


libs/components/editor-blocks/src/utils/WithTreeViewChildren.tsx
```diff
@@ -170,93 +150,79 @@ export const withTreeViewChildren = (
...
-    },
-    node: {},
+const Wrapper = styled('div')({ display: 'flex', flexDirection: 'column' });
 
-    children: {
-        display: 'flex',
-        flexDirection: 'column',
-    },
-});
+const Children = Wrapper;
```
lawvs, line 155: The correct usage of the styling component is like this, sorry I didn't explain it clearly.

```tsx
const Wrapper = styled(BlockContainer)({
  display: "flex",
  flexDirection: "column",
});
const Children = styled("div")({ display: "flex", flexDirection: "column" });

// Usage

<BlockContainer
  editor={props.editor}
  block={block}
  selected={isSelect}
></BlockContainer>;


```

See https://emotion.sh/docs/styled#styling-any-component

lawvs (approved):

PR #92, opened 2022-08-04
fix bacakspace in group first block

DiamondThree:


lawvs (approved):

libs/components/editor-blocks/src/blocks/text/TextView.tsx
```diff
@@ -99,7 +99,7 @@ export const TextView: FC<CreateTextView> = ({
             if (!parentBlock) {
                 return false;
             }
-
+            let preParent = await parentBlock.previousSibling();
```
lawvs, line 102: prefer const
```suggestion
            const preParent = await parentBlock.previousSibling();
```

libs/components/editor-blocks/src/blocks/text/TextView.tsx
```diff
@@ -129,34 +129,15 @@ export const TextView: FC<CreateTextView> = ({
...
-
-                    if (!prevGroupBlock) {
-                        const childrenBlock = await parentBlock.children();
-                        if (childrenBlock.length) {
-                            if (children.length) {
-                                await parentBlock.append(...children);
-                            }
                             await block.remove();
-                            return true;
+                            let parentChild = await parentBlock.children();
```
lawvs, line 133: prefer const

```suggestion
                            const parentChild = await parentBlock.children();
```

PR #122, opened 2022-08-06
fix(plugin): left menu item refresh on blockId change

Austaras:


libs/components/editor-plugins/src/menu/left-menu/LeftMenu.tsx
```diff
@@ -18,42 +18,43 @@ interface LeftMenuProps {
...
 
 export function LeftMenu(props: LeftMenuProps) {
-    const { editor, anchorEl, hooks, blockId } = props;
-    const menu: CascaderItemProps[] = [
-        {
-            title: 'Delete',
-            callback: () => {
-                editor.commands.blockCommands.removeBlock(blockId);
+    const { editor, anchorEl, hooks, blockId, onClose } = props;
+    const menu: CascaderItemProps[] = useMemo(
```
lawvs, line 22: Is this `memo` necessary?

lawvs (approved):

libs/components/editor-plugins/src/menu/left-menu/LeftMenu.tsx
Austaras: No, I just tend to memo everywhere

PR #125, opened 2022-08-06
refactor: remove code repetition

CarlosZoft:
## Description

For each production plugin, there was the same check

## Solution

Check only once


darkskygit:
webpack plugins need to be in the correct order, no need to adjust the ordering

CarlosZoft:
I understand, but what do you think of something more broken and less polluted?

CarlosZoft:
my suggestion is to create a base configuration and branch into prod environment or not, i can do it.

darkskygit:
we have several other internal projects using the common webpack configuration. we plan to merge into this repository in the near future, and then we will adjust the configuration of the existing projects together.

PR #164, opened 2022-08-09
feat: new grid drop logic

SaikaSakura:
add new grid drop logic、if has no parent grid create grid others create a grid item

### DIFF
--- a/libs/components/editor-plugins/src/menu/command-menu/Menu.tsx
+++ b/libs/components/editor-plugins/src/menu/command-menu/Menu.tsx
@@ -18,6 +18,7 @@ import {
     menuItemsMap,
 } from './config';
 import { QueryResult } from '../../search';
+import { getBlockIdByNode } from '@toeverything/utils';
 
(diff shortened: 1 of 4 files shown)


lawvs (approved):

libs/components/editor-plugins/src/menu/command-menu/Menu.tsx
```diff
@@ -18,6 +18,7 @@ import {
     menuItemsMap,
 } from './config';
 import { QueryResult } from '../../search';
+import { getBlockIdByNode } from '@toeverything/utils';
```
Austaras, line 21: Doesn't seem to be used

00aaaa11:
T

libs/components/editor-plugins/src/menu/command-menu/Menu.tsx
SaikaSakura: done

Austaras (approved):

PR #178, opened 2022-08-10
fix(group): delete group when there's no children

Austaras:


libs/components/editor-blocks/src/blocks/grid/index.ts
```diff
@@ -36,6 +36,9 @@ export class GridBlock extends BaseView {
             }
             return block.remove();
```
lawvs, line 37: Is this early return is correct?

lawvs (approved):

PR #204, opened 2022-08-12
chore: remove residual `FC` and `FunctionComponent` from components

liby:
## Description

I found that some `FC` were still missing from the https://github.com/toeverything/AFFiNE/commit/40b617c42944f1cc12a11482fa2a5b9283cc3fec and https://github.com/toeverything/AFFiNE/commit/d3b4906da91da50627a0cf3e5bf5665ff6c2cd61 that were not removed. 

What this PR does is find them and remove them, or replace them with other type definitions.

## Other

I updated the `svgo.js` file and tried to run the `nx run components-common:figmaRes` command afterwards, but the SVG component in the `libs/components/common/src/lib/icon` directory did not change at all.


tzhangchi:
> 

Good job! Thank you so much

darkskygit (approved):

lawvs (approved):

PR #218, opened 2022-08-12
fix(whiteboard): error when drag away last element in Chrome

Austaras:


libs/components/editor-core/src/editor/commands/block-commands.ts
```diff
@@ -138,6 +138,7 @@ export class BlockCommands {
         await dropBlock.after(layoutBlock);
         await leftBlock.remove();
         await rightBlock.remove();
+        await sleep(10);
```
lawvs, line 141: Please add some comment

lawvs (approved):

PR #231, opened 2022-08-12
refactor: remove unused React import

liby:
## Motivation

[As of Babel 7.9.0](https://babeljs.io/blog/2020/03/16/7.9.0#a-new-jsx-transform-11154-https-githubcom-babel-babel-pull-11154), when using runtime: automatic in `@babel/preset-react` or `@babel/plugin-transform-react-jsx`, you will not need to explicitly import React for compiling jsx.

This PR removes the redundant import statements. It also converts default imports (`import React from 'react'`) to named imports (e.g. `import { useState } from 'react'`).

[React codemod scripts](https://github.com/reactjs/react-codemod).

### DIFF
--- a/apps/ligo-virgo/src/pages/ui/index.tsx
+++ b/apps/ligo-virgo/src/pages/ui/index.tsx
@@ -1,5 +1,3 @@
-import React from 'react';
-
 export const UIPage = () => {
     return (
         <div className="">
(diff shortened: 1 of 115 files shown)


darkskygit:
some of our internal branches are being revised and ready to be merged into the current repository. They contain a series of code refactor that will conflict with the current pr. We will consider your updates in the future, thank you for your suggestions.
for the code fixed by automatic code generation, you can submit an issue or write a script and save it in the tools folder then submit it as a pr, we will add you as a co-author to the commit after reviewing and executing the script with code changes.

PR #239, opened 2022-08-13
build: experimental vite support

doodlewind:
This PR tries to start the project using vite.

Faster development environment startup allows us to launch a large number of static pages concurrently without packaging, greatly increasing the speed of automated testing in parallel.

Style9 had to be removed here due to the need for an incompatible non-standardized build-time configuration. However, the equivalent standardized code does not actually incur extra maintenance costs. The page style of the Webpack environment on this branch appears to remain normal.

Except for style9, vite requires almost no additional configuration or modification to work ([config](https://github.com/toeverything/AFFiNE/blob/78fcee5da7e7fbfde36d1a1bd5862258dbc73dd7/apps/ligo-virgo/vite.config.ts)).

Currently the cold start speed using vite is about 6s.

Try it out locally:

``` sh
pnpm vite-start:affine
```

AD900820-0395-45AF-98A4-6562EB7D2F01

Known issue: workspace seems to be staying blank due to routing issues, hopefully someone can try based on this branch if interested in further solutions.

### DIFF
--- a/apps/ligo-virgo/index.html
+++ b/apps/ligo-virgo/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
     <head>
-        <!-- local dev index.html -->
+        <!-- Vite local dev index.html -->
         <meta charset="utf-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <link rel="icon" href="/favicon.ico" />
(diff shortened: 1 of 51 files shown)


Austaras:
Emotion needs on Babel too, I don't believe use vite is possible at this point.

Proful:
This is very good effort @doodlewind

darkskygit:
close this pr that because we migration to pathfinder

PR #241, opened 2022-08-13
Code block: Display lang chooser & copy btn on hover

Proful:
Now the language chooser & copy button displayed all the time. It cause unnecessary visual distraction. As these 2 option are not going to used very frequently.

image

Fix: Display language chooser & copy btn on mouse enter (hover)
https://user-images.githubusercontent.com/354596/184478448-3efc5557-003c-45a9-8a64-b856df1b8468.mp4

### DIFF
--- a/libs/components/editor-blocks/src/blocks/code/CodeView.tsx
+++ b/libs/components/editor-blocks/src/blocks/code/CodeView.tsx
@@ -3,7 +3,7 @@ import { StyleWithAtRules } from 'style9';
 
 import { CreateView } from '@toeverything/framework/virgo';
 import CodeMirror, { ReactCodeMirrorRef } from './CodeMirror';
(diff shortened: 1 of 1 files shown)


libs/components/editor-blocks/src/blocks/code/CodeView.tsx
```diff
@@ -144,11 +153,18 @@ export const CodeView = ({ block, editor }: CreateCodeView) => {
...
     useOnSelect(block.id, (_is_select: boolean) => {
         if (codeMirror.current) {
             codeMirror?.current?.view?.focus();
         }
     });
+    const handleMouseEnter = () => {
+        setShowCodeActions(true);
+    };
+    const handleMouseLeave = () => {
+        setShowCodeActions(false);
```
DiamondThree, line 166: I think that using CSS is better。

PR #245, opened 2022-08-14
Edgeless mode > Edit text mode on dblclick

Proful:
Now in edgeless mode it is very difficult to edit text.

This PR helps below:-
- Single Click allow to select text shape & ability to drag
- On double click enter the edit text mode. Now much better experience.

Default ondblclick event is not triggering due to TLDraw hijacking the pointer events. That's why used a dbl click detection logic.

### DIFF
--- a/libs/components/board-shapes/src/editor-util/EditorUtil.tsx
+++ b/libs/components/board-shapes/src/editor-util/EditorUtil.tsx
@@ -114,10 +114,28 @@ export class EditorUtil extends TDShapeUtil<T, E> {
                 }
             }, [onShapeChange]);
 
-            const stopPropagation = useCallback(
+            let numClicks = 0;
+            const handlePointerDown = useCallback(
                 (e: SyntheticEvent) => {
-                    if (isEditing) {
+                    numClicks++;
+                    let isDblClick = false;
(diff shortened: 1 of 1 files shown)


tzhangchi:
> Now in edgeless mode it is very difficult to edit text.
> 
> This PR helps below:-
> 
> * Single Click allow to select text shape & ability to drag
> * On double click enter the edit text mode. Now much better experience.
> 
> Default ondblclick event is not triggering due to TLDraw hijacking the pointer events. That's why used a dbl click detection logic.

It's awesome，thank you so much

Austaras:
Could you try [dblclick](https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event)?

Proful:
@Austaras I have tried [dblclick](https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event) event, it does not work. Events not bubbling up properly.

Austaras:
You could use `MouseDown` event with [detail](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail)

Austaras:
Thank you for this pull request. We'll close this pull request in favour of #287 which should also cover this issue.

PR #261, opened 2022-08-15
fix(component): make Select close when option chosen with Enter key

caleboleary:
When selecting a language in the code block, using arrow keys, enter selects the language but gets the drop down stuck in an open state, no clicking elsewhere can close it, you must pick another (or the same) option

Fixed by explicitly controlling the open state, though left it open for override if need be by putting it before the spread props.

BEFORE:
https://www.loom.com/share/8d1e360e18034468817be165bdb698e2

AFTER:
https://www.loom.com/share/34dd8bb52e7146d780e02a383da5919f

### DIFF
--- a/libs/components/ui/src/select/Select.tsx
+++ b/libs/components/ui/src/select/Select.tsx
@@ -1,5 +1,6 @@
 import {
     forwardRef,
+    useState,
     type CSSProperties,
     type ForwardedRef,
     type ReactNode,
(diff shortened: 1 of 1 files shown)


caleboleary:
Note: I think this has something to do with the component rerendering while the pop is open and losing its ref to the previous version of the drop down, so it's unable to close it. 

Also: <Select> is also consumed in Mention but I'm unsure how to use that, so I was unable to confirm I didn't negatively affect Mention.

QiShaoXuan (approved):

PR #263, opened 2022-08-16
chore: add Star History Chart to README

boojack:
image


lawvs:
Currently the ReadMe is quite long and we are working to refine the content inside. We will consider your update for a future version. Thank you for your suggestion.

PR #265, opened 2022-08-16
Fix/text with root id

lawvs:
- Need to update another text block, for example, todo, list
- Rename db suspend API to `batch`


DiamondThree (approved):

DiamondThree:
Todo and  number list must fix the “enter"  keyboard  event

PR #270, opened 2022-08-16
fix: whiteboard -> edgeless

alt1o:


darkskygit (approved):

darkskygit (approved):

libs/components/editor-blocks/src/components/source-view/format-url/affine.ts
```diff
@@ -1,5 +1,5 @@
 const _regex =
-    /^(https?:\/\/(localhost:4200|(nightly|app)\.affine\.pro|.*?\.ligo-virgo\.pages\.dev)\/\w{28}\/)?(affine\w{16})(\/whiteboard)?$/;
+    /^(https?:\/\/(localhost:4200|(nightly|app)\.affine\.pro|.*?\.ligo-virgo\.pages\.dev)\/\w{28}\/)?(affine\w{16})(\/edgeless)?$/;
```
lawvs, line 2: is the check too strict? @darkskygit

lawvs (approved):

libs/components/editor-blocks/src/components/source-view/format-url/affine.ts
darkskygit: domain can relax to current domain, fix it after the two-way link work correct

PR #272, opened 2022-08-16
fix: add macOS hotkey support for pendant-modify-panel component

chenmoonmo:
when i edit options of a tag by the panel, I found the tooltip  shows ' ctrl + backspcae' . And 'command + backspcae' can not delete a extra option.
The PR make tooltip of closeIcon show '⌘ + backspace' in macOS and respond to it correctly.

### DIFF
--- a/libs/components/editor-core/src/block-pendant/pendant-modify-panel/Select.tsx
+++ b/libs/components/editor-core/src/block-pendant/pendant-modify-panel/Select.tsx
@@ -1,27 +1,28 @@
-import React, {
-    CSSProperties,
-    useEffect,
-    useState,
-    KeyboardEvent,
-    useRef,
-} from 'react';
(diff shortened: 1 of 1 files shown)


mitsuhatu:
Excuse me, the PR can work as expectation in Windows?

QiShaoXuan:
Unfortunately this pr can't be adopted, the command + backspace shortcut combination is already used to quickly clear the input internal text

chenmoonmo:
> Unfortunately this pr can't be adopted, the command + backspace shortcut combination is already used to quickly clear the input internal text
But the ctrl + backspace shortcut is also a shortcut of system to remove last word, and control is not a common key in macOS.
Perhaps command is not a good substitute, but ctrl does not fit the user's operating habits

PR #277, opened 2022-08-17
feat: add theme support to code editor block

m1911star:
Purpose:
- add theme suport and theme select component

Showcase:

https://user-images.githubusercontent.com/4948120/185084326-2d2f1572-3cf0-4d6e-91ff-e4c60f596937.mov

There are some issues I'm not sure:
1. whether the data is saved properly
2. some styles are confusing. since lack of codemirror6 official themes , I use https://github.com/craftzdog/cm6-themes,the bundle size is ok by checking https://bundlephobia.com/package/cm6-theme-solarized-dark@0.2.0
3. I just don't make too much modification to the styles of theme select component.( Just copy the language select component lol~)


DiamondThree:
Thanks for your push request.

At the moment we are working hard on improving compatibility across browsers and devices. And this customization may add some complexity to that. 

This is an interesting suggestion that we may consider and implement at a later stage.

PR #283, opened 2022-08-17
Fix/experience

QiShaoXuan:
1. refactor code of copy
2. fix not being able to paste a block with children in it
3. fix not being able to paste a group in white board to doc


libs/components/editor-core/src/editor/clipboard/utils.ts
```diff
@@ -12,3 +14,38 @@ export const shouldHandlerContinue = (event: Event, editor: Editor) => {
...
+    const clipInfos = await Promise.all(
+        blockIds.map(blockId => getClipInfoOfBlockById(editor, blockId))
+    );
+    return [
+        OFFICE_CLIPBOARD_MIMETYPE.DOCS_DOCUMENT_SLICE_CLIP_WRAPPED,
+        JSON.stringify({
+            data: clipInfos,
+        }),
+    ];
+};
```
lawvs, line 51: Please declare the type explicitly

libs/components/editor-core/src/editor/clipboard/copy.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private async _getClipInfoOfBlock(selBlock: SelectBlock) {
+        const block = await this._editor.getBlockById(selBlock.blockId);
+        const blockView = this._editor.getView(block.type);
+        assert(blockView);
+        const blockInfo: ClipBlockInfo = {
+            type: block.type,
+            properties: blockView.getSelProperties(block, selBlock),
+            children: [] as any[],
```
lawvs, line 91:
```suggestion
            children: [] as ClipBlockInfo[],
```

libs/components/editor-core/src/editor/clipboard/copy.ts
```diff
@@ -0,0 +1,136 @@
...
+                    });
+                } catch (e) {
+                    // TODO handle exception
+                }
+            }
+        }
+    }
+
+    async getClips() {
+        const clips: any[] = [];
```
lawvs, line 50:
```suggestion
        const clips: Clip[] = [];
```

libs/components/editor-core/src/editor/clipboard/copy.ts
```diff
@@ -0,0 +1,102 @@
+import { Editor } from '../editor';
+import { SelectInfo } from '../selection';
+import { OFFICE_CLIPBOARD_MIMETYPE } from './types';
+import { Clip } from './clip';
+import ClipboardParse from './clipboard-parse';
+import { getClipDataOfBlocksById } from './utils';
+
+class Copy {
```
lawvs, line 8: The class name should use a noun that is easy to understand, maybe `CopyHelper` or `CopyOperator` is a good choice?

libs/components/editor-core/src/editor/clipboard/copy.ts
```diff
@@ -0,0 +1,102 @@
...
+        const selectInfo: SelectInfo =
+            await this._editor.selectionManager.getSelectInfo();
+
+        return getClipDataOfBlocksById(
+            this._editor,
+            selectInfo.blocks.map(block => block.blockId)
+        );
+    }
+
+    private _copyToClipboardFromPc(clips: any[]) {
```
lawvs, line 70: we have a utils fun `copyToClipboard`, is it enough?

```ts
import { copyToClipboard } from '@toeverything/utils';

```

libs/components/editor-core/src/editor/clipboard/utils.ts
```diff
@@ -12,3 +14,39 @@ export const shouldHandlerContinue = (event: Event, editor: Editor) => {
...
+    blockId: string
+) => {
+    const block = await editor.getBlockById(blockId);
+    const blockView = editor.getView(block.type);
+    const blockInfo: ClipBlockInfo = {
+        type: block.type,
+        properties: blockView.getSelProperties(block, {}),
+        children: [] as ClipBlockInfo[],
+    };
+    const children = await block?.children();
```
lawvs, line 29:
```suggestion
    const children = await block?.children() ?? [];
```

lawvs (approved):

libs/components/editor-core/src/editor/clipboard/copy.ts
QiShaoXuan: copyToClipboard  not support multi-type custom additions

PR #307, opened 2022-08-21
refactor: refactor clipboard

QiShaoXuan:


libs/components/common/src/lib/text/slate-utils.ts
```diff
@@ -958,7 +959,41 @@ class SlateUtils {
...
+        return Editor.node(this.editor, path);
+    }
+
+    public getNodeByRange(range: Range) {
+        return Editor.node(this.editor, range);
+    }
+
+    // This may should write with logic of render slate
+    public convertLeaf2Html(textValue: any) {
+        const { text, fontColor, fontBgColor } = textValue;
```
lawvs, line 971: Need to escape HTML special chars to prevent security problems

QiShaoXuan: Here is exporting html, escape should be done by the user who is use the clipData

libs/components/editor-core/src/editor/clipboard/clipboardUtils.ts
```diff
@@ -0,0 +1,222 @@
...
+                        selectInfo: selectBlockInfo,
+                    });
+                })
+            )
+        ).join('');
+    }
+
+    async convertHTMLString2Blocks(html: string): Promise<ClipBlockInfo[]> {
+        const htmlEl = document.createElement('html');
+        htmlEl.innerHTML = html;
```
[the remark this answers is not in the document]
QiShaoXuan: HTML string is escaped

lawvs (approved):

PR #321, opened 2022-08-24
Bugfix/board style

DiamondThree:


libs/components/affine-board/src/Board.tsx
```diff
@@ -51,11 +51,10 @@ const AffineBoard = ({
         };
     });
 
-    const { shapes, bindings } = useShapes(workspace, rootBlockId);
-
+    const { shapes } = useShapes(workspace, rootBlockId);
     useEffect(() => {
         if (app) {
-            app.replacePageContent(shapes || {}, bindings, {});
+            app.replacePageContent(shapes || {}, {}, {});
```
alt1o, line 57: Maybe you should remove the save bindings logic

libs/components/affine-board/src/hooks/use-shapes.ts
```diff
@@ -43,11 +43,9 @@ export const useShapes = (workspace: string, rootBlockId: string) => {
                     return shapes;
                 }),
         ]).then(shapes => {
-            getBindings(workspace, rootBlockId).then(bindings => {
-                setBlocks({
-                    shapes,
-                    bindings: bindings,
-                });
+            setBlocks({
```
alt1o, line 46: Maybe you should remove the 'getBindings' implement.

libs/components/board-draw/src/TlDraw.tsx
```diff
@@ -361,11 +361,11 @@ const InnerTldraw = memo(function InnerTldraw({
...
-    const hideBounds =
-        (isInSession && app.session?.constructor.name !== 'BrushSession') ||
-        !isSelecting ||
-        isHideBoundsShape ||
-        !!pageState.editingId;
+    // const hideBounds =
+    //     (isInSession && app.session?.constructor.name !== 'BrushSession') ||
+    //     !isSelecting ||
+    //     isHideBoundsShape ||
+    //     !!pageState.editingId;
```
alt1o, line 368: Should the selected bounds be displayed in all cases? For example, during brush painting

alt1o (approved):

libs/components/board-draw/src/TlDraw.tsx
DiamondThree: NO, i will fix the bug

PR #332, opened 2022-08-25
Feat/board paste

DiamondThree:


libs/components/editor-core/src/editor/editor.ts
```diff
@@ -479,7 +479,22 @@ export class Editor implements Virgo {
...
+    public copyBlock = async (blockId: string, parentId: string) => {
+        let block = await this.getBlockById(blockId);
+        let newBlcok: AsyncBlock;
+        newBlcok = await this.createBlock(block.type, parentId);
+        let parentProperties = await block.getProperties();
+        newBlcok.setProperties(parentProperties);
+        let children = await block.children();
+        if (children.length) {
+            await Promise.all(
+                children.map(async child => {
```
lawvs, line 491: The `map` method has semantics that returns something

```suggestion
                children.forEach(async child => {
```

libs/components/editor-core/src/editor/editor.ts
```diff
@@ -479,7 +479,22 @@ export class Editor implements Virgo {
...
+                    this.copyBlock(child.id, newBlcok.id);
+                })
+            );
```
lawvs, line 494: It seems you forgot to await or return a promise at this block

libs/components/editor-core/src/editor/editor.ts
```diff
@@ -479,7 +479,22 @@ export class Editor implements Virgo {
             return false;
         });
     }
-
+    public copyBlock = async (blockId: string, parentId: string) => {
```
lawvs, line 482: Is the parent id is optional?

QiShaoXuan: This doesn't look like a copy of the block, but rather a new blcock created and inserted into the parent block

PR #338, opened 2022-08-26
Feat toc scroll active

tzhangchi:


apps/ligo-virgo/src/pages/workspace/docs/components/toc/toc.css
```diff
@@ -0,0 +1,13 @@
...
+}
+
+@keyframes blinker {
+    0% {
+        background: rgba(152, 172, 189, 0.1);
+    }
+    100% {
+        background: rgba(152, 172, 189, 0.1);
+    }
+}
```
lawvs, line 13: You can use `styled` like this

```tsx
const Blinker = styled('div')``;

// ---
const className = Blinker.toString();

block.dom.classList.add(className);

```

tzhangchi: I need more details example about keyframes styed usage

lawvs (approved):

PR #345, opened 2022-08-30
fix: can not copy text other than block

QiShaoXuan:


lawvs (approved):
Can you describe the reasons for this?

PR #361, opened 2022-09-02
1.adjust export/import database;

mitsuhatu:


libs/datasource/jwt/src/yjs/index.ts
```diff
@@ -324,53 +324,59 @@ export class YjsAdapter implements AsyncDatabaseAdapter<YjsContentOperation> {
...
+
+                    const update_check = new Promise<void>(resolve => {
+                        const check = async () => {
+                            while (!isUpdated || updated > 0) {
+                                await sleep();
+                            }
+                            resolve();
+                        };
+                        check();
+                    });
```
alt1o, line 367: What scenarios need to listen to update

libs/components/layout/src/header/FileSystem.tsx
```diff
@@ -80,34 +80,20 @@ export const FileSystem = () => {
...
-                >
-                    {t('Sync to Disk')}
-                </StyledFileSystem>
-            </>
+            <MuiSnackbar
+                anchorOrigin={{ vertical: 'top', horizontal: 'right' }}
+                open={error}
+                message="Request File Permission failed, please check if you have permission"
+                sx={{ marginTop: '3em' }}
+                action={
```
alt1o, line 88: What scenarios will fail

libs/datasource/jwt/src/index.ts
```diff
@@ -357,7 +357,7 @@ export class BlockClient<
      * @param name block name
      * @returns block instance
      */
-    private async get_named_block(
+    public async get_named_block(
```
alt1o, line 360: Naming convention

libs/datasource/jwt/src/yjs/index.ts
mitsuhatu: i don't know, only add try.catch logic.
There is a problem with the previous logic: when click cancel btn, will throw error

libs/components/layout/src/header/FileSystem.tsx
mitsuhatu: when click cancel-btn, will show it

mitsuhatu: when click cancel-btn, will show it

PR #430, opened 2022-09-15
fix: can't add image and delete more action button

DiamondThree:


lawvs (approved):
The `setTimeout` may need to optimize

PR #432, opened 2022-09-16
Fix/bugs

QiShaoXuan:


libs/components/editor-core/src/editor/clipboard/utils.ts
```diff
@@ -5,6 +5,34 @@ import { ClipBlockInfo } from './types';
 const getIsLink = (htmlElement: HTMLElement) => {
     return ['A', 'IMG'].includes(htmlElement.tagName);
 };
+export const getIsTextLink = (str: string) => {
+    const regex = new RegExp(
+        /https?:\/\/(www\.)?[-a-zA-Z\d@:%._+~#=]{1,256}\.[a-zA-Z\d()]{1,6}\b([-a-zA-Z\d()@:%_+.~#?&//=]*)/gi
```
lawvs, line 10: Can you add a comment to explain the regex?

lawvs (approved):

PR #485, opened 2022-10-09
add some punctuation

yzr006:


CJSS:
Thank you for your suggestions, they were considered for merged PR #487.

PR #490, opened 2022-10-11
Fix/20221010 bugs

alt1o:
fix: edgeless pencil color not work
https://user-images.githubusercontent.com/21084335/194991759-2505b47a-948b-424b-af45-c9b471cc123a.mov

fix: hide the mode switcher in header when viewport size is smaller than 970px, because of the compatibility. It will be fixed in future.
https://user-images.githubusercontent.com/21084335/194991588-78abbf7b-40e7-4cda-bff0-3c67b39ff177.mov

fix: mixed Chinese and English
image

fix: some warning
https://user-images.githubusercontent.com/21084335/194989841-b5d885aa-321d-41af-86b6-1ff6c132e70c.png
image

### DIFF
--- a/libs/components/ui/src/model/index.tsx
+++ b/libs/components/ui/src/model/index.tsx
@@ -32,11 +32,13 @@ export const TransitionsModal = (props: TransitionsModalProps) => {
             onClose={props.onClose}
             closeAfterTransition
         >
-            <MuiClickAwayListener onClickAway={props.onClose}>
-                <Fade in={props.open} timeout={300}>
-                    {props.children}
-                </Fade>
-            </MuiClickAwayListener>
+            <div>
+                <MuiClickAwayListener onClickAway={props.onClose}>
+                    <Fade in={props.open} timeout={300}>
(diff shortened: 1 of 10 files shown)


libs/components/ui/src/model/index.tsx
```diff
@@ -32,11 +32,13 @@ export const TransitionsModal = (props: TransitionsModalProps) => {
             onClose={props.onClose}
             closeAfterTransition
         >
-            <MuiClickAwayListener onClickAway={props.onClose}>
-                <Fade in={props.open} timeout={300}>
-                    {props.children}
-                </Fade>
-            </MuiClickAwayListener>
+            <div>
```
lawvs, line 35: What is the purpose of the new div?

lawvs (approved):
Can you paste some screenshots about the changes?

libs/components/ui/src/model/index.tsx
alt1o: image

PR #553, opened 2022-12-13
feat: init support for mouse mode

himself65:
Waiting for https://github.com/toeverything/blocksuite/pull/296


himself65:
See https://github.com/toeverything/AFFiNE/pull/557

PR #575, opened 2022-12-20
chore: bump @blocksuite to latest and fix monorepo

himself65:


himself65:
Don’t merge this now. I’ve mistakenly upgraded lit to 2.5.0

packages/app/tsconfig.json
```diff
@@ -17,7 +17,9 @@
     "experimentalDecorators": true,
     "baseUrl": ".",
     "paths": {
-      "@/*": ["src/*"]
+      "@/*": ["src/*"],
+      "@pathfinder/data-services": ["../data-services/src/index"],
+      "@toeverything/pathfinder-logger": ["../logger/src/index"]
```
alt1o, line 22: there shouldn't be config by tsconfig, we depend on the package not the source code. because we have done some resolve modules when deploy to production environment.

alt1o (changes requested):

PR #587, opened 2022-12-26
staged feature merge

darkskygit:
- document folder
- full-text search
- blob storage
- basic edgeless support
- cloud sync support

cc @toeverything/official-contributors add the feature you are responsible for
just for avoid going further down the road of no tests and lint, some features not finish yet


Himself65 (approved):
I read the code. LGTM if CI passes!

PR #642, opened 2022-12-30
Feat/e2e test improvement

tzhangchi:


tests/local-first-export-page.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { test, expect } from '@playwright/test';
 import { loadPage } from './libs/load-page';
-
+import { newPage } from './libs/page-logic';
 loadPage();
 
 test.describe.skip('Local first export page', () => {
```
darkskygit, line 6: need to remove skip flag

darkskygit: not just this file include skip flag

darkskygit (changes requested):

tests/local-first-export-page.spec.ts
tzhangchi: > not just this file include skip flag

Yes, I commit new files again

darkskygit (approved):

PR #649, opened 2023-01-01
test: add e2e test for validation of editor version

tzhangchi:


packages/app/src/pages/_app.tsx
```diff
@@ -34,7 +34,9 @@ type AppPropsWithLayout = AppProps & {
 
 const App = ({ Component, pageProps }: AppPropsWithLayout) => {
   const getLayout = Component.getLayout || (page => page);
-
+  useEffect(() => {
+    printBuildInfo();
+  }, []);
```
himself65: I think the better practice is put printBuildInfo into the head

like 

```html
<head>
  <meta name="affine:version" content="xxx">
</head>
```

```js
var meta = document.createElement('meta');
meta.name = "affine:version";
meta.content = publicRuntimeConfig.EDITOR_VERSION;
document.getElementsByTagName('head')[0].appendChild(meta);
```

himself65 (changes requested):

packages/app/src/pages/_app.tsx
tzhangchi: playwright how to get this value?

himself65: also, to remove the useEffect, because this log is outside of the react lifecycle

```js
// top of the entry point file
if (typeof window !== 'undefined') {
  document.addEventListerner('load', () => {
    // xxx
  })
}
```

tzhangchi: > also to remove the useEffect, because this log is outside of the react lifecircle
> 
> ```js
> // top of the entry point file
> if (typeof window !== 'undefined') {
>   document.addEventListerner('load', () => {
>     // xxx
>   })
> }
> ```

@lawvs  what's your suggestion?

tests/console.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+
+test.describe('web console', () => {
+  test('editor version', async ({ page }) => {
+    // https://playwright.dev/docs/evaluating
+    // https://github.com/microsoft/playwright/issues/13059
+    // Get the handle to a specific function.
+    //Later on, call this function with some arguments.
+    // const msg = await getEditoVersionHandle.evaluate((post, args) => post);
+    // console.log(getEditoVersionHandle);
+    await page.waitForTimeout(500);
```
lawvs, line 15: I think this wait should be placed on the `loadPage`.

tzhangchi: > I think this wait should be placed on the `loadPage`.

I remove the   `page.waitForTimeout(1000)` in all scene  before test,  it is a litte redundant.  it is ok now , waitForTimeout is not necessary  for  every test cases

darkskygit (approved):

himself65 (approved):

PR #691, opened 2023-01-09
chore: change jest to vitest

alt1o:
- @blocksuite support ESM originally. 
- Vitest is ESM first.


doodlewind (approved):

alt1o:
Using Playwright instead.

PR #694, opened 2023-01-09
chore: move i18n ci scripts to scripts

alt1o:
i18n ci scripts should place in the scripts folder. We don't need to lint it in build and publish. @JimmFly will add a new package for all i18n things. 
This pr only makes i18n scripts not block the build progress.


darkskygit (approved):

darkskygit:
cc @JimmFly

PR #696, opened 2023-01-09
feat: extract i18n into a package

JimmFly:


packages/i18n/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "build": "tsc --project ./tsconfig.json"
+  },
+  "keywords": [],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/toeverything/AFFiNE.git"
+  },
+  "dependencies": {
+    "i18next": "^21.9.1",
+    "prettier": "^2.7.1",
```
lawvs, line 22: We have installed `prettier` in the root package

```suggestion
```

packages/i18n/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "react-i18next": "^11.18.4"
+  },
+  "devDependencies": {
+    "@types/prettier": "^2.7.2",
+    "typescript": "^4.8.4"
```
lawvs, line 27:
```suggestion
```

packages/i18n/package.json
JimmFly: This package does depend on Prettier. I don't want it looking outside for this module.

darkskygit (approved):

PR #716, opened 2023-01-12
chore: update BlockSuite version including introducing feature flag, adding upgrade npm script

thorseraq:


package.json
```diff
@@ -19,7 +19,8 @@
     "test:unit": "playwright test --config=playwright.config.unit.ts",
     "postinstall": "husky install",
     "notify": "node --experimental-modules scripts/notify.mjs",
-    "check:ci": "pnpm lint & pnpm test"
+    "check:ci": "pnpm lint & pnpm test",
+    "update:blocksuite": "chmod 755 update-dep.sh &&  sh update-dep.sh"
```
darkskygit, line 23: shell is not available in windows, and it is best to write commands directly in scripts.
you can use `pnpm i -w --filter "@affine/*"` to install dependencies for multiple packages

thorseraq: Is there any mechanism to specify `@nightly` version of certain kind of dependency? In this case, we want to upgrade all all dependencies in `@blocksuite` scope with `@nightly` version, we now have two sub packages, `app`, `data-center`. 

`pnpm i -w --filter "@affine/*"` doesn't specify version, and as not all packages may have `@nightly` tag, we should restrict the scope to `@blocksuite`

thorseraq: Thanks! resolved now

darkskygit (approved):

PR #770, opened 2023-02-02
feat: add `MessageCenterHandler`

QiShaoXuan:


pengx17:
I refactored some error handling in https://github.com/toeverything/AFFiNE/pull/936 as well.

apps/web/src/pages/_app.tsx
```diff
@@ -77,17 +78,20 @@ const App = ({ Component, pageProps }: AppPropsWithLayout) => {
...
-            <Suspense fallback={<PageLoading />}>
-              <DataCenterPreloader>
-                <AppDefender>
-                  {getLayout(<Component {...pageProps} />)}
-                </AppDefender>
-              </DataCenterPreloader>
-            </Suspense>
-          )}
+          <>
+            <MessageCenterHandler />
```
himself65: use children

apps/web/src/components/message-center-handler/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+    const instance = MessageCenter.getInstance();
+    if (instance) {
+      return instance.onMessage(async message => {
+        if (message.code === MessageCenter.messageCode.noPermission) {
+          // todo: translate message
+          // todo: more specific message for accessing different resources
+          // todo: error toast style
+          toast('You have no permission to access this workspace');
+          getApis().auth.clear();
+          // the status of the app right now is unknown, and it won't help if we let
```
himself65, line 22: use error boundary

pengx17: IMO error boundary is only for unexpected exceptions that could crash an app. We could use it to report errors or provide a last-resort error screen to let the user do some actions manually. Handling 401 errors are not in this case.
If we know the reason for the error & know how to deal with it, error boundary should be avoided.

Himself65 (approved):

Issue #854, opened 2023-02-07
Missing valid browser titles in different pages


pengx17:
![Image](https://user-images.githubusercontent.com/584378/217160305-a2114fd2-641f-46e9-b4ce-df4ff7f93ee2.png)

Right now it shows the current URL, but it should show something like "AFFiNE | xxx" etc.
Related routes
- all pages
- settings
- trash

@joebeijing

joebeijing:
Agree with that, also for Favourites page.

PR #883, opened 2023-02-07
fix: add page titles

pengx17:
fix https://github.com/toeverything/AFFiNE/issues/854


JimmFly:
I have modified the test case of quick search here #899 , you can merge the master branch again after it is merged.

darkskygit (approved):

PR #886, opened 2023-02-07
fix: logout will jump to 404 error

DiamondThree:


packages/app/src/components/workspace-modal/index.tsx
```diff
@@ -132,7 +132,7 @@ export const WorkspaceModal = ({ open, onClose }: WorkspaceModalProps) => {
         onClose={async wait => {
           if (!wait) {
             await logout();
-            router.replace(`/workspace`);
```
alt1o, line 135: what is the reason for using 'push' replace 'replace'?

darkskygit (approved):

PR #900, opened 2023-02-08
chore: add delete modules in npm script

thorseraq:


package.json
```diff
@@ -18,7 +18,8 @@
     "test:unit": "playwright test --config=playwright.config.unit.ts",
     "postinstall": "husky install",
     "notify": "node --experimental-modules scripts/notify.mjs",
-    "update:blocksuite": "pnpm i --filter @affine/app --filter @affine/datacenter @blocksuite/blocks@nightly @blocksuite/store@nightly && pnpm i --filter @affine/app @blocksuite/editor@nightly"
+    "update:blocksuite": "pnpm i --filter @affine/app --filter @affine/datacenter @blocksuite/blocks@nightly @blocksuite/store@nightly && pnpm i --filter @affine/app @blocksuite/editor@nightly",
+    "delete:node_modules": "find . -name \"node_modules\" -exec rm -rf '{}' +"
```
himself65: could add a shell script instead of in package.json?

Himself65 (changes requested):

package.json
thorseraq: Make sense, how about changes in https://github.com/toeverything/AFFiNE/pull/900/commits/ff772e6024089a3f91b804ed673d64f0c403e6d3

lawvs (changes requested):
It's should not be a common operation

thorseraq:
> It's should not be a common operation

Are you suggesting close it? Add an tool doesn't mean it's a common operation though

PR #963, opened 2023-02-13
fix: userInfo being initialized correctly

pengx17:
fix https://github.com/toeverything/AFFiNE/issues/964

### DIFF
--- a/apps/web/src/providers/app-state-provider/Provider.tsx
+++ b/apps/web/src/providers/app-state-provider/Provider.tsx
@@ -33,7 +33,9 @@ export const AppStateProvider = ({
         await createDefaultWorkspace(dataCenter);
       }
       setUser(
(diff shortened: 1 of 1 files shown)


apps/web/src/providers/app-state-provider/Provider.tsx
```diff
@@ -33,7 +33,9 @@ export const AppStateProvider = ({
         await createDefaultWorkspace(dataCenter);
       }
       setUser(
-        (await dataCenter.getUserInfo(dataCenter.providers[0]?.id)) || null
+        (await dataCenter.getUserInfo(
+          dataCenter.providers.filter(p => p.id !== 'local')[0]?.id
+        )) || null
```
pengx17, line 38: Right now `dataCenter.providers` field has `local`(`tauri-ipc` in desktop app) and `affine` in order.
However the `local` provider does not have valid user info and shall be excluded.

linonetwo (approved):

PR #998, opened 2023-02-14
refactor: switch to zustand in `useConfirm` hook

Himself65:


apps/web/src/providers/ConfirmProvider.tsx
```diff
@@ -1,62 +1,100 @@
...
+        })
+      )
+    )
+  );
+
+type Store = ReturnType<typeof create>;
+
+export const ConfirmContext = createContext<Store | null>(null);
+
+export const useConfirmApi = () => {
```
doodlewind, line 65: What's the convention (or difference) between `useConfirmApi` and `useConfirm`? Is this a will-be-widely-used pattern?

himself65: `useApi` retunrs 

```
{
  setState,
  getState,
  subscribe
}
```

Which is the vanilla API from javascript

https://github.com/pmndrs/zustand/blob/main/docs/guides/initialize-state-with-props.md#common-patterns

doodlewind (approved):

PR #1005, opened 2023-02-15
chore: assign target for rust-cache

linonetwo:


himself65:
Why are there always "3 cancelled" after pushing the new commits? Can we run them or ignore them because merge have to pass all of the checks

/cc @darkskygit 

image

linonetwo:
Seems canceled due to high concurrency in our repo.

After this PR, build speed will increase, and the cancel rate will drop.

himself65:
> Seems canceled due to high concurrency in our repo.
> 
> After this PR, build speed will increase, and the cancel rate will drop.

of course, canceling the previous run is needed. but it should cancel **current** runs

himself65:
> Why are there always "3 cancelled" after pushing the new commits? Can we run them or ignore them because merge have to pass all of the checks
> 
> /cc @darkskygit
> 
> image

we have fixed this by differ the group-id

himself65 (approved):

.github/workflows/client-app.yml
```diff
@@ -22,7 +22,7 @@ permissions:
   security-events: write
 
 concurrency:
-  group: release-ci-group
+  group: ${{ (github.event_name == 'pull_request' && github.event.pull_request.number) || github.head_ref }}
```
lawvs, line 25: Need `github.workflow` to distinguish from other CI

I suggest use
```suggestion
  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.sha }}
```

linonetwo:
Just finish my lunch...

himself65:
> Just finish my lunch...

I have to merge this because it will boost up the CI

PR #1008, opened 2023-02-15
ci: use cancel workflow action

himself65:
https://github.com/toeverything/AFFiNE/pull/1005#issuecomment-1430656013


himself65:
the new CI won't work I think because the administrator disallow to run it in settings

linonetwo:
Why not `concurrency`?

himself65:
Landed in https://github.com/toeverything/AFFiNE/pull/1005/commits/5c72bf025583fdedcbcdaa17dadf738dddc812d9

PR #1012, opened 2023-02-15
refactor(store): port to `useGlobalState` with zustand

Himself65:
Related: https://github.com/toeverything/AFFiNE/issues/902


doodlewind (approved):

apps/web/src/providers/app-state-provider/Provider.tsx
```diff
@@ -72,6 +43,24 @@ export const AppStateProvider = ({
...
+      if (onceRef.current) {
+        setAppState({
+          workspaceList: dataCenter.workspaces,
+          currentWorkspace: null,
+          pageList: [],
+          synced: true,
+        });
+        onceRef.current = false;
+      } else {
+        console.warn('dataCenter Effect called twice. Please fix this ASAP');
```
himself65, line 59: this warning will be removed in the future when the new store is stable

Himself65 (commented):
In the future, I will use suspense mode on loading dataCenter as it's necessary for every sub-component

PR #1013, opened 2023-02-15
build(desktop): fix typescript error

himself65:


himself65:
The job is still being canceled somehow...

darkskygit (approved):

PR #1050, opened 2023-02-16
fix: move data center into global level

himself65:


apps/web/src/store/app/datacenter/index.tsx
```diff
@@ -114,10 +111,8 @@ export function DataCenterPreloader({ children }: React.PropsWithChildren) {
   }, [api]);
   //# endregion
 
-  if (!dataCenter && !dataCenterPromise) {
-    const promise = getDataCenter();
-    api.setState({ dataCenterPromise: promise });
-    promise.then(async dataCenter => {
+  if (!dataCenter) {
```
pengx17, line 114: May I ask that `dataCenterPromise` is eagerly fetched on module load in this pr is because `!dataCenter && !dataCenterPromise` could enter multiple times and thus call `getDataCenter` multiple times?
I think it is still possible that `dataCenterPromise.then` call will be called multiple times and thus call `createDefaultWorkspace`, right?

apps/web/src/store/app/datacenter/index.tsx
```diff
@@ -130,9 +125,6 @@ export function DataCenterPreloader({ children }: React.PropsWithChildren) {
         dataCenterPageList: [],
       });
     });
-    throw promise;
-  }
-  if (!dataCenter) {
```
pengx17, line 135: I think we should either use a third-party library or make our own abstraction to make the suspense pattern more clear and maintainable.

apps/web/src/store/app/datacenter/index.tsx
himself65: See https://github.com/toeverything/AFFiNE/pull/1076

himself65:
https://github.com/toeverything/AFFiNE/pull/1076

PR #1061, opened 2023-02-16
fix: userinfo issue after state refactor

pengx17:


doodlewind (approved):

doodlewind:
We don't have to use GitHub labels to tag pull requests (by following the conventional commits standard, the commit type is already there). We are just using the labels to classify issues.

For pull requests, we generally only mark special cases, e.g., `don't merge`, `test-needed`, `breaking`, `housekeeping-needed`.

pengx17:
@doodlewind thanks, I have removed the labels.

PR #1069, opened 2023-02-16
feat: set the correct mode when loading the page

zqran:
When the page is converted to `edgeless` mode and reloaded, the editor should be in `edgeless` mode


Himself65 (approved):

zqran:
When the page is in `edgeless` mode, there is no Title, which causes the test to fail.

Should it be handled differently depending on the page mode?

darkskygit:
Hello! Thank you for your contribution.

In order for us to review and merge your suggested changes, please sign at https://github.com/toeverything/AFFiNE/edit/master/.github/CLA.md

darkskygit:
> When the page is in `edgeless` mode, there is no Title, which causes the test to fail.
> 
> Should it be handled differently depending on the page mode?

you can adjust the test cases. judging from the current design, there is indeed no title in edgeless.

Himself65 (changes requested):
Check your git history

zqran:
> Check your git history

yeah, i'm checking~

tests/quick-search.spec.ts
```diff
@@ -9,18 +9,24 @@ const openQuickSearchByShortcut = async (page: Page) =>
   await withCtrlOrMeta(page, () => page.keyboard.press('k', { delay: 50 }));
 
 async function assertTitle(page: Page, text: string) {
-  const locator = page.locator('.affine-default-page-block-title').nth(0);
-  const actual = await locator.inputValue();
-  expect(actual).toBe(text);
+  const edgeless = page.locator('affine-edgeless-page');
+  if (!edgeless) {
```
zqran, line 13:
```suggestion
  const edgeless = page.locator('affine-edgeless-page');
  if (!edgeless) {
```

I'm not very familiar with the testing of this project, will this work?

himself65: no

himself65: you need `await page.locator('affine-edgeless-page').toHaveCount(0);`

Himself65 (approved):

PR #1076, opened 2023-02-16
refactor: data center initialization

himself65:


himself65:
See: https://github.com/toeverything/AFFiNE/pull/1118

PR #1077, opened 2023-02-17
refactor: remove esm module

Himself65:
Remove ESM support before typescript 5.0 as Next.js doesn't support ESM, and we don't release the package to the npm. So why are we torturing ourselves?

## BREAKING CHANGE

- Unit test now uses `vitest`
- ESM droped


doodlewind (changes requested):
I thinks Next.js has ESM support (https://nextjs.org/blog/next-12#es-modules-support-and-url-imports), what's the problem of CJS?

doodlewind (approved):
Got it, we can fallback to CJS if this could help us avoid the redundant build command.

package.json
```diff
@@ -4,12 +4,11 @@
...
-  "type": "module",
   "scripts": {
-    "dev": "cross-env NODE_ENV=development pnpm --filter=!@affine/app build && pnpm --filter @affine/app dev",
-    "dev:ac": "pnpm --filter=!@affine/app build && cross-env NODE_API_SERVER=ac pnpm --filter @affine/app dev",
-    "dev:local": "pnpm --filter=!@affine/app build && cross-env NODE_API_SERVER=local pnpm --filter @affine/app dev",
-    "build": " pnpm --filter=!@affine/app build && pnpm --filter!=@affine/datacenter -r build",
+    "dev": "pnpm --filter @affine/app dev",
+    "dev:ac": "cross-env NODE_API_SERVER=ac pnpm --filter @affine/app dev",
+    "dev:local": "cross-env NODE_API_SERVER=local pnpm --filter @affine/app dev",
+    "build": "pnpm --filter @affine/app build",
```
himself65, line 11: remove duplicated build script

doodlewind (approved):

PR #1092, opened 2023-02-17
fix: `zh-Hans` i18n fails on page reload

zqran:
Steps to reproduce:

1. switch to "简体中文"
2. fefresh page

expect: "简体中文"
received: 'en'


Himself65 (approved):

Himself65 (commented):
does 'zh-cmn-Hans-CN‘ and 'cmn-Hans-CN' also valid inputs here？

zqran:
When switching languages, "简体中文" corresponds to `zh-Hans`. There is no case of 'zh-cmn-Hans-CN' or 'cmn-Hans-CN'.

PR #1106, opened 2023-02-17
refactor(data-center): remove side effect in affine provider

himself65:
First step to support fully SSR mode


doodlewind (approved):

doodlewind:
Please consider the priority of SSR, since our main use case should be within the native client bundle.

himself65:
> Please consider the priority of SSR, since our main use case should be within the native client bundle.

This is not only for SSR. It could also the foundation of the SEO because we need fetch data in server side

PR #1144, opened 2023-02-20
fix: public workspace loading

Himself65:


apps/web/src/pages/index.tsx
```diff
@@ -1,7 +1,15 @@
...
+import { useRouter } from 'next/router';
+import { useEffect } from 'react';
+
+import { PageLoading } from '@/components/loading';
 
 const Home: NextPage = () => {
-  return <div title="Home Page"></div>;
+  const router = useRouter();
+  useEffect(() => {
+    router.push('/workspace');
```
himself65:
```suggestion
    router.replace('/workspace');
```

apps/web/src/pages/_app.tsx
```diff
@@ -73,35 +64,16 @@ const App = ({ Component, pageProps }: AppPropsWithLayout) => {
...
+        </Suspense>
       </ProviderComposer>
     </>
   );
 };
 
-const AppDefender = ({ children }: PropsWithChildren) => {
-  const router = useRouter();
-  useEffect(() => {
-    if (['/index.html', '/'].includes(router.asPath)) {
```
pengx17, line 99: This tauri fix is no longer needed?

apps/web/src/pages/index.tsx
```diff
@@ -1,7 +1,15 @@
...
 const Home: NextPage = () => {
-  return <div title="Home Page"></div>;
+  const router = useRouter();
+  useEffect(() => {
+    router.replace('/workspace');
+  }, [router]);
+  return <PageLoading />;
 };
 
 export default Home;
```
pengx17, line 15: Since Home does not require client logic, we can use next js redirect instead:

```suggestion
const Home: NextPage = () => {
  return <PageLoading />;
};

export default Home;

export function getStaticProps() {
  return {
    redirect: {
      destination: '/workspace',
      permanent: false,
    },
  }
}
```

himself65: Dont do this, `next export` doesn't support this

apps/web/src/pages/_app.tsx
himself65: I believe so

apps/web/src/pages/index.tsx
himself65: See https://github.com/toeverything/AFFiNE/actions/runs/4226181756/jobs/7339338840

pengx17 (approved):
LGTM

PR #1147, opened 2023-02-20
refactor(web): create workspace

Himself65:


packages/store/src/app/index.tsx
```diff
@@ -84,6 +85,22 @@ export const useGlobalState: UseBoundStore<Store> = ((
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
 }) as any;
 
+export async function mutateDataCenter(
```
pengx17, line 88: I think this function should be changed to something like `createWorkspace`

PR #1154, opened 2023-02-21
feat: remove edundant code

zqran:


darkskygit:
target will be undefined when an invalid profile is incoming externally

apps/web/next.config.mjs
```diff
@@ -23,7 +23,7 @@ const profileTarget = {
 };
 
 const getRedirectConfig = profile => {
-  const target = profileTarget[profile || 'dev'] || profileTarget['dev'];
+  const target = profileTarget[profile || 'dev'];
```
zqran, line 26:
```ts
const target = profileTarget[profile || 'dev'] || profileTarget['dev'];
```

when profile is `undefined` will fallback to `'dev'`, `profileTarget[profile || 'dev']` has been achieved.

darkskygit: if we pass profile like `abcd`, at this point target is `undefined`

zqran: oh, got it~

PR #1155, opened 2023-02-21
fix: load page on first render

himself65:


himself65:
need to refactor, the `pageList` is not syncing with `currentDataCenterWorkspace`

JimmFly (approved):

JimmFly (approved):

PR #1169, opened 2023-02-21
feat(data-center): add broadcast channel in local provider

himself65:


himself65:
Dont merge now. If the solution is ok I will refactor the code to fit the sturcture

PR #1176, opened 2023-02-22
refactor!: next generation AFFiNE code structure

Himself65:
Closes: https://github.com/toeverything/AFFiNE/issues/1197


apps/web/src/shared/apis.ts
```diff
@@ -0,0 +1,24 @@
...
+} from '@affine/datacenter';
+
+import { config } from './env';
+
+let prefixUrl = '/';
+if (typeof window === 'undefined') {
+  // SSR
+  if (config.serverAPI.startsWith('100')) {
+    // This is for Server side rendering support
+    prefixUrl = new URL('http://' + config.serverAPI + '/').origin;
```
Saul-Mirone, line 15: Can the protocol here be `https`?

himself65: it's http. Anyway this field haven't been used really so it wont make things danger

darkskygit (approved):

Issue #1193, opened 2023-02-24
Add a logout mechanism for accounts so that local data is preserved even when the Token expires


joebeijing:
When token is expired, a Toast message with the text "Session expired, please log in again." will be displayed and clicking on "Login" will bring up the login window.
The toast do not auto-close, need to be closed manually.

QiShaoXuan:
The temproary solution now is show a toast display: 'Automatic login fail, please login self later'， and then jump to the local workspace

PR #1194, opened 2023-02-24
fix: stay loading when refresh token  fail

QiShaoXuan:
fix: [#1193](https://github.com/toeverything/AFFiNE/issues/1193)

### DIFF
--- a/apps/web/src/components/message-center-handler/index.tsx
+++ b/apps/web/src/components/message-center-handler/index.tsx
@@ -1,17 +1,20 @@
 import { toast } from '@affine/component';
 import { DataCenter, MessageCenter } from '@affine/datacenter';
 import { AffineProvider } from '@affine/datacenter';
+import { DebugLogger } from '@affine/debug';
 import { useRouter } from 'next/router';
 import { ReactNode, useCallback, useEffect } from 'react';
 
 import { useGlobalState } from '@/store/app';
+
+const logger = new DebugLogger('messageCenter');
 
 const clearAuth = (dataCenter: DataCenter, providerName: string) => {
   const affineProvider = dataCenter.providers.find(p => p.id === providerName);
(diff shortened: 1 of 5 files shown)


apps/web/src/components/message-center-handler/index.tsx
```diff
@@ -1,11 +1,19 @@
...
 import { ReactNode, useCallback, useEffect } from 'react';
 
 import { useGlobalState } from '@/store/app';
 
+const clearAuth = (dataCenter: DataCenter, providerName: string) => {
+  const affineProvider = dataCenter.providers.find(p => p.id === providerName);
+  if (affineProvider && affineProvider instanceof AffineProvider) {
+    affineProvider.apis.auth.clear();
+  } else {
+    console.error('cannot find affine provider, please fix this ASAP');
```
pengx17, line 14: Please use `DebugLogger` in `'@affine/debug'`

darkskygit (approved):

apps/web/src/components/message-center-handler/index.tsx
```diff
@@ -36,9 +37,16 @@ export function MessageCenterHandler({ children }: { children?: ReactNode }) {
           await router.push('/');
           router.reload();
         }
+
+        if (message.code === MessageCenter.messageCode.refreshTokenError) {
```
pengx17, line 41: It seems `refreshTokenError` and `noPermission` errors are handled in the same way right now (other than the error toast). Maybe we can merge the codes?

QiShaoXuan: RefreshTokenError is handled as a temporary and quick solution, with complex logic added later

pengx17 (approved):
I left several insignificant issues. Free free to ignore them if it is urgent.

PR #1195, opened 2023-02-24
Search hotkey prevent default

m1212e:
Ctrl + k is the default Firefox shortcut to focus the search bar. The default behavior should be prevented.


lawvs (approved):
Nice catch!

PR #1290, opened 2023-03-03
ci: revise workflows to use .nvmrc (node 18)

hehex9:
update `.nvmrc` to match [readme](docs/BUILDING.md)

> Up to now, the major node.js version is 18.x


Himself65 (commented):
@darkskygit im afraid this will break cloudflare building

darkskygit (approved):

darkskygit:
> @darkskygit im afraid this will break cloudflare building

we use env to fixed the version in cf building, this should have a higher priority than .nvmrc

PR #1296, opened 2023-03-03
fix: workspace setting avatar flickering

### LINKED ISSUE #1175: Workspace settings avatar flicker
Flicker

Particularly the workspace image will flicker when changing tabs.

You can see a flicker of the default green image in this case.

sinchang:
Fixes #1175 

Kapture 2023-03-04 at 00 05 37

### DIFF
--- a/apps/web/src/components/affine/workspace-setting-detail/panel/general/index.tsx
+++ b/apps/web/src/components/affine/workspace-setting-detail/panel/general/index.tsx
@@ -56,7 +56,7 @@ export const GeneralPanel: React.FC<PanelProps> = ({
 
   return (
     <>
-      <StyledRow>
+      <StyledRow height={72}>
         <StyledSettingKey>{t('Workspace Avatar')}</StyledSettingKey>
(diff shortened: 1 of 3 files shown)


himself65 (changes requested):
Thanks for contributing, but this is not a good approach.
For example, there're too many if else and undefined and null

I'm thinking of using suspense mode on the whole workspace loading. It need a little time

himself65 (changes requested):
Blocking until https://github.com/toeverything/AFFiNE/pull/1304

himself65:
As I have landed the suspense mode, we can support suspense on avatar now.

sinchang:
> As I have landed the suspense mode, we can support suspense on avatar now.

Could you give me some hints on how to use suspense on the avatar component?

himself65:
> > As I have landed the suspense mode, we can support suspense on avatar now.
> 
> Could you give me some hints on how to use suspense on the avatar component?

Sorry I didn't see the message and I've fix the issue here https://github.com/toeverything/AFFiNE/pull/1319

PR #1307, opened 2023-03-04
chore: add launch.json for vscode

AlessioGr:
That way, people can launch the project quicker in vscode after cloning it


Himself65 (approved):
LGTM

Issue #1257, opened 2023-03-02
Increase "workspace settings" entrance and "whether sync with affine" menu


Svaney-ssman:
https://www.figma.com/file/DGWHxU8cZHGT1pksAbRZlZ/virgo-2-1.0-editor-area(AFFiNE)?node-id=11316%3A15984&t=ot36cnPSh1DHkmJ5-4

![Image](https://user-images.githubusercontent.com/110808979/222386398-ec3b2871-1330-4279-af0d-2fed3c8c6fd6.png)

PR #1310, opened 2023-03-04
feat: show all pages in the sidebar

AlessioGr:
https://user-images.githubusercontent.com/70709113/222915606-e751e22c-6dd1-4260-a255-bf0a04864b03.mp4

This was feature thing I missed a lot from the pre-alpha: seeing all pages in the sidebar.

Additionally, I also put "All Pages" under "Favorites" as I think that's better UX - favorites are pretty much the most important pages, so they should be displayed on top.


AlessioGr:
https://user-images.githubusercontent.com/70709113/222916145-a76dab8c-5b20-4023-a5fe-bd2395d82b4d.mp4

Also added a plus button inside of the "all pages" sidebar item. Makes creating new pages a lot more intuitive in my opinion.

AlessioGr:
Fixed all tests! The test "debug-page-broadcast.spec.ts" seems to be broken in general, which is not related/caused by this PR

himself65 (changes requested):
Code itself LGTM. But the UI/UX needs to be confirmed by design team

/cc @joebeijing

AlessioGr:
Alright! What I'd also like to add is moving the "Workspace Settings" button over to the top, where the workspace is:

Google Chrome 2023-03-04 at 18 31 24@2x

And maybe move the "Trash" button all the way down where the "New page" button is - that way, the "All pages" section would be _most-bottom_ section in the top part of the sidebar. I think for me personally, that would feel cleaner & more intuitive UX-wise, let me know what you think about that too!

himself65:
> 

The product team said that the sidebar items and the workspace creation and settings need to be refactored.

For example, when the user deletes all workspaces. The creation modal should forcibly display. And creation should have workspace type selection(for example, local, affine, or even self-host)

Related: https://github.com/toeverything/AFFiNE/issues/1257

That may be a good start for you. 

I'm refactoring the workspace loading logic supporting React 18 suspense mode. https://github.com/toeverything/AFFiNE/pull/1304 . I think this PR also breaks anything old logic

AlessioGr:
> I think this PR also breaks anything old logic

any old logic for just the Workspace modal, or for the sidebar as well?

What I'd suggest would pretty much only affect the sidebar, not the Workspace creation modal. I'd leave the modal alone for now (I do like the changes proposed in #1304 though).

What I'd propose first is updating the sidebar to look something like this:
image

So basically putting the Trash button down, and removing the "Workspace Settings" button inside of the Workspace selection button - but just as the icon. That way, it would be all together and users can change the workspace settings with just one click.

If they then want to change settings of other Workspaces, they could click on it (which would be visually _suggested_ with this icon:
image
) and change it there.

Also added a little spacing between "Search" and "Favorites" - notion seems to do the same. I think it adds more clarity as to where the pages section starts

What do you think?

himself65:
> any old logic for just the Workspace modal, or for the sidebar as well?

The sidebar code is still from the old code structure from months ago. You can search `fixme(himself65)` under that directory. What I want it to be is a single pure component and pluggable. I can specify the sidebar items in the `WorkspacePlugin` UI. In the future, I will combine `public-workspace` also into `workspace` by declaring a new workspace type `RemWorkspaceFlavour.PUBLIC_AFFINE`; this feature would help to block some UI from the public page

```tsx
const LocalWorkspacePlugin = {
// ...
SidebarMenu: ({ router }) => {
  const currentWorksapce = useCurrentWorkspace()
  return (
    <Menu>
      <MenuItem>
        All Pages
      </MenuItem>
      <MenuItem>
        Settings
      </MenuItem>
    </Menu>
  )
} 
}
```

joebeijing:
We are designing a separate sidebar Page Tree for sets of pages that need to be presented in a Wiki structure. 

Currently, to reduce the pressure of creating pages, pages that appear in All Pages will not be displayed in the sidebar, so they will not be presented in a heavy tree structure by default. Only pages selectively placed in the Wiki will be displayed in the Page Tree and shared with all members of the Workspace. 

However, we will also consider putting this display option in the custom options in the future.

CC @Himself65

darkskygit:
we support full-text search for pages in workspace, so it can largely replace the functionality of the sidebar page tree, so we dont need always display the page tree in sidebar.

himself65:
clsoing as we have page tree now

PR #1319, opened 2023-03-05
fix: avoid avatar flickering

himself65:


himself65:
I found that I wrote the wrong co-author. Trying to force push

PR #1321, opened 2023-03-05
fix: always show workspace list modal when no workspaces

Himself65:
Related: https://github.com/toeverything/AFFiNE/issues/1316


Himself65 (commented):
need test case

PR #1363, opened 2023-03-06
ci: remove self-host in build.yml

Himself65:


.github/workflows/build.yml
```diff
@@ -80,13 +84,19 @@ jobs:
...
-          npx playwright install chromium
-          pnpm run test:coverage
-          pnpm run test:unit
+      - name: Install playwright browsers
+        run: npx playwright install chromium
+      - name: Run playwright tests
+        run: pnpm test -- --forbid-only --shard=${{ matrix.shard }}/${{ strategy.job-total }}
+        env:
+          COVERAGE: true
+      - name: Unit Test
```
darkskygit, line 96: will run 4 times

playwright.config.ts
```diff
@@ -25,11 +31,12 @@ const config: PlaywrightTestConfig = {
     // Record video only when retrying a test for the first time.
     video: 'on-first-retry',
   },
-
-  fullyParallel: true,
-  forbidOnly: !!process.env.CI,
```
darkskygit, line 30: forbid only should be enabled

darkskygit (changes requested):

darkskygit (approved):

PR #1399, opened 2023-03-08
fix: button "jumps" between workplace settings tabs

### LINKED ISSUE #1345: Button "jumps" between tabs (workplace settings)
Collaboration <-> Publish tabs

The button jumps between switching tabs.

jump

sinchang:
Closes #1345


Himself65 (approved):
Nice hack. But I think we should refactor the code in the future

PR #1404, opened 2023-03-08
chore: bump version

Himself65:


doodlewind (commented):
We have migrated to Virgo, having the `workspace.isEmpty` and `workspace.id`. So I would like to bump to a new nightly version adapting to these changes.

doodlewind (approved):

Issue #1423, opened 2023-03-08
Add "Remove Avatar" button on Workspace Settings


albertodlc:
![image](https://user-images.githubusercontent.com/32411964/223814121-520de1a4-c0a6-4c58-bf92-f1e0e2c25c1a.png)

joebeijing:
I would like to understand more about this requirement. In what scenarios do you need to remove the Workspace icon? Currently, you can modify the Workspace icon by uploading an image. @albertodlc

albertodlc:
Hello @joebeijing, I think it is a typical feature to add, por example Notion has it
![image](https://user-images.githubusercontent.com/32411964/224817090-38898347-5c9f-423c-922a-45722b7e0e66.png)

![image](https://user-images.githubusercontent.com/32411964/224817362-2f9cd25a-e956-4d56-a058-42e13556016f.png)

joebeijing:
I tried it and found that even if we remove the icon of the workspace in Notion, Notion will still generate one automatically, similar to the logic of the icons we generate now. However, I’m curious why you need “remove”  function? You can simply upload a replacement icon or do you want to restore the default automatically generated icon? @albertodlc 
<img width="459" alt="CleanShot 2023-03-15 at 4 29 20@2x" src="https://user-images.githubusercontent.com/22443345/225251729-4761c90c-6b9f-4276-bc0b-176238a35051.png">

houjoe0829:
Already released in latest Canary version
<img width="415" alt="CleanShot 2023-09-22 at 14 38 54@2x" src="https://github.com/toeverything/AFFiNE/assets/22443345/64923fc2-f9b4-47fc-9b72-4226765bd920">

PR #1424, opened 2023-03-08
feat: new button "remove avatar"

albertodlc:
Added #1423 

- New button
- Modified rows related styles


himself65 (commented):
/cc @joebeijing

himself65:
I think we don't have the plan to add a delete avatar button here. Closing

PR #1489, opened 2023-03-10
feat: add animation to mode switch when hover

QiShaoXuan:


doodlewind (changes requested):
Minify these 2 JSON files may be better

doodlewind (approved):

Himself65 (changes requested):
let me handle the test error

Himself65 (approved):

Issue #1292, opened 2023-03-03
Improve client build process and release workflow


m1911star:
- [x] Support manually release target version
- [x] Should have a dedicated release page

<img width="369" alt="image" src="https://user-images.githubusercontent.com/4948120/222740196-749b0e18-8fcb-48f3-8113-ac48ee75e593.png">

<img width="1185" alt="image" src="https://user-images.githubusercontent.com/4948120/222740297-2a2c189e-e2c5-47a0-9f1d-e1a6db36d260.png">

ref: https://github.com/m1911star/affine-client/releases

PR #1532, opened 2023-03-11
feat: modify tauri client release workflow

m1911star:
make client app build workflow clear and add release
related issue: https://github.com/toeverything/AFFiNE/issues/1292

> currently we only support build app manually

## How to build a client app

1. update package number

path: apps/desktop/src-tauri/tauri.conf.json

image

2. enter a version & dispatch workflow 

image

3. check action process

image

4. edit changelog and release!

image

5. check release page

## note:

1. rust cache also works 
image

2. release a client build will create a new tag

image


.github/workflows/client-app.yml
```diff
@@ -31,27 +23,29 @@ jobs:
...
+      - name: create release
+        id: create-release
+        uses: actions/github-script@v6
         with:
-          draft: true
-          tag_name: '${{ env.version }}'
-          release_name: 'AFFiNE ${{ env.version }}'
-          body: 'See the assets to download this version and install.'
+          script: |
+            const { data } = await github.rest.repos.createRelease({
```
pengx17, line 39: TIL we can write bare js in gha

.github/workflows/client-app.yml
```diff
@@ -85,24 +79,38 @@ jobs:
           workspaces: |
             apps/desktop/src-tauri
 
-      - name: Make App Binary
-        run: pnpm build:client
+      - name: Build Affine
```
pengx17, line 82: I think this can be removed. tauri build has already included the two in `beforeBuildCommand`.

m1911star: Yes, removed

darkskygit (approved):

PR #1559, opened 2023-03-14
docs: update README.md

ShortCipher5:
Add quick links table for contributing

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,11 @@ Before we tell you how to get started with AFFiNE, we'd like to shamelessly plug
 
 ## Contributing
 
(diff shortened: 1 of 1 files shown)


joebeijing:
Sounds like it doesn't works when I preview the ReadMe. Is it a problem with my operation? @ShortCipher5 
CleanShot 2023-03-15 at 4 22 58@2x

ShortCipher5:
I think that's only a partial view - that part has not been updated - in the full view you can see the following.

Header:
Screenshot 2023-03-16 at 10 02 48

Table:
Screenshot 2023-03-16 at 10 02 54

The only change to this is the table section.

joebeijing (approved):

Himself65 (approved):

PR #1619, opened 2023-03-20
refactor: use yarn

himself65:


himself65:
@darkskygit need check cloudflare settings

pengx17 (commented):
does yarn cross-env work out-of-the box?

himself65:
> yarn cross-env

I dont know

pengx17 (approved):
LGTM

AlessioGr:
@himself65 just out of curiosity: what was the reason for moving from pnpm to yarn?

himself65:
> @himself65 just out of curiosity: what was the reason for moving from pnpm to yarn?

pnpm is not compatible with electron

PR #1624, opened 2023-03-20
ci: refactor build&test logic

himself65:


himself65:
need force merge /cc @doodlewind

PR #1625, opened 2023-03-20
test: enable affine workspace test

himself65:


himself65:
Blocking until https://github.com/toeverything/AFFiNE/pull/1624

PR #1631, opened 2023-03-21
fix: remove duplicated link css

lawvs:
Link CSS has been declared in `affine-link` 

Screenshot 2023-03-21 at 12 50 44 PM


himself65:
I think it will affect the link that is not in the editor. but so far we don't have such that

himself65 (approved):

PR #1647, opened 2023-03-21
fix: remove pageMode from pageMeta

himself65:


himself65:
Status: https://github.com/toeverything/AFFiNE/actions/runs/4485102849

PR #1657, opened 2023-03-22
fix: flash screen when creating new page

himself65:


apps/web/src/providers/ThemeProvider.tsx
```diff
@@ -40,9 +40,9 @@ const ThemeInjector = memo<{
 const ThemeProviderInner = memo<React.PropsWithChildren>(
   function ThemeProviderInner({ children }) {
     const { resolvedTheme: theme } = useTheme();
-    const currentPage = useCurrentPage();
+    const [pageId] = useCurrentPageId();
     const record = useAtomValue(workspacePreferredModeAtom);
-    const editorMode = currentPage ? record[currentPage.id] ?? 'page' : 'page';
+    const editorMode = pageId ? record[pageId] ?? 'page' : 'page';
```
pengx17, line 45: it looks to me that the mode can also be an atom, but I guess it is not necessary tho

pengx17 (approved):

apps/web/src/providers/ThemeProvider.tsx
himself65: updated

PR #1672, opened 2023-03-23
feat(client): add octobase-node to electron

pengx17:
Add octobase-node to electron. The octobase-node napi is incomplete right now and we will use this PR as a starting point for next stage of building client app stuff.

This PR also includes:
- [x] use yarn instead of pnpm in electron build
- [x] add octobase-node to pacakages/octobase-node. This is a Napi.rs wrapper around octobase.
- [x] use octobase-node in Electron to demonstrate IPC usages.


apps/electron/package.json
```diff
@@ -14,13 +14,14 @@
     "make-macos-arm64": "electron-forge make --platform=darwin --arch=arm64",
     "make-macos-x64": "electron-forge make --platform=darwin --arch=x64",
     "make-windows-x64": "electron-forge make --platform=win32 --arch=x64",
-    "postinstall": "ELECTRON_RUN_AS_NODE=1 electron scripts/update-electron-vendors.mjs"
+    "postinstall": "ELECTRON_RUN_AS_NODE=1 electron scripts/update-electron-vendors.mjs && yarn workspace @affine/octobase-node build"
```
himself65: here we run `cargo build` but all of CI doesn't have such that command

himself65 (changes requested):

apps/electron/package.json
pengx17: I will remove this and let it only run in CI instead

himself65 (approved):

PR #1710, opened 2023-03-28
fix: windows rust cache

pengx17:

PR #1716, opened 2023-03-28
ci: fetch version by node

himself65:


himself65:
Fixed in https://github.com/toeverything/AFFiNE/pull/1710

PR #1722, opened 2023-03-28
feat(cli): add `octobase-service`

himself65:


packages/cli/src/scripts/octobase-service.ts
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env ts-node-esm
```
pengx17, line 1: It looks like this file is purely js. Maybe we can just use node instead and get rid of one extra dep?

pengx17 (approved):

himself65:
I think we should make things duplicated. You should learn how to use docker

PR #1729, opened 2023-03-28
feat: support sentry report

Himself65:


Himself65 (commented):
Need to conform that whether we need this feature

darkskygit (approved):

Issue #1594, opened 2023-03-16
Change log support


joebeijing:
This is mainly used to showcase what AFFiNE has recently updated.

xell:
Please fix the UI bug:

![CleanShot 2023-05-15 at 12 00 46@2x](https://github.com/toeverything/AFFiNE/assets/132558/7a3b8a7a-0fda-4943-a1d4-03222a64d2d8)

pengx17:
I think in the new design, changelog will be merged into the app updater button so that this one was recently removed from the latest version. cc @hwangdev97

JimmFly:
> I think in the new design, Change Log component will be merged into the app updater button so that this one was recently removed from the latest version. cc @hwangdev97

Does it still need to be displayed on the web?

hwangdev97:
I think both the website and the client should display the new design, the new design should be structured as follows:

<img width="703" alt="shot-2023-05-15-13 17 35@2x" src="https://github.com/toeverything/AFFiNE/assets/24713927/3ff2e011-8029-410d-9957-7cf3bdf848ef">

pengx17:
@m1911star will you be able to make the ui here as well?

m1911star:
> @m1911star will you be able to make the ui here as well?

I'm not entirely sure when the deadline is. Will it be released together with the beta?

pengx17:
> I'm not entirely sure when the deadline is. Will it be released together with the beta? @xell @pengx17

Yeah, I hope so. I think the UI part shall be trivial, but you may need to handle different states.

xell:
> I'm not entirely sure when the deadline is. Will it be released together with the beta? @xell @pengx17 

The deadline of beta is May 18 17:30 UTC+8. Please refer [the internal doc](https://toeverything.feishu.cn/wiki/OyKBw7yJhiXQJVkyogacP7tZnEc) (you can request permission from me).

m1911star:
Sorry. I cant catch up with beta release date these two days. Maybe will find some other bugs to fix later.

xell:
> Sorry. I cant catch up with beta release date these two days. Maybe will find some other bugs to fix later. 

Ok. It's understandable. Thanks for all the work.

pengx17:
No worries @m1911star, I can take over the task

PR #1734, opened 2023-03-29
feat: add change log

JimmFly:
Work for #1594


Himself65 (commented):
.

apps/web/src/atoms/first-load.ts
```diff
@@ -1,5 +1,47 @@
...
+  version: string;
+  openTips: boolean;
+  showWhatsNew: boolean;
+};
+
+export const isFirstLoadAtom = atomWithStorage<FirstLoad>('isFirstLoad', {
+  version: '0.0.0',
+  openTips: false,
+  showWhatsNew: false,
+});
```
himself65: I think you should split them to each single atom

JimmFly: Indeed, but I feel like there might be a way to unify storage? something like this.
```
export type FirstLoadGuide = {
  version: string;
  closedInCurrentSession: Record<string, boolean>;
  permanentlyClosed: Record<string, boolean>;
};

export const firstLoadAtom = atomWithStorage<FirstLoadGuide>('isFirstLoad', {
  version: CURRENT_APP_VERSION,
  closedInCurrentSession: {},
  permanentlyClosed: {},
});
```

apps/web/src/atoms/first-load.ts
```diff
@@ -1,5 +1,21 @@
...
-export const openTipsAtom = atom<boolean>(false);
+export type Visibility = Record<string, boolean>;
+
+const DEFAULT_VALUE = '0.0.0';
+//atomWithStorage always uses initial value when first render
+//https://github.com/pmndrs/jotai/discussions/1737
+export const lastVersionAtom = atomWithStorage(
+  'lastVersion',
+  JSON.parse(
+    globalThis.localStorage?.getItem('lastVersion') ??
```
himself65: Just notice that this will cause SSR hydration error. We didn’t see it just I disabled the SSR in the root component

JimmFly:
```suggestion
//     globalThis.localStorage?.getItem('lastVersion') ??
function getInitialValue() {
  if (typeof window !== 'undefined') {
    const storedValue = window.localStorage.getItem('lastVersion');
    if (storedValue) {
      return JSON.parse(storedValue);
    }
  }
  return DEFAULT_VALUE;
}

export const lastVersionAtom = atomWithStorage(
  'lastVersion',
  getInitialValue())
```
Can adding a window detection solve it?

apps/web/src/atoms/first-load.ts
himself65: I won't work I believe. I think you should hide all tips by default and only show then in the second render. Not showing all of them in first time

Issue #1573, opened 2023-03-14
Support responsive UI Layout


pengx17:
Provide different UI layouts for small and extra large screens.
- [x] design spec by @xell 
- [ ] implementation

PR #1741, opened 2023-03-29
feat: responsive ui above 640px

pengx17:
- resizing left sidebar
- responsive ui >640 (see https://cqdm24.axshare.com/#id=a6k335&p=responsive_ui_demo)
- [x] fix tests
- [x] add test for resizing sidebar

fix https://github.com/toeverything/AFFiNE/issues/1573


xell (changes requested):
1. While the window width W > 960, the sidebar width should be constrained by both a) >= 256, and b) <= `min(W-698, 480)`.
2. The drag hover area is too narrow.
3. And the cursor shape might be changed to `col-resize` rather than `ew-resize`.

hwangdev97:
Something about color:
- The hover indication color of dragging components (such as the blue line when dragging blocks and the sidebar) should use the primary color. However, the primary color for dark mode will be adjusted later.
- Modal dialog’s background color is `rgba(0,0,0,0.6)` in light mode, and `rgba(0,0,0,0.8)` in dark mode, and it’s named as a variable backgroundModalColor.

pengx17:
Expand the resizing censoring area to make it easier to grab.
image

apps/web/src/components/pure/workspace-slider-bar/index.tsx
```diff
@@ -64,117 +73,162 @@ export const WorkSpaceSliderBar: React.FC<WorkSpaceSliderBarProps> = ({
...
-            visible={sidebarOpen}
-            tooltipContent={t('Collapse sidebar')}
-            testid="sliderBar-arrowButton-collapse"
-          />
-        </StyledSidebarSwitchWrapper>
+      <StyledSliderBarWrapper>
+        <StyledSliderBar
+          resizing={isResizing}
+          floating={floatingSlider}
+          style={{ width: actualWidth }}
```
himself65, line 121: can we use `sx`?

pengx17: I tried `sx`as well as passing width as a parameter to styled-component. Looks like it will create a new class for every new width. I guess it is better to use inline-styles in this case.

https://user-images.githubusercontent.com/584378/228752960-c26fd546-7152-4f34-a5a8-e5220a3512b0.mp4

Himself65 (approved):

PR #1744, opened 2023-03-29
feat: add `getUsage` api

Himself65:


packages/workspace/src/affine/__tests__/api.spec.ts
```diff
@@ -208,4 +212,30 @@ describe('api', () => {
...
+      const blobId = await workspaceApis.uploadBlob(
+        id,
+        imageBuffer,
+        'image/png'
+      );
+      const buffer = await workspaceApis.getBlob(blobId);
+      expect(buffer.byteLength).toEqual(imageBuffer.byteLength);
+      const newUsageResponse = await userApis.getUsage();
+      usageResponseSchema.parse(newUsageResponse);
+      expect(usageResponse.blob_usage.usage).not.equals(
```
himself65, line 234: /cc @zuoxiaodong0815 

`getUsage` won't return the correct usage after we upload a blob

PR #1773, opened 2023-03-31
fix: first workspace create logic

pengx17:
- [x] add `@affine/jotai` package. Right now it only has `atomWithSyncStorage` to replace official `atomWithStorage`
- [x] replace all `atomWithStorage` with `atomWithSyncStorage`
- [x] make sure only one workspace will be created if there is only one workspace
- [x] use jotai v2 api 'store.sub' to handle `createFirst` workspace logic

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -14,6 +14,7 @@
     "@affine/debug": "workspace:*",
     "@affine/env": "workspace:*",
     "@affine/i18n": "workspace:*",
+    "@affine/jotai": "workspace:*",
     "@affine/templates": "workspace:*",
     "@affine/workspace": "workspace:*",
     "@blocksuite/blocks": "0.5.0-20230324040005-14417c2",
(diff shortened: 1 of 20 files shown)


himself65:
Will this cause the hydration issue?

pengx17:
> Will this cause the hydration issue?

No, I didn't see any so far

himself65:
> > Will this cause the hydration issue?
> 
> No, I didn't see any so far

It's because 

https://github.com/toeverything/AFFiNE/blob/fd65dd66a1943edadc8be633a9139f79c875b83b/apps/web/src/layouts/index.tsx#L159

pengx17:
> It's because
> 
> https://github.com/toeverything/AFFiNE/blob/fd65dd66a1943edadc8be633a9139f79c875b83b/apps/web/src/layouts/index.tsx#L159

Hmm... With this, SSR is always off right? I think our app is not SSR compatible in many other aspects. Do we need to think about supporting SSR with storage now?

himself65 (approved):

PR #1777, opened 2023-04-02
refactor: enable next.js `appDir`

himself65:


himself65:
closing as we don't have plan to use next.js appDir

PR #1805, opened 2023-04-04
feat: single page sharing support

JimmFly:


himself65:
I'm handling this PR now.

himself65 (approved):
I'm merging this because the basic logic and UI have finished. I think next PR should embed this component into affine and polish the UI part

PR #1849, opened 2023-04-07
refactor: try using atom family for workspaces

pengx17:
This allows workspace can be loaded individually, instead of loading each of them all at once.
ONLY FOR DEMO purpose.


himself65 (commented):
This doesn't have much benefit to boost the initial time since all workspaces must be settled before the router.

himself65 (commented):
I suggest splitting the `workspaces` and `currentWorkspace` atoms.

8UUE4LGG7m

pengx17:
I will try to figure out some solutions in a day or two

pengx17:
> This doesn't have much benefit to boost the initial time since all workspaces must be settled before the router.

This is challenging and may need a broader refactor. Let me see how far I can go for this one.

packages/workspace/src/type.ts
```diff
@@ -79,13 +79,20 @@ export interface WorkspaceRegistry {
   [WorkspaceFlavour.LOCAL]: LocalWorkspace;
 }
 
+interface CRUDOptions {
+  local: boolean;
```
himself65, line 94: I think this should be removed

himself65: I think upstream (which is next.js layer) shouldn't care about the detail of the loading logic in each plugin

packages/component/src/components/workspace-avatar/index.tsx
```diff
@@ -31,7 +31,7 @@ function stringToColour(str: string) {
 
 export type WorkspaceAvatarProps = {
   size?: number;
-  workspace: LocalWorkspace | AffineWorkspace | null;
+  workspace?: LocalWorkspace | AffineWorkspace | null;
```
himself65: Too much `?` here, could you please narrative it to `null` instead?

apps/web/src/plugins/affine/index.tsx
```diff
@@ -140,9 +140,9 @@ export const AffinePlugin: WorkspacePlugin<WorkspaceFlavour.AFFINE> = {
       });
       await mutate(matcher => matcher === QueryKey.getWorkspaces);
     },
-    get: async workspaceId => {
+    get: async (workspaceId, opts) => {
```
himself65, line 147: I think the correct implementation should be 

```ts
const workspaces = await loadFromIndexeddb()
// do not await remote fetch
loadFromCloud().then(remoteWorkspaces => {
  const result = merge(workspaces, remoteWorkspaces);
  // tell the upstream to rerender
  context.setWorkspaces(ws => merge(ws, result))
})

return workspaces 
```

himself65 (changes requested):

apps/web/src/plugins/affine/index.tsx
himself65: In this case, typeof Plugin should be replaced by `PluginFunction`

```ts
type PluginFunction = (context: AppContext) => Plugin
interface AppContext {
  setWorkspaces: Function
  // ...
}
```

pengx17: Some of my thoughts:

For data loading in two phases:
The current implementation does not think about load all workspaces all together. In the design, each workspace atom is a derived atom from its ID, where the ID is from different providers (merged); after we have ID (and its type), we will then load individual data by different plugin types and its value can be originated from multiple providers.

For introducing a context to trigger rerender:
I think the philosophy of WorkspacePlugin is just a upstream data fetcher, which implements how to load/mutate data without inferencing other states. Having a context here may make the data flow harder to be understood.

Ideally, I want to have a GET function which will emit value multiple times, so that we have the optimal data flow that reactively drives the UI. This is not possible with an async function, but is possible with an Observable (in Rxjs). Before  we are going to have a more reactive library, I think right now the possible solution is to have GET (+ List?) to return `Promise[]`, so that the caller of GET can be responsible for dealing with the multiple outputs.

himself65:
Closing as too many conflicts

PR #1859, opened 2023-04-10
feat: connect pinboard and reference link

QiShaoXuan:


apps/web/src/hooks/affine/use-reference.ts
```diff
@@ -0,0 +1,36 @@
...
+      subpageLinked?.({ pageId });
+    });
+    editor.slots.subpageUnlinked.on(({ pageId }) => {
+      subpageUnlinked?.({ pageId });
+    });
+
+    return () => {
+      editor.slots.pageLinkClicked.dispose();
+      editor.slots.subpageLinked.dispose();
+      editor.slots.subpageUnlinked.dispose();
```
lawvs, line 33: You should not call the dispose method on the slot in the editor. Try using the following pattern.

```ts
  const disposable = editor.slots.pageLinkClicked.on(({ pageId }) => {
      pageLinkClicked?.({ pageId });
   });
  return () => disposable.dispose()
```

apps/web/src/hooks/use-pinboard-handler.ts
```diff
@@ -34,35 +34,61 @@ export function usePinboardHandler({
...
+            reference: {
+              type: 'Subpage',
+              pageId: referenceId,
+            },
+          },
+        },
+      ]);
+      const [frame] = page.getBlockByFlavour('affine:frame');
+
+      frame && page.addBlock('affine:paragraph', { text }, frame.id);
```
lawvs, line 57: You should throw an error as this behavior is unexpected.

QiShaoXuan: It can not pass unit test if throw error, cause there is no editor

Himself65 (approved):

PR #1871, opened 2023-04-11
feat: refactor trash, page would delete from its parent's subpageIds …

QiShaoXuan:
…after move to trash

### DIFF
--- a/apps/web/src/hooks/affine/use-affine-helper.ts
+++ b/apps/web/src/hooks/affine/use-affine-helper.ts
@@ -1,0 +1,70 @@
+import { useCallback } from 'react';
+
+import type { BlockSuiteWorkspace } from '../../shared';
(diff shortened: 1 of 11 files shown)


apps/web/src/hooks/affine/use-affine-helper.ts
```diff
@@ -0,0 +1,70 @@
+import { useCallback } from 'react';
+
+import type { BlockSuiteWorkspace } from '../../shared';
+import { usePageMeta, usePageMetaHelper } from '../use-page-meta';
+
+export function useAffineHelper(
```
himself65: Why we have a `Affine` helper but doesn't use any API of affine. I think this should be moved to normal `metaHelper`

Himself65 (changes requested):

Himself65 (approved):

PR #1888, opened 2023-04-12
fix: api server address in electron

pengx17:


packages/y-indexeddb/src/index.ts
```diff
@@ -319,14 +322,20 @@ export const createIndexedDBProvider = (
...
           ],
         });
+        const callback = (_update: Uint8Array, origin: unknown) => {
+          if (origin === selfUpdateOrigin) {
+            early = false;
+            doc.off('update', callback);
+            resolve();
+          }
+        };
+        doc.on('update', callback);
```
himself65: What's this

himself65: I force pushed wrong commits today, please drop this commit anyway

Himself65 (approved):

Issue #1952, opened 2023-04-14
flaly e2e test


Himself65:

himself65:
https://github.com/toeverything/AFFiNE/actions/runs/4703304889/jobs/8341670864?pr=1953

himself65:
https://github.com/toeverything/AFFiNE/actions/runs/4703304889/jobs/8341671326

himself65:
https://github.com/toeverything/AFFiNE/pull/1951

himself65:
https://github.com/toeverything/AFFiNE/pull/1955

Issue #1810, opened 2023-04-04
Workspace plugin abstraction


himself65:
# Introduction

- The Workspace plugin is mainly for better maintenance and allows the community to add more workspace types (like connecting with the database on the desktop or other authorized workspace functions).
- Each part of the workspace plugin should work individually well. UI could run on `storybook` without real logic, and logic should work fine on `vitest.`

## Code Structure

- `@affine/workspace`
  - AFFiNE: affine workspace
  - Local: local workspace
  - ... you self-designed a workspace 

Each Workspace Plugin has its state and is isolated from each other workspace. 

There are two main parts of the plugin: `UI` and `logic.`

## Logic

Each workspace plugin should provide basic CRUD API for the upstream calling.

```tsx
type CRUD = {
  get: (id: string) => Promise<Workspace | null>
  create: (id: string) => Promise<Workspace>
  delete: (id: string) => Promise<Workspace>
  list: () => Promise<Workspace[]>
}
```

Each workspace CRUD should maintain its persistence state, whether saved in localStorage, indexedDB, remote server, random data, or a combination of them. AFFiNE won't care about it. AFFiNE will only store `id` and `flavor` for each workspace on top to make sure each refresh page will keep track of all tracked workspaces in the last time.

## UI

UI is highly coupled with its type of workspace. The last `data-center` code design does not isolate the workspace data but combines all workspace types into a single UI, making tiny code changes that will cause a high maintenance cost.

The new code design **requires** you to design your UI for the type of workspace; of course, you can reuse the other components, but remember that UI should be combined with logic. Doing this allows you to code much stronger type and data assertion in each component.

## Data Structure

For example, the type of local workspace

```tsx
export interface LocalWorkspace {
  flavour: WorkspaceFlavour.LOCAL;
  id: string;
  blockSuiteWorkspace: BlockSuiteWorkspace;
  providers: Provider[];
}
```

### Provider

The provider is the way to connect a single `yDoc` (or blocksuite store) to the external world, like persistence with indexeddb or OctoBase server. It should promise to clean up the side effect. For example, when the user switches a workspace, all previous providers should be cleaned up correctly.

There are also two types of providers: `background` and `foreground.` This concept is easy to understand that when the user enters the AFFiNE, all of the local data should be loaded on the workspace list.

himself65:
## Related

- https://github.com/toeverything/AFFiNE/pull/1661
- https://github.com/toeverything/AFFiNE/pull/1778
[...]

PR #1966, opened 2023-04-16
refactor: workspace loading logic

Himself65:
Fixes: https://github.com/toeverything/AFFiNE/issues/1952

Related: https://github.com/toeverything/AFFiNE/issues/1810


apps/web/src/atoms/root.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Fetch all workspaces from the Plugin CRUD
+ */
+export const workspacesAtom = atom<Promise<AllWorkspace[]>>(async get => {
+  const flavours: string[] = Object.values(WorkspacePlugins).map(
+    plugin => plugin.flavour
+  );
+  const jotaiWorkspaces = get(rootWorkspacesMetadataAtom).filter(workspace =>
+    flavours.includes(workspace.flavour)
+  );
+  const workspaces = await Promise.all(
```
pengx17, line 25: I think we do not actually need to wait for whole workspaces to load. We just need the workspace id and then load individually.

apps/web/src/atoms/index.ts
```diff
@@ -1,40 +1,68 @@
...
-export const currentWorkspaceIdAtom = atom<string | null>(null);
-export const currentPageIdAtom = atom<string | null>(null);
-export const currentEditorAtom = atom<Readonly<EditorContainer> | null>(null);
+/**
+ * @deprecated Use `rootCurrentWorkspaceIdAtom` directly instead.
+ */
+export const currentWorkspaceIdAtom = rootCurrentWorkspaceIdAtom;
+
+// todo(himself65): move this to the workspace package
+rootWorkspacesMetadataAtom.onMount = setAtom => {
```
pengx17, line 24: I am not sure about onMount, does it run one every use cases? Which means the creatFirst could run every place that is being used?

himself65: When there's no workspace, we should call `app:init` event

apps/web/src/atoms/root.ts
himself65: image

For now there is only modals needs to read all workspaces. Not a big deal

PR #2014, opened 2023-04-18
feat(graphql): generate types from graphql files

Brooooooklyn:
Initialize the `@affine/graphql` package, you can run `yarn workspace @affie/graphql gql-gen` to generate `src/graphql/index.ts` and `src/schema.ts` files.

The `src/graphql/index.ts` is generated by `packages/graphql-client/export-gql-plugin.cjs`;

A SWR request wrapper has been written and placed in `app/web/shared/gql.ts`

Queries:

```ts
import { someQuery, someQueryWithNoVars } from '@affine/graphql'

const swrResponse1 = useQuery({
  query: workspaceByIdQuery,
  variables: { id: '1' }
})

const swrResponse2 = useQuery({
  query: someQueryWithNoVars
})

```

Mutations:

```ts
import { someMutation } from '@affine/graphql'

const { trigger } = useMutation({
  mutation: someMutation
})

trigger({ name: 'John Doe' })

```


packages/graphql-client/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@affine/graphql-client",
+  "version": "0.0.0",
+  "description": "Autogenerated GraphQL client for affine.pro",
+  "license": "MIT",
```
himself65: not mit

packages/graphql-client/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@affine/graphql-client",
```
forehalo, line 2: why not just `@affine/graphql`, where we put all gql/schema/codegen stuffs in, and `GraphQLClient` that used to send gql request as well

apps/web/src/shared/gql.ts
```diff
@@ -0,0 +1,105 @@
...
+    SWRConfiguration<
+      QueryResponse<Query>,
+      GraphQLError | GraphQLError[],
+      typeof fetcher<Query>
+    >,
+    'fetcher'
+  >
+): SWRResponse<QueryResponse<Query>, GraphQLError | GraphQLError[]>;
+export function useQuery(options: QueryOptions<any>, config?: any) {
+  return useSWR(options.query.id, () => fetcher(options), config);
```
promer94, line 54: You might also want to include `options.variables` as the `key` here.

```tsx
const swrResponse1 = useQuery({
 query: workspaceByIdQuery,
 variables: { id: '1' }
})
const swrResponse2 = useQuery({
 query: workspaceByIdQuery,
 variables: { id: '2' }
})
```
Then `swrResponse1` and `swrResponse2` could be cached individually

apps/web/src/shared/__tests__/gql.spec.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { render } from '@testing-library/react';
+import type { Mock } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { useMutation, useQuery } from '../gql';
+
+let fetch: Mock;
+describe('GraphQL wrapper for SWR', () => {
+  beforeEach(() => {
+    fetch = vi.fn(() =>
```
himself65, line 13: if we wrap the `fetch` into `fetchAtom`. This could be easy to control the test behavior

himself65: but for now it's ok

tsconfig.json
```diff
@@ -28,6 +28,7 @@
       "@affine/env/*": ["./packages/env/src/*"],
       "@affine/utils": ["./packages/utils"],
       "@affine/workspace/*": ["./packages/workspace/src/*"],
+      "@affine/graphql": ["./packages/graphql/src/index.ts"],
```
himself65:
```suggestion
      "@affine/graphql": ["./packages/graphql/src"],
```

PR #2026, opened 2023-04-19
feat(server): port resolvers to node server

forehalo:
- [ ] user
  - [x] signin
  - [ ] oauth
- [x] workspace
- [x] permission
- [ ] doc


apps/server/src/modules/workspaces/types.ts
```diff
@@ -0,0 +1,6 @@
+export enum Permission {
+  Read = 0,
+  Write = 1,
+  Admin = 10,
```
Brooooooklyn, line 4:
```suggestion
  Admin = 100,
  Owner = 1000,
```

forehalo: The ones migrated from https://github.com/toeverything/OctoBase/blob/74867b7cbb96c3f67d362ac3db84c160d23fa711/libs/cloud-database/src/model.rs#L226

forehalo:
> Error: # src/modules/users/guard.ts(15,14): error TS2339: Property 'user' does not exist on type 'Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>'.

The server test cases complain about invalid access guarded by type system, but the `user` should exist actually by type merging in `apps/server/global.d.ts`. Seems the typechecking system is not working as expect @Brooooooklyn @Himself65

himself65:
> > Error: # src/modules/users/guard.ts(15,14): error TS2339: Property 'user' does not exist on type 'Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>'.
> 
> The server test cases complain about invalid access guarded by type system, but the `user` should exist actually by type merging in `apps/server/global.d.ts`.
> 
> Seems the typechecking system is not working as expect @Brooooooklyn @Himself65, and I didn't find anywhere where there is a typecheck working now in CI.

Fixed. Now you can see the correct error

himself65 (commented):
Do we need asymmetric encryption for auth part?

@Brooooooklyn

Brooooooklyn:
@Himself65 I recommend [argon2](https://github.com/napi-rs/node-rs/tree/main/packages/argon2)

himself65 (commented):
I will handle this today

PR #2037, opened 2023-04-19
feat: store local data to local db

pengx17:
Data will be saved in `/Users/xxxx/Library/Application Support/@affine/electron/workspaces/{workspace_id}/storage.db`.
Each workspace has its own DB file.

CleanShot 2023-04-20 at 00 55 53@2x

some facts:
- still use local workspace to simplify the logic, but adding a new SQLite provider to handling the API calls been render process & main (sqlite)
- on provider connection, it will sync the current ydoc with SQLite (duplex) 

```[tasklist]
### todo
- [x] blob handling - right now blob is only in idb and cloud (as in BlockSuite blob storage). does not allow users to specify blob handling logic
- [x] tests
- [x] ~~sqlite optimization (trim the size of updates)~~ not in scope
- [x] https://github.com/toeverything/AFFiNE/issues/1875
```


Himself65 (commented):
The new blob storage example is here; you can implement it now

https://github.com/toeverything/blocksuite/pull/2169/files#diff-47e0b587214e6f48b210ff793116e9ffccaba5bad2c3fc3eba4fcdb7bb2fe931R5

Himself65 (changes requested):
test needed

Himself65 (changes requested):
at least you need a unit test for new provider

darkskygit (approved):

PR #2042, opened 2023-04-20
docs: add more instruction for yarn

howarddo2208:
I was trying to set up my dev environment and came across a problem with yarn version, so I added more instructions for it for later contributors if they encounter it


himself65 (approved):

himself65:
thanks for the contribution!

Issue #2073, opened 2023-04-22
Command + / can fold the sidebar, but no guide is given.


tzhangchi:
<img width="1727" alt="image" src="https://user-images.githubusercontent.com/5910926/233755374-4871c798-8381-4cf3-a929-cebf6c24f648.png">

PR #2075, opened 2023-04-22
feat: add short cuts for sidebar

tzhangchi:
related to the issues #2073


apps/web/src/components/affine/sidebar-switch/index.tsx
```diff
@@ -1,6 +1,7 @@
+// eslint-disable-next-line simple-import-sort/imports
```
himself65: blocking

Himself65 (changes requested):

Himself65 (approved):

PR #2077, opened 2023-04-22
feat: add shared page empty tip

tzhangchi:


JimmFly (approved):

apps/web/src/components/affine/sidebar-switch/index.tsx
```diff
@@ -1,6 +1,7 @@
+// eslint-disable-next-line simple-import-sort/imports
```
himself65: blocking

Himself65 (changes requested):

Himself65 (approved):

PR #2080, opened 2023-04-22
refactor: use rpc server with MessagePort

himself65:
The lack of old implementation:

1. Logging/Debug
2. Need to write two copy of code `ipcMain.handle` and `ipcMain.revoke`
3. Only support in MainWorld, but in the future we might have many tabs
4. Duplicated and complex code when we need two-way communication

Upstream: https://github.com/Jack-Works/async-call-rpc/issues/56


pengx17 (approved):

pengx17 (changes requested):
Some thoughts:

All of the API calls so far is one-direction that allows the renderer process to invoke functions in the main thread. I think it is a solved problem and we can also get benefit of types with the help of `dts-for-context-bridge`.

I also agree that it is not clean that we have some duplicated codes in preload and in main, but I think it is not worth of adding Jack-Works/async-call-rpc for this sole purpose.

Regarding logging, we are relying on adding manual electron-log logs to the APIs right now, which will also adding logs to files for better diagnoses in production. I believe automatic logging has a limit and we have to add them manually most of the time.

---

IMHO Jack-Works/async-call-rpc may help in cases if we want to have the main thread to notify the specific renderer window. This is not a solved issue yet right? I see your code include some example code in https://www.electronjs.org/docs/latest/tutorial/message-ports.

himself65:
> dts-for-context-bridge

Don't use a 2-year deprecated repp and lazy on type writing

the workflow should be, write type in `env` package and implement them in electron app, instead of auto-generated type

pengx17:
I agree, the auto-generated types are also incorrect in some cases and it seems it is not worth the efforts to maintain it based on its complexity. Are you suggesting manually managing the electron preload types?

himself65:
> I agree, the auto-generated types are also incorrect in some cases and it seems it is not worth the efforts to maintain it based on its complexity. Are you suggesting manually managing the electron preload types?

Also one feature is MessagePort, I think we should use Web standard and forget the main window and other windows

pengx17:
I think when using messagePort we will need more code to do the setup, like managing its lifecycle and the ports for different peers. It seems to be an overkill for now based on our limited time schedule. Shall we have this feature when we build multi-window feature?

himself65:
> I think when using messagePort we will need more code to do the setup, like managing its lifecycle and the ports for different peers. It seems to be an overkill for now based on our limited time schedule. Shall we have this feature when we build multi-window feature?

replace the existing code to the message port, for now, is not hard. Also, multi-window is just an example. I think we should reduce the burden of mind. exportToMianWorld of course is a burden of mind.

pengx17:
We will need to ship more features this week for more features this week. Based on my knowledge so far around the Jack-Works/async-call-rpc I am afraid it does more confusion to me than exportToMainWorld.
I am not pushing back the solution, I just think it is not worth the efforts for now. We shall focus more on the core functions, other than using a different solution to replace existing solved problems.

PR #2082, opened 2023-04-22
refactor: use vite dev server

Himself65:


pengx17:
According to its document, it only reload the preload script but for main thread it still needs to be relaunched for code changes. I think the benefits for us is minimum and not worth to use another tool to add complexity to our code.

Issue #2084, opened 2023-04-22
Potential workspace content lost after page refresh


doodlewind:
### What happened?

This happened rarely after writing piece of text and press `cmd-R` to refresh immediately. When encountered, the workspace in IndexedDB will be cleared. So it's unable to recover the workspace content there.

![505eab8d-51cb-43b9-9c4c-a3179bc7345a](https://user-images.githubusercontent.com/7312949/233805425-ad69122e-6aa0-48e6-9150-d1c01f97554c.jpeg)

![460d0adb-5061-4548-9e04-25e3ee9be7f0](https://user-images.githubusercontent.com/7312949/233805501-7130ff61-d7de-4035-b13f-5dea42c4fc7c.jpeg)

I was using `canary.7`, will upgrade to see the luck.

### Version

macOS ARM 64

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

PR #2085, opened 2023-04-22
fix(y-indexeddb): alert user when write operation unfinished

Himself65:
Related: https://github.com/toeverything/AFFiNE/issues/2084


tzhangchi:
Cool

PR #2122, opened 2023-04-25
feat: create workspace from loading existing exported file

pengx17:
This pull request adds the feature to export and import workspaces as SQLite files, refactors the IPC communication between the main and renderer processes in the Electron app, and fixes some bugs.

### Data flow design

image

### How does adding workspace work behind the scene
image

### How does main registers handlers?
image

New workspace:
<video src="https://user-images.githubusercontent.com/584378/235758141-0c633fee-5cca-4f0c-94dd-b7e0d17ffc03.mp4"></video>

Add workspace (import from existing db file):
<video src="https://user-images.githubusercontent.com/584378/235758129-c8af7436-6ad0-4f8d-b764-93a67b16d007.mp4"></video>

New workspace general setting
a8305f57-684b-4cb8-9a0c-ebc5262e466c


apps/electron/layers/main/src/events/register.ts
```diff
@@ -0,0 +1,28 @@
+import { app,BrowserWindow } from 'electron';
```
Brooooooklyn, line 1: How does this line pass the lint & format?

apps/electron/layers/main/src/exposed.ts
```diff
@@ -0,0 +1,5 @@
+import { allEvents as events } from './events';
+import { allHandlers as handlers } from './handlers';
+
+// this will be used by preload script to expose all handlers and events to the renderer process
+export { events,handlers };
```
Brooooooklyn, line 5: `prettier`?

apps/electron/layers/main/src/events/register.ts
pengx17: No idea 🥵

pengx17:
@Himself65 should be ready for review

apps/electron/layers/main/src/events/register.ts
```diff
@@ -0,0 +1,28 @@
...
+function getActiveWindows() {
+  return BrowserWindow.getAllWindows().filter(win => !win.isDestroyed());
+}
+
+export function registerEvents() {
+  // register events
+  for (const [namespace, namespaceEvents] of Object.entries(allEvents)) {
+    for (const [key, eventRegister] of Object.entries(namespaceEvents)) {
+      const subscription = eventRegister((...args) => {
+        const chan = `${namespace}:${key}`;
```
himself65, line 21: what is `chan`

apps/electron/layers/main/src/handlers/db/ensure-db.ts
```diff
@@ -0,0 +1,89 @@
+import { watch } from 'chokidar';
```
himself65, line 1: This project has not been maintained for a long time. I'm worried about this

https://github.com/paulmillr/chokidar

Why don't we use Node.js API?

apps/electron/layers/main/src/handlers/db/sqlite.ts
```diff
@@ -33,75 +34,100 @@ interface BlobRow {
...
     }
 
     // use cached version?
-    const db = (this.sqliteDB = sqlite(this.path));
+    const db = (this.db = sqlite(this.path));
     db.exec(schemas.join(';'));
 
     if (!this.firstConnect) {
-      this.ydoc.on('update', this.addUpdateToSQLite);
+      this.ydoc.on('update', (update: Uint8Array, origin) => {
```
himself65, line 70: what about `off`?

apps/electron/layers/main/src/events/register.ts
pengx17: chan === channel

apps/electron/forge.config.js
```diff
@@ -45,6 +45,7 @@ module.exports = {
           teamId: process.env.APPLE_TEAM_ID,
         }
       : undefined,
+    // do we need the following line?
```
m1911star, line 48: yes for prod version

apps/electron/layers/main/src/handlers/db/ensure-db.ts
pengx17: I tried to use Node.js api first, but it does not work pretty well.

On the README of chokidar, it listed some issues:

```
Why?
Node.js fs.watch:

Doesn't report filenames on MacOS.
Doesn't report events at all when using editors like Sublime on MacOS.
Often reports events twice.
Emits most changes as rename.
Does not provide an easy way to recursively watch file trees.
Does not support recursive watching on Linux.

Node.js fs.watchFile:

Almost as bad at event handling.
Also does not provide any recursive watching.
Results in high CPU utilization.
Chokidar resolves these problems.
```

I do encounter most of the issues when using either `fs.watch` or `fs.watchFile`. For a more reliable implementation, I used the popular chokidar instead.

BTW do you have any other suggestions over chokidar?
I saw in https://github.com/parcel-bundler/parcel/issues/863 it listed some alternatives like: `watchman`, `fb-watchman` and `nsfw`. Not sure which one is better.

apps/electron/layers/main/src/handlers/db/sqlite.ts
pengx17: I think the listener will be gc'ed after `ydoc` is destoryed, thus there is no need to unlisten it expclitly.

apps/electron/layers/main/src/handlers/db/ensure-db.ts
himself65: Let's move to a better package in the next PR.

Himself65 (approved):

Issue #2039, opened 2023-04-20
Add Onboarding carousel


joebeijing:
（1）Hyper merged whiteboard and docs
Description：Easily switch between Page mode for structured document creation and Whiteboard mode for freeform visual expression of creative ideas.

（2）Intuitive & robust block-based editing
Description：Create structured documents with ease, using a modular interface to drag and drop blocks of text, images, and other content.

xell:
Add slide left/right effects to the carousel, taking the three components, i.e. the heading/title, the image/video, and the description, as a whole to slide by a same speed.

PR #2144, opened 2023-04-26
feat: add onboarding for client

JimmFly:
Close for #2039 

https://user-images.githubusercontent.com/102217452/236111182-75d9360e-abff-4d6e-95b6-3906bb928bab.mov


apps/electron/tests/basic.spec.ts
```diff
@@ -16,6 +16,9 @@ test.beforeEach(async () => {
     colorScheme: 'light',
   });
   page = await electronApp.firstWindow();
+  await page.getByTestId('onboarding-modal-close-button').click({
+    delay: 100,
```
pengx17, line 20: Is `delay` required for every click calls?

apps/web/src/components/pure/OnboardingModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const helperGuide = getHelperGuide();
+    return helperGuide?.onBoarding ?? true;
+  }, []);
+
+  useEffect(() => {
+    if (shouldShow) {
+      setOpenOnboarding(true);
+    }
+  }, [shouldShow, setOpenOnboarding]);
+  if (!env.isDesktop) {
```
pengx17, line 44: I think it is not necessary to check here. You can add a check on where `OnboardingModal` is being used instead and thus this component will not get render at all.

pengx17 (approved):

PR #2182, opened 2023-04-28
test(electron): theme check

Himself65:


apps/electron/tests/basic.spec.ts
```diff
@@ -19,5 +37,30 @@ test('new page', async () => {
...
+  {
+    const themeMode = await root.evaluate(element =>
+      element.getAttribute('data-theme')
+    );
+    expect(themeMode).toBe('dark');
+  }
+  const after = await page.screenshot({
+    path: resolve(testResultDir, 'affine-dark-theme-electron.png'),
+  });
+  expect(prev).not.toEqual(after);
```
pengx17, line 65: The return result of page.screenshot is a Buffer and comparing them will always be false. 
I guess we may want to have a better image diff solution to solve this issue, but I guess we do not have to be part of Electron test suite. This can be done in the storybook or web instead.

pengx17 (approved):

PR #2185, opened 2023-04-28
feat: add mac release zip file and release info yml

m1911star:
image

https://github.com/toeverything/AFFiNE/actions/runs/4829332390

purpose:
we will use electron-updater as a base module to implement auto updater with channel(canary/beta/stable). But we use electron-forge as builder. So we need to generate assets with script

ref: 
- https://www.electron.build/auto-update.html
- https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/README.md


apps/electron/package.json
```diff
@@ -44,6 +45,7 @@
     "electron-window-state": "^5.0.3",
     "esbuild": "^0.17.18",
     "fs-extra": "^11.1.1",
+    "js-yaml": "^4.1.0",
```
pengx17, line 48: I think you did not use this library in generate-yml. Maybe remove it or generate JSON and convert it to yml using this lib instead?

apps/electron/package.json
```diff
@@ -15,6 +15,7 @@
     "make-macos-x64": "electron-forge make --platform=darwin --arch=x64",
     "make-windows-x64": "electron-forge make --platform=win32 --arch=x64",
     "make-linux-x64": "electron-forge make --platform=linux --arch=x64",
+    "generate-release-yml": "zx scripts/generate-yml.mjs",
```
pengx17, line 18: It seems this script does not exist and `generate-yml.js` does not use zx. You can remove it.

apps/electron/package.json
m1911star: will use fs.writeFileSync to write a yml string

apps/electron/package.json
m1911star: removed

.github/workflows/release-desktop-app.yml
```diff
@@ -169,7 +175,17 @@ jobs:
         with:
           name: affine-linux-x64-builds
           path: ./
-
+      - name: Download Artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: release-yml-build-script
+          path: ./
+      - uses: actions/setup-node@v3
```
pengx17, line 183: may use ./.github/actions/setup-node instead

.github/workflows/release-desktop-app.yml
```diff
@@ -169,7 +175,17 @@ jobs:
...
+        with:
+          node-version: 18
+      - name: Generate Release yml
+        run: |
+          BUILD_TYPE=${{ env.BUILD_TYPE }} RELEASE_VERSION=${{ github.event.inputs.version }} node generate-yml.js
```
pengx17, line 188: I think you do not need to set BUILD_TYPE here, which is already available in the workflow globally

.github/workflows/release-desktop-app.yml
m1911star: in the final release step we only need a simple nodejs env, dont need to checkout repo or restore cache, so maybe we dont need to use the exist action

.github/workflows/release-desktop-app.yml
m1911star: not sure whether can access build type in generate-yml.js

m1911star: checked & removed

pengx17 (approved):

PR #2203, opened 2023-04-30
fix(electron): ignore .DS_Store on MacOS

fundon:
Screenshot 2023-04-30 at 9 39 25 AM


pengx17 (approved):
LGTM

pengx17:
Maybe we can also check if the folder contains a storage.db file

PR #2207, opened 2023-05-02
feat(y-indexeddb): cleanup

sxzz:


Himself65 (commented):
Could you please add a test case for the update?

packages/y-indexeddb/src/index.ts
```diff
@@ -274,8 +274,8 @@ export const createIndexedDBProvider = (
       doc.off('update', handleUpdate);
       doc.off('destroy', handleDestroy);
     },
-    cleanup() {
-      // todo
+    async cleanup() {
+      (await dbPromise).delete('workspace', id);
```
himself65, line 287: There might have some edge case.

For example, it should throw an error when the provider is still connecting.

Himself65 (changes requested):

Himself65 (approved):

packages/y-indexeddb/src/index.ts
```diff
@@ -275,6 +281,9 @@ export const createIndexedDBProvider = (
       doc.off('destroy', handleDestroy);
     },
     async cleanup() {
+      if (connect) {
```
sxzz, line 284: I'm a bit confused. Is `connect` for connecting or connected?

himself65: connected

PR #2213, opened 2023-05-03
perf(web): make workspace title button easier to click

SevenOutman:
Make quick search button on workspace title easier to recognize and to click.

Before

https://user-images.githubusercontent.com/8225666/235833728-54d0a075-534f-40ee-b0b6-1f1bdc35c134.mov

After

https://user-images.githubusercontent.com/8225666/235833734-1bf99272-2d87-446d-8e9d-921c7e1e4ecd.mov


Himself65 (approved):
LGTM but /cc @joebeijing

joebeijing:
It looks really good, but there's a small issue. Can we keep the same active area for both the page title and the non-page title? Also, CC @xie to have a look at the visual effects of this part.

xell:
Thanks for the PR. However in fact the All Pages is being reconstructed thoroughly and the title might be removed. We will consider a better solution next week.

SevenOutman:
> Thanks for the PR. However in fact the All Pages is being reconstructed thoroughly and the title might be removed. We will consider a better solution next week.

That's OK! Glad to hear that you're making changes on current designs. I can also contribute implementation of the new design if you're open to it.

Issue #1572, opened 2023-03-14
Client app updater


pengx17:
- update checker
- in update update
- "new version available" notification

May need to think about db migration.

As this issue has been requested multiple times. Maybe we shall increase its priority.

ShortCipher5:
1) Should we consider a "back up your data first" option - auto or manual - before performing the update?

2) I would suggest having some sort of brief update intro with an update notification. That could be a simple change log (general user friendly) that outlines any new features/updates and fixes.

PR #2214, opened 2023-05-03
feat: implement latest version updater for macos

m1911star:
update: 
Only fix some simple styles, want to merge first.
Since I'm not familiar with how to manage svg asset and new ui design, maybe need some help

ref: https://github.com/hwangdev97/AFFiNE-Design-Playground

image

https://user-images.githubusercontent.com/4948120/236255798-79298f04-4311-423c-bbbd-aff9c24b98bc.mov

ref: https://github.com/toeverything/AFFiNE/issues/1572

check strategy:

canary: check only prerelease version and latest-mac.yml file info -> console.log version info
stable: check only release and latest one that contains latest-mac.yml -> console.log version info

for local test:
you need to change version to one lower than the latest canary.20

TODO:

- [x] display ui
- [x] download event
- [x] handle errors


apps/electron/dev-app-update.yml
```diff
@@ -0,0 +1,4 @@
+owner: toeverything
+repo: AFFiNE
+provider: github
```
m1911star, line 3: for local dev testing

apps/electron/resources/app-update.yml
```diff
@@ -0,0 +1,4 @@
+owner: toeverything
```
m1911star, line 1: for prod build

Himself65 (approved):

Issue #2205, opened 2023-04-30
​tool bar out page when the  cursor is at the beginning of the line


aiyou9:
### What happened?

![image](https://user-images.githubusercontent.com/15712822/235361319-0e79e933-b617-4def-8111-15b9b1702bb2.png)

### Version

Windows x64

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

himself65:
looks like zIndex incorrect

PR #2215, opened 2023-05-03
style: fix popover z-index

MichaelYuhe:
Fixes: https://github.com/toeverything/AFFiNE/issues/2234

image
Fix issue #2205 
z-index of the pop-over comoponent should not be smaller than the side bar component.


Himself65 (commented):
I think this is not a good approach, this file is exported from design side I think.

Himself65 (approved):

PR #2222, opened 2023-05-04
fix: reduce pageMeta instance

Himself65:


packages/hooks/src/use-block-suite-page-meta.ts
```diff
@@ -19,30 +21,24 @@ declare module '@blocksuite/store' {
...
+  if (!weakMap.has(blockSuiteWorkspace)) {
+    const baseAtom = atom<PageMeta[]>(blockSuiteWorkspace.meta.pageMetas);
+    weakMap.set(blockSuiteWorkspace, baseAtom);
+    baseAtom.onMount = set => {
       const dispose = blockSuiteWorkspace.meta.pageMetasUpdated.on(() => {
-        setPageMeta(blockSuiteWorkspace.meta.pageMetas);
+        set(blockSuiteWorkspace.meta.pageMetas);
       });
       return () => {
         dispose.dispose();
```
pengx17, line 37: may also need to remove the atom from map after dispose

himself65: i think we don't, because It's weak map. the atom will be auto GC, when blocksuite workspace is deleted

pengx17: You are right, this is not actually needed

PR #2223, opened 2023-05-04
fix: reduce `useState` and `useEffect`

Himself65:


packages/hooks/src/use-block-suite-workspace-page-title.ts
```diff
@@ -1,25 +1,35 @@
 import type { Workspace } from '@blocksuite/store';
-import { assertExists } from '@blocksuite/store';
-import { useEffect, useState } from 'react';
+import { assertEquals, assertExists } from '@blocksuite/store';
+import type { Atom } from 'jotai';
+import { atom, useAtomValue } from 'jotai';
+
+const weakMap = new WeakMap<Workspace, [string, Atom<string>]>();
```
pengx17, line 6: It implies that this map will only cache one atom for a single workspace, but we may want to pair workspace+page id as the key right?

himself65: yeah, the pageId could be a dynamic value

pengx17: How about this? Using a two-layered map instead.

```ts
import type { Workspace } from '@blocksuite/store';
import { assertEquals, assertExists } from '@blocksuite/store';
import type { Atom } from 'jotai';
import { atom, useAtomValue } from 'jotai';

const weakMap = new WeakMap<Workspace, Map<string, Atom<string>>>();

function getAtom(w: Workspace, pageId: string) {
  if (!weakMap.has(w)) {
    weakMap.set(w, new Map());
  }
  const map = weakMap.get(w);
  assertExists(map);
  const baseAtom = atom<string>(w.getPage(pageId)?.meta.title || 'Untitled');
  baseAtom.onMount = set => {
    const disposable = w.meta.pageMetasUpdated.on(() => {
      const page = w.getPage(pageId);
      assertExists(page);
      set(page?.meta.title || 'Untitled');
    });
    return () => {
      disposable.dispose();
    };
  };
  map.set(pageId, baseAtom);
  return baseAtom;
}

export function useBlockSuiteWorkspacePage(
  blockSuiteWorkspace: Workspace,
  pageId: string | null
) {
  const atom = getAtom(blockSuiteWorkspace, pageId || '');
  return useAtomValue(atom);
}

export function useBlockSuiteWorkspacePageTitle(
  blockSuiteWorkspace: Workspace,
  pageId: string
) {
  const titleAtom = getAtom(blockSuiteWorkspace, pageId);
  assertExists(titleAtom);
  return useAtomValue(titleAtom);
}
```

pengx17 (approved):

Himself65 (commented):
```shell
Error: Worker terminated due to reaching memory limit: JS heap out of memory
 ❯ new NodeError node:internal/errors:399:5
 ❯ [kOnExit] node:internal/worker:287:26
 ❯ Worker.<computed>.onexit node:internal/worker:209:20
```

I believe we have a memory leak somewhere

PR #2228, opened 2023-05-04
perf: remove `data-testid` in production

cijiugechu:


Himself65 (approved):

Himself65 (changes requested):
I suggest to enable it in the development environment

Himself65 (approved):

PR #2240, opened 2023-05-05
refactor: use atomWithQuery

himself65:


packages/jotai/src/atomWithQuery.ts
```diff
@@ -0,0 +1,72 @@
...
+  };
+  const setQuery = (newValue: string | null) =>
+    void router.push({
+      pathname: router.pathname,
+      query: {
+        ...router.query,
+        [key]: newValue,
+      },
+    });
+  const strAtom = atom<string | null>(null);
```
pengx17, line 34: Do not use type to name variables

packages/jotai/src/atomWithQuery.ts
```diff
@@ -0,0 +1,72 @@
...
+    get => get(valueAtom),
+    (get, set, update: SetStateActionWithReset<string>) => {
+      const nextValue =
+        typeof update === 'function'
+          ? (update as (prev: string | null) => string | typeof RESET)(
+              get(valueAtom)
+            )
+          : update;
+      if (nextValue === RESET) {
+        set(strAtom, null);
```
pengx17, line 63: Not sure if setting here is necessary, since changing query here should trigger history event and then set its value in the event callback

himself65:
See https://github.com/toeverything/AFFiNE/pull/2342

PR #2241, opened 2023-05-05
fix: remove `useEffect` on router sync with atoms

Himself65:
Fixes: https://github.com/toeverything/AFFiNE/issues/2333


tests/parallels/affine/affine-public-workspace.spec.ts
```diff
@@ -24,6 +24,8 @@ test('enable public workspace', async ({ page, context }) => {
   await clickPublishPanel(page);
   await page.getByTestId('publish-enable-affine-cloud-button').click();
   await page.getByTestId('confirm-enable-affine-cloud-button').click();
+  await page.waitForSelector('v-line');
+  await clickSideBarSettingButton(page);
```
himself65: The new behavior will jump to a detailed page. I think it because we move push out of useEffect. So that it will be called on top, and right before the hooks inside settings page

PR #2267, opened 2023-05-08
feat(electron): app menu item and hotkey for creating new page

SevenOutman:
Press <kbd>Command + N</kbd> on macOS or <kbd>Ctrl + N</kbd> on Windows/Linux to create new page in current workspace

https://user-images.githubusercontent.com/8225666/236797359-7743def1-60de-4c3e-9fff-ebb410245a9b.mov


apps/electron/layers/preload/preload.d.ts
```diff
@@ -3,4 +3,16 @@
...
+     * Adds a listener for "New Page" action in application menu
+     *
+     * @returns A cleanup function that removes the listener when called
+     */
+    onNewPage: (callback: (e: Electron.IpcRendererEvent) => void) => () => void;
+    /**
+     * Enable/disable "New Page" action in application menu
+     */
+    toggleNewPageAction: (enabled: boolean) => void;
+  };
```
himself65: I think this should be moved into apis? 
/cc @pengx17 

BTW, there're more and more APIs on the electron side; I think we should restart MessagePort PR(https://github.com/toeverything/AFFiNE/pull/2080) to separate them

SevenOutman: > The implementation should be refactored after #2122 is merged. Can you @SevenOutman please make changes based on it?

Sure. I'll keep an eye on that PR and update my changes once it's merged.

himself65: It merged now

apps/electron/layers/main/src/application-menu.ts
```diff
@@ -0,0 +1,127 @@
+import { app, Menu } from 'electron';
+
+import { subjects } from './events';
+
+// Unique id for menuitems
+const MENUITEM_NEW_PAGE = 'affine:new-page';
+
+export function createApplicationMenu() {
+  const isMac = process.platform === 'darwin';
```
pengx17, line 9: You can use `isMacOS()` in util.ts

apps/electron/layers/main/src/application-menu.ts
```diff
@@ -0,0 +1,127 @@
...
+    },
+    {
+      role: 'help',
+      submenu: [
+        {
+          label: 'Learn More',
+          click: async () => {
+            // eslint-disable-next-line @typescript-eslint/no-var-requires
+            const { shell } = require('electron');
+            await shell.openExternal('https://electronjs.org');
```
pengx17, line 113: Can you change the URL to https://affine.pro/?

apps/electron/layers/main/src/application-menu.ts
SevenOutman: I've updated it.

apps/electron/layers/main/src/application-menu.ts
SevenOutman: Good idea. Updated the link!

pengx17 (approved):
Works great, thanks!

PR #2275, opened 2023-05-09
refactor(component): input component

Himself65:


packages/component/src/ui/input/index.css.ts
```diff
@@ -0,0 +1,34 @@
...
+  height: heightVar,
+  lineHeight: '22px',
+  padding: '8px 12px',
+  color: 'var(--affine-text-primary-color)',
+  border: '1px solid',
+  borderColor: 'var(--affine-border-color)', // TODO: check out disableColor,
+  backgroundColor: 'var(--affine-white)',
+  borderRadius: '10px',
+  selectors: {
+    '&[data-no-border="true"]': {
```
pengx17, line 21: Do you think we should use styleVariants?

himself65: Fixed

pengx17 (approved):

PR #2299, opened 2023-05-10
fix: should not show open folder if it is not moved

pengx17:
fix https://github.com/toeverything/AFFiNE/issues/2313
fix https://github.com/toeverything/AFFiNE/issues/2312
fix https://github.com/toeverything/AFFiNE/issues/2311


packages/i18n/src/resources/en.json
```diff
@@ -260,6 +260,7 @@
   "Move folder hint": "Select a new storage location.",
   "Storage Folder": "Storage Folder",
   "DB_FILE_INVALID": "Invalid Database file",
+  "FILE_ALREADY_EXISTS": "File already exists",
```
JimmFly, line 263: It might be a good idea to give higher priority to improving the naming conventions for i18n keys, considering the current multitude of approaches.

pengx17: What is the convention btw? `com.affine.xxx`?

pengx17: Let's fix this issue after we got a convention for i18n.

Himself65 (approved):

Himself65 (approved):

JimmFly (approved):

JimmFly (approved):

JimmFly (approved):

PR #2307, opened 2023-05-10
build(electron): add internal release channel

himself65:


packages/env/src/constant.ts
```diff
@@ -1,3 +1,13 @@
+import { z } from 'zod';
+
+export const ReleaseTypeSchema = z.enum([
+  'stable',
+  'beta',
+  'canary',
+  'internal',
```
himself65, line 7: We want to add an 'internal' version for our testing members

himself65: What do you think? @Brooooooklyn

himself65:
See https://github.com/toeverything/AFFiNE/pull/2308

PR #2308, opened 2023-05-10
build(electron): add internal release channel 

himself65:


himself65:
See https://github.com/toeverything/AFFiNE/pull/2309

PR #2329, opened 2023-05-11
feat(electron): use affine native

Brooooooklyn:


apps/electron/layers/main/src/handlers/db/ensure-db.ts
```diff
@@ -1,94 +1,130 @@
...
+import { ts } from '../../utils';
 import type { WorkspaceSQLiteDB } from './sqlite';
 import { openWorkspaceDatabase } from './sqlite';
 
-const dbMapping = new Map<string, Promise<WorkspaceSQLiteDB>>();
-const dbWatchers = new Map<string, () => void>();
+const databaseInput$ = new Subject<string>();
+const terminate$ = app ? fromEvent(app, 'before-quit') : NEVER;
+
+const database$ = databaseInput$.pipe(
```
pengx17, line 38: I think we may still need to have mutliple db's to be watched at the same time soon later, e.g. we have multiple-tab/window and each of them need to watch for db file changes.

Brooooooklyn: The watch logics are not changed at all; the `database$` emits different `workspaceId` from the upstream

apps/electron/layers/main/src/handlers/db/ensure-db.ts
```diff
@@ -1,94 +1,130 @@
...
+  distinctUntilChanged(),
+  groupBy(identity),
+  mergeMap(workspaceDatabase$ =>
+    workspaceDatabase$.pipe(
+      // only open the first db, and emit it to the downstream
```
Brooooooklyn, line 43:
```suggestion
      // only open the first db with the same workspaceId, and emit it to the downstream
```

apps/electron/layers/main/src/handlers/db/ensure-db.ts
Brooooooklyn: See my comment below

pengx17 (dismissed):
flawless

Himself65 (approved):

pengx17 (approved):
LGTM

PR #2328, opened 2023-05-11
feat: add disable function for RouteNavigation

JimmFly:
Close #2316 

https://github.com/toeverything/AFFiNE/assets/102217452/328c5bf0-9541-4746-9199-a7e5013b9cb6

Issue #2316, opened 2023-05-11
back / forward button should be grayed out while there's no action.


ShortCipher5:
These two icons appear to have no functionality

<img width="312" alt="Screenshot 2023-05-11 at 09 56 03" src="https://github.com/toeverything/AFFiNE/assets/4605025/cc01f4d3-a02e-4faf-a76e-27b2e5cac5e1">

ShortCipher5:
Grey out / disable when no actions to perform.

pengx17:
It sorta like a feature request. I did attempt to find a solution before, but there was no good solution to detect the history stack in the Nextjs case. I think we can revisit this issue later.

pengx17:
Fixed by https://github.com/toeverything/AFFiNE/pull/2336

PR #2336, opened 2023-05-11
feat(electron): track router history

Himself65:
Fixes: https://github.com/toeverything/AFFiNE/pull/2328, https://github.com/toeverything/AFFiNE/issues/2316

Upstream: https://github.com/toeverything/AFFiNE/pull/2342


apps/web/src/atoms/history.ts
```diff
@@ -0,0 +1,87 @@
...
+          return {
+            stack: newStack,
+            current: newStack.length - 1,
+            skip: false,
+          };
+        }
+      }
+    });
+  };
+
```
pengx17, line 54: I think we need to put one item onto the stack by default.

apps/web/src/components/root-app-sidebar/index.tsx
```diff
@@ -97,10 +98,22 @@ export const RootAppSidebar = ({
   }, [sidebarOpen]);
 
   const clientUpdateAvailable = useAtomValue(updateAvailableAtom);
+  const [history, setHistory] = useAtom(historyAtom);
+  const router = useMemo(() => {
+    return {
+      forward: () => {
```
pengx17, line 104: We may also want to add `allowForward` & `allowBackward` here that to be consumed by the buttons to control the usibility.

pengx17 (approved):

PR #2342, opened 2023-05-12
fix: correct router logic

Himself65:
Now, we only listen to the `routeChangeStart` and don't case other things.


apps/web/src/hooks/__tests__/use-recent-views.spec.tsx
```diff
@@ -0,0 +1,139 @@
+/**
```
himself65, line 1: I moved this test into a new file because I found that next-router-mock has some corner case which might not isolate with other test case

packages/workspace/src/atom.ts
```diff
@@ -32,12 +33,49 @@ export const rootCurrentWorkspaceIdAtom = atomWithStorage<string | null>(
...
 export const rootCurrentPageIdAtom = atomWithStorage<string | null>(
   'root-current-page-id',
   null,
   createJSONStorage(() => sessionStorage)
 );
 
+rootCurrentPageIdAtom.onMount = set => {
+  if (typeof window !== 'undefined') {
+    const callback = (url: string) => {
+      const value = url.split('/')[3];
```
himself65, line 64: hard code here, but it's ok because we don't have other use cases

apps/web/src/layouts/workspace-layout.tsx
```diff
@@ -137,13 +134,21 @@ export const AllWorkspaceContext = ({
...
-  useSyncRouterWithCurrentPageId(router);
-  useRouterWithWorkspaceIdDefense(router);
   const metadata = useAtomValue(rootWorkspacesMetadataAtom);
   const exist = metadata.find(m => m.id === workspaceId);
+  const router = useRouter();
+  const push = router.push;
+  useEffect(() => {
+    const id = setTimeout(() => {
+      if (!exist) {
+        void push('/');
```
himself65, line 145: This avoids entering a nonexisting workspace. we will fallback to the index page

apps/web/src/providers/modal-provider.tsx
```diff
@@ -152,7 +152,7 @@ export function Modals() {
               setOpenCreateWorkspaceModal(false);
               setOpenWorkspacesModal(false);
               setCurrentWorkspace(id);
-              return jumpToSubPath(id, WorkspaceSubPath.SETTING);
+              return jumpToSubPath(id, WorkspaceSubPath.ALL);
```
pengx17, line 155: I think this is a requirement for creating new workspace that we will redirect to the settings pages after creation.

PR #2390, opened 2023-05-16
chore: add responsive styles for workspace card

### LINKED ISSUE #2384: The popup window of My Workspaces doesn't perform as expected in narrow mode
CleanShot 2023-05-16 at 11 30 41@2x

JimmFly:
Close #2384 

https://github.com/toeverything/AFFiNE/assets/102217452/81ba5677-5d10-4b00-8271-09e3cf1f6d41

### DIFF
--- a/packages/component/src/components/workspace-card/styles.ts
+++ b/packages/component/src/components/workspace-card/styles.ts
@@ -61,6 +61,9 @@ export const StyledCard = styled('div')<{
         pointerEvents: 'auto',
       },
     },
+    '@media (max-width: 720px)': {
+      width: '100%',
+    },
   };
 });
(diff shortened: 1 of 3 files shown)


packages/component/src/components/workspace-card/styles.ts
```diff
@@ -61,6 +61,9 @@ export const StyledCard = styled('div')<{
         pointerEvents: 'auto',
       },
     },
+    '@media (max-width: 720px)': {
```
pengx17, line 64: I think we can still use something like `[breakpoints.down('sm', true)]`

pengx17 (approved):

PR #2400, opened 2023-05-16
chore: disable image modal by default

pengx17:


apps/web/preset.config.mjs
```diff
@@ -20,7 +20,7 @@ export const blockSuiteFeatureFlags = {
 export const buildFlags = {
   enableImagePreviewModal: process.env.ENABLE_IMAGE_PREVIEW_MODAL
     ? process.env.ENABLE_IMAGE_PREVIEW_MODAL === 'true'
-    : true,
+    : false,
```
himself65: No, you should enable it by default, but disable it in the CI

Himself65 (changes requested):

apps/web/preset.config.mjs
pengx17: Do you mean that we just add a new environment variable in GitHub settings?

pengx17: Oh you mean update the workflow files directly?

Himself65 (approved):

Issue #2360, opened 2023-05-15
The link for "What's new" for the Beta version has not been modified


joebeijing:
@JackiePengp Please provide a link for Beta changelog.

And please note that the AFFiNE Canary version still redirects to the GitHub Release page.

PR #2401, opened 2023-05-16
feat: update button enhancements

pengx17:
fix https://github.com/toeverything/AFFiNE/issues/1594 #1572 #2360

New updater logic
1. Check for new updates every certain period of time (10 minutes), or check for updates through menu->help.
2. If there is a new version available, it will automatically start downloading and display the download progress here.
3. After the download is complete, a prompt will appear saying "update available". Clicking on it will restart the application and update it.
4. When entering the new version, if it is found that the user has not viewed the update log for this version, "what's new" will be displayed.

For Windows/Linux, steps 1 and 4 are consistent except that they directly display "Open download page".

Added check for updates item under help menu category.

- when you never open or dismiss a changelog for current version
image

- when the new version is ready
image

- when downloading a new version

image


packages/component/src/components/app-sidebar/app-updater-button/index.tsx
```diff
@@ -1,35 +1,111 @@
...
   style?: React.CSSProperties;
 }
 
+const currentVersionAtom = atom(async () => {
+  const currentVersion = await window.apis?.updater.currentVersion();
+  return currentVersion;
+});
+
+const currentChangelogUnreadAtom = atom(async get => {
+  assertExists(window.appInfo);
```
himself65: unused

packages/component/src/components/app-sidebar/app-updater-button/index.tsx
```diff
@@ -1,35 +1,111 @@
...
+import { changelogCheckedAtom, updateReadyAtom } from '../index.jotai';
 import * as styles from './index.css';
 
 interface AddPageButtonProps {
   className?: string;
   style?: React.CSSProperties;
 }
 
+const currentVersionAtom = atom(async () => {
+  const currentVersion = await window.apis?.updater.currentVersion();
```
himself65, line 24: You assume this is a `no ssr` atom, but in another atom, you check no ssr. There may be consistent problems here.

packages/component/src/components/app-sidebar/app-updater-button/index.tsx
```diff
@@ -1,35 +1,111 @@
...
+  useEffect(() => {
+    window.apis?.updater.checkForUpdatesAndNotify();
+  }, []);
+
+  const currentChangelogUnread = useAtomValue(currentChangelogUnreadAtom);
+  const updateReady = useAtomValue(updateReadyAtom);
+  const currentVersion = useAtomValue(currentVersionAtom);
+  const onReadOrDismissChangelog = useSetAtom(changelogCheckedAtom);
+
+  const [, transition] = useTransition();
```
himself65: use `startTransition` instead if you don't need isTransiting

https://react.dev/reference/react/startTransition

packages/component/src/components/app-sidebar/app-updater-button/index.tsx
```diff
@@ -1,35 +1,111 @@
...
+  }
+  return false;
+});
+
 // Although it is called an input, it is actually a button.
 export function AppUpdaterButton({ className, style }: AddPageButtonProps) {
   const t = useAFFiNEI18N();
 
+  useEffect(() => {
+    window.apis?.updater.checkForUpdatesAndNotify();
```
himself65: What about `window.apis` being undefined in the first render?

packages/component/src/components/app-sidebar/app-updater-button/index.tsx
```diff
@@ -1,35 +1,111 @@
...
+  const onReadOrDismissCurrentChangelog = (visit: boolean) => {
+    if (visit) {
+      window.open('https://github.com/toeverything/AFFiNE/releases', '_blank');
+    }
+
+    transition(() =>
+      onReadOrDismissChangelog(mapping => {
+        return {
+          ...mapping,
+          [currentVersion!]: true,
```
himself65, line 63: Does `!` necessary here?

pengx17: Yeah the type inferred from `currentVersionAtom` is `string|undefined`. We could also use assertsExist to bypass it though.

packages/component/src/components/app-sidebar/app-updater-button/index.tsx
pengx17: windows.apis will always be non-undefined in desktop, so I guess it should be fine. The windows.api.? check is a workaround for web only version.

packages/component/src/components/app-sidebar/app-updater-button/index.tsx
pengx17: I think since this atom is async, ssr should be fine because aync atom in SSR will always block rendering right?

himself65: actually no, suspense is not a guarantee of no ssr. Here it works just because upstream do the no ssr job.

The sever will response html with suspense in the first render if it did throw a promise(`onShellReady`)

Please read https://react.dev/reference/react-dom/server/renderToPipeableStream for more details

Issue #2381, opened 2023-05-16
Client app infinite reloading issue


pengx17:
This happens in the latest version. When there are 'big' workspaces stored, it will cause the app to reload infinitely.

PR #2405, opened 2023-05-17
fix: infinite reloading

Himself65:
Fixes: https://github.com/toeverything/AFFiNE/issues/2381


packages/workspace/src/atom.ts
```diff
@@ -28,11 +28,7 @@ export const rootWorkspacesMetadataAtom = atomWithStorage<
 );
 
 // two more atoms to store the current workspace and page
-export const rootCurrentWorkspaceIdAtom = atomWithStorage<string | null>(
```
himself65, line 31: storage is unnecessary

pengx17 (approved):

PR #2408, opened 2023-05-17
build: fix electron release build process

Brooooooklyn:


himself65 (approved):

himself65 (dismissed):

himself65:
https://github.com/toeverything/AFFiNE/actions/runs/5000536889/jobs/8958120512

himself65:
Error: 'macos' is an invalid platform. Choices are 'darwin', 'mas', 'win32' or 'linux'.

pengx17:
The reason of using macos instead of darwin is that normal user does not know what darwin is when looking for downloads in the download page :D

Brooooooklyn:
> The reason of using macos instead of darwin is that normal user does not know what darwin is when looking for downloads in the download page :D

normal users should come from our home page rather than the GH releases page

pengx17 (dismissed):
Please double check if the new naming rules still work in https://affine.pro/download

pengx17 (approved):

PR #2420, opened 2023-05-17
fix: add workflow to check version match with package.json

m1911star:
avoid version dismatch


apps/electron/scripts/generate-assets.mjs
```diff
@@ -19,9 +20,20 @@ console.log('build with following dir', {
   publicAffineOutDir,
 });
 
+// step 0: check version match
+const electronPackageJson = await fs.readFile(
```
Brooooooklyn, line 24: `import()` it?

m1911star: use import will cause ERR_IMPORT_ASSERTION_TYPE_MISSING error, can we set type=module in package.json?

m1911star: done

Brooooooklyn (approved):

Himself65 (approved):

PR #2422, opened 2023-05-17
feat: fav page references

pengx17:
https://github.com/toeverything/AFFiNE/assets/584378/a48b355f-906c-4942-8695-21f9d56415c9

### DIFF
--- a/packages/hooks/src/use-block-suite-page-references.ts
+++ b/packages/hooks/src/use-block-suite-page-references.ts
@@ -1,0 +1,38 @@
+import type { Page, Workspace } from '@blocksuite/store';
+import { atom, useAtomValue } from 'jotai';
+import { atomFamily } from 'jotai/utils';
+
+import { useBlockSuiteWorkspacePage } from './use-block-suite-workspace-page';
+
+function getPageReferences(page: Page): string[] {
+  // todo: is there a way to use page indexer to get all references?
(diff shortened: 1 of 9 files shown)


packages/hooks/src/use-block-suite-page-references.ts
```diff
@@ -0,0 +1,38 @@
+import type { Page, Workspace } from '@blocksuite/store';
+import { atom, useAtomValue } from 'jotai';
+import { atomFamily } from 'jotai/utils';
+
+import { useBlockSuiteWorkspacePage } from './use-block-suite-workspace-page';
```
pengx17, line 5: I just realized that the way we wrote in `useBlockSuiteWorkspacePage` using weakmap can be replaced by `atomFamily` utility

himself65: yeap

himself65: I want to make it simple so write it manually

Himself65 (commented):
Could please add coverage for this line?

https://app.codecov.io/gh/toeverything/AFFiNE/pull/2422/blob/packages/component/src/components/app-sidebar/menu-item/index.tsx#L48

Himself65 (approved):

PR #2434, opened 2023-05-18
doc: update build guideline and add dev:local for cli

fourdim:


packages/cli/package.json
```diff
@@ -3,7 +3,8 @@
   "type": "module",
   "private": true,
   "bin": {
-    "dev-web": "./src/dev.mjs"
+    "dev-web": "./src/dev.mjs",
+    "dev-web:local": "./src/dev-local.mjs"
```
himself65: does this script necessary?

fourdim: Not that necessary, but it would be convient for someone (like me) to create a task for it to auto relaunch the application (as it does not need any interactive selection).

himself65: I'd like to remove it? I think we could add instruct in the document instead here

fourdim: done.

Himself65 (approved):

Himself65 (approved):

Himself65 (approved):

PR #2449, opened 2023-05-18
feat: draft daily note

himself65:


himself65:
See https://github.com/toeverything/AFFiNE/pull/2853

PR #2497, opened 2023-05-23
feat(native): move sqlite operation into Rust

Brooooooklyn:
Use `sqlx` to replace the `better-sqlite3`


pengx17:
Does this mean when using rusqlite, all operations will be async?

Brooooooklyn:
> Does this mean when using rusqlite, all operations will be async?

You can choose which ones are async and which ones are sync, for those small operations sync is enough

apps/electron/layers/main/src/__tests__/integration.spec.ts
```diff
@@ -1,5 +1,6 @@
 import assert from 'node:assert';
 import path from 'node:path';
+import { setTimeout } from 'node:timers/promises';
```
pengx17, line 3: TIL

apps/electron/layers/main/src/db/secondary-db.ts
```diff
@@ -74,13 +67,12 @@ export class SecondaryWorkspaceSQLiteDB extends BaseSQLiteAdapter {
 
   // wrap the fn with connect and close
   // it only works for sync functions
-  run = (fn: () => void) => {
+  run = async (fn: () => Promise<void>) => {
     try {
       if (this.runCounter === 0) {
-        this.connect();
```
pengx17, line 80: `run` was designed so that db operations inside will release DB connection after use. I am not sure where is secondary DB is connected after the change?

PR #2568, opened 2023-05-28
fix: dropdown menu entire right can be pulled down

### LINKED ISSUE #2545: New page option - all pages - dropdown button issue
The drop down arrow interaction area is too small.

The whole right side of the button should activate the extra options - but the interaction box is only for the down icon.

The red area will create a new page directly and not trigger the dropdown menu.

Screenshot 2023-05-26 at 16 26 20

HydenLiu:
fix #2545

### DIFF
--- a/packages/component/src/ui/button/dropdown.tsx
+++ b/packages/component/src/ui/button/dropdown.tsx
@@ -8,27 +8,24 @@ import {
 import * as styles from './styles.css';
 
 type DropdownButtonProps = {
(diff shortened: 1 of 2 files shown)


lawvs (approved):
Good work! I have some minor adjustments, mostly for consistency, so I made the changes and pushed them. Thanks a lot for your contribution

Himself65 (approved):

PR #2585, opened 2023-05-30
feat: init support for multiple tiles

Himself65:


apps/web/src/components/page-detail-editor.css.ts
```diff
@@ -0,0 +1,5 @@
+import { globalStyle } from '@vanilla-extract/css';
+
+globalStyle('.mosaic.mosaic-blueprint-theme', {
+  backgroundColor: 'var(--background-color)',
```
pengx17, line 4: Is this a valid css var?

pengx17 (approved):

apps/web/src/components/page-detail-editor.css.ts
himself65: yes

PR #2589, opened 2023-05-30
fix: optimize DB pull

pengx17:
When SQLite DB is pulled, its updates will be applied to a local yDoc. This is a blocking operation and surprisingly will also block the renderer process. See https://twitter.com/NumaanAshraf/status/968496732278374400

This PR proposed a fix that it will offload merge updates in a worker thread and then only apply a single update in the main process, without blocking the UI for too long.

This shall be a temporary solution. We may think about moving all DB operations off main process later.

Also added some tests for this fix


apps/electron/layers/main/src/workers/merge-update.worker.ts
```diff
@@ -0,0 +1,15 @@
+import { parentPort, workerData } from 'node:worker_threads';
+
+import * as Y from 'yjs';
+
+function getMergeUpdate(updates: Uint8Array[]) {
```
Brooooooklyn, line 5: Does `SharedArrayBuffer` work here?

pengx17: Should do, let me have a try

apps/electron/layers/main/src/workers/index.ts
```diff
@@ -0,0 +1,35 @@
...
+      const buffer = new SharedArrayBuffer(update.byteLength);
+      const view = new Uint8Array(buffer);
+      view.set(update);
+      return view;
+    });
+
+    const worker = new Worker(workerFile, {
+      workerData: sharedArrayBufferUpdates,
+    });
+
```
himself65, line 26: use async-call-rpc with the worker. I will handle this part

Himself65 (changes requested):

apps/electron/layers/main/src/workers/index.ts
pengx17: Will you work on it in this PR or perhaps in a following one?

Himself65 (approved):

PR #2617, opened 2023-05-31
chore: update filter style

JimmFly:
image


apps/web/src/components/blocksuite/workspace-header/styles.css.ts
```diff
@@ -149,7 +149,24 @@ export const pageListTitleWrapper = style({
...
   alignItems: 'center',
 });
-
+export const allPageListTitleWrapper = style({
+  fontSize: 'var(--affine-font-base)',
+  color: 'var(--affine-text-primary-color)',
+  display: 'flex',
+  justifyContent: 'center',
+  alignItems: 'center',
+  '::after': {
```
pengx17, line 158: Is this valid? I mean it is not inside of `selectors`

himself65: I think it invalid

Himself65 (commented):
update build flag

apps/web/src/components/blocksuite/workspace-header/styles.css.ts
himself65: it valid

Himself65 (approved):

PR #2639, opened 2023-06-01
refactor: move handlers into bookmark block plugin

himself65:


himself65 (commented):
In the future, we need a better way to handle this

1. isolated env
2. permission check
3. pure ESM

PR #2640, opened 2023-06-01
perf: isMacOS() -> isMac

Simon-He95:


apps/electron/layers/main/src/utils.ts
```diff
@@ -2,10 +2,10 @@ export function getTime() {
   return new Date().getTime();
 }
 
-export const isMacOS = () => {
+export const isMac = (() => {
```
Brooooooklyn, line 5:
```suggestion
export const isMac = process.platform === 'darwin';
```

Simon-He95: Wow

himself65:
We are going to remove the env const in the future. I'd like to close this for now

PR #2643, opened 2023-06-01
build: prevent tsconfig includes sources outside

Brooooooklyn:


packages/component/tsconfig.json
```diff
@@ -1,18 +1,18 @@
...
-    "./src/**/*.json",
-    "../workspace/src",
-    "../../apps/electron/layers/**/src"
+  "exclude": [
+    "lib",
+    "./src/components/page-list/filter/shared-types.tsx",
+    "./src/components/page-list/filter/logic/custom-type.ts",
+    "./src/components/page-list/filter/logic/matcher.ts",
+    "./src/components/page-list/filter/logic/typesystem.ts"
   ],
```
himself65, line 9: does this mean page-list is  actually a `@affine/workspace` component?

Brooooooklyn: No, it's also used by `@affine/component` itself

pengx17 (approved):

PR #2656, opened 2023-06-01
chore: bump typescript to `5.1.3`

himself65:


himself65:
See https://github.com/toeverything/AFFiNE/pull/2735

PR #2660, opened 2023-06-02
feat: plugin system with isolated bundles

Himself65:
Upstream: https://github.com/toeverything/AFFiNE/pull/2654


apps/electron/layers/main/src/ui/index.ts
```diff
@@ -1,10 +1,17 @@
...
+
 import { app, BrowserWindow, nativeTheme } from 'electron';
 
 import type { NamespaceHandlers } from '../type';
 import { isMacOS } from '../utils';
-import { getMetaData } from './get-meta-data';
 import { getGoogleOauthCode } from './google-auth';
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const handlers = require(join(
```
himself65, line 10: need a better way to inject plugin handlers

pengx17 (approved):

PR #2675, opened 2023-06-03
poc: enable auto start on login

m1911star:


m1911star:
current master build crashes on start.

pengx17:
> current master build crashes on start. image

It looks like you need to run `yarn install` once.

PR #2680, opened 2023-06-04
refactor: use worker_thread on db

himself65:


apps/electron/layers/main/src/db/ensure-db.ts
pengx17, line 1: I think since `ensureSQLiteDB` runs in another worker thread, it will not be able to access the same context with `getWorkspaceMeta$` because the latter is still in the main thread.

pengx17:
I think maybe another solution here is completely move all handlers (except the UI ones) off the main process so that we do not worry about cross thread/process access.
The easiest solution could be wrapping all the handlers in a headless browser window process and let the primary renderer process directly talk to the headless browser window instead of having another layer of data transformation.

Also I finally got how the async-call-rpc works and it looks amazing. Maybe it only require only a few dozen of lines to migrate existing in-house solution to that.

PR #2682, opened 2023-06-04
feat: init `window.affine`

Himself65:


pengx17:
Could you explain the reason why exposing plain IPCs here instead of using an RPC abstraction?

Issue #2671, opened 2023-06-02
add workspace may not be able to load DB correctly​


pengx17:
### What happened?

When adding a exported workspace db file, the db contents are not being loaded. Sometimes there will only be the default "preload" page. 

### Version

macOS ARM 64

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

SolarDesalination:
I also have this issue. When I load my old storage.db file into the Affine.pro stable version on Linux, it doesn't show the underlying data. It only shows the default page. I know that the data is there because the DB is 7mb and the default DB is less.

PR #2685, opened 2023-06-05
fix: import workspace may only show default preload page

pengx17:
fix https://github.com/toeverything/AFFiNE/issues/2671

Thanks for @thorseraq for helping me debug the issue.
The steps in the red area should be skipped for adding workspace.
image


Himself65 (changes requested):
not a good approach

Himself65 (approved):

PR #2693, opened 2023-06-06
fix(native): secondary db polling logic

pengx17:
It seems the our Rust wrapper for SQLite is not working properly.


apps/electron/layers/main/src/db/ensure-db.ts
```diff
@@ -82,10 +105,7 @@ function getWorkspaceDB$(id: string) {
 
 function startPollingSecondaryDB(db: WorkspaceSQLiteDB) {
```
pengx17, line 106: I refactored how polling works

PR #2709, opened 2023-06-07
fix(electron): should not continue pull when db closed

pengx17:
After moving the database, a meta change was triggered. At this time, the main database was destroyed, but the sync process of the external database was still ongoing.

### DIFF
--- a/apps/electron/layers/main/src/db/__tests__/ensure-db.spec.ts
+++ b/apps/electron/layers/main/src/db/__tests__/ensure-db.spec.ts
@@ -78,6 +78,10 @@ beforeEach(() => {
 
 afterEach(async () => {
   await runHandler('before-quit');
+  // wait for the db to be closed on Windows
+  if (process.platform === 'win32') {
+    await setTimeout(200);
+  }
   await fs.remove(tmpDir);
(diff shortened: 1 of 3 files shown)


apps/electron/layers/main/src/db/__tests__/ensure-db.spec.ts
```diff
@@ -78,6 +78,10 @@ beforeEach(() => {
 
 afterEach(async () => {
   await runHandler('before-quit');
+  // wait for the db to be closed on Windows
+  if (process.platform === 'win32') {
+    await setTimeout(200);
+  }
```
himself65, line 84: Can we remove this?

pengx17: I think we can, like repeatedly check if the file is removed or not, but it may not worth doing so right now.

himself65: This is a magic code, add comment or do the right thing

pengx17: There is already a comment above.

PR #2743, opened 2023-06-09
feat: add kalam font

pengx17:
image


Himself65 (commented):
Is there already npm bundle, and we can import it in _app.tsx

pengx17:
> Is there already npm bundle, and we can import it in _app.tsx

Do you mean importing? https://github.com/toeverything/blocksuite/blob/955ad21051552a83b196db75f8d7da67f864a18a/packages/editor/themes/affine.css#L2

The CSS rule here uses a dynamic css import rule will is considered a bad practice.
Also, directly embedding the font rules in the bundle for faster load for desktop app.

PR #2746, opened 2023-06-09
fix: electron dev crash

pengx17:


Himself65 (approved):

apps/electron/layers/main/src/logger.ts
```diff
@@ -3,6 +3,10 @@ import log from 'electron-log';
 
 export const logger = log;
 
+process.on('exit', () => {
+  log.transports.console.level = false;
```
himself65: Does this just a hacky way? I don't think ignore everything is a good approach

pengx17: I think it should be ok. This only mutes logging for dev mode. On production it will write to files.

PR #2753, opened 2023-06-11
feat: add helper process

### LINKED ISSUE #2614: Move DB handler implementations off main process
We found that some of the DB operations in the main process may also blocks renderer process. We[ used workers to move out one of the heavy operations,](https://github.com/toeverything/AFFiNE/pull/2589/files) but some actions may still blocks the renderer.

pengx17:
fix https://github.com/toeverything/AFFiNE/issues/2614

The goal is to have another helper process using Electron's [new UtilityProcess API](https://github.com/electron/electron/pull/34980) to offload the blocking APIs (workspace, DB, dialog).

In order to do so, we need to have a different solution than the existing IPC based solutions. Here in this PR we introduced a MessagePort enabled bi-directional communication between UtilityProcess and renderer, and abstract the API with [async-rpc-call](https://github.com/Jack-Works/async-call-rpc) recommended by @Himself65.

The issues
- since we added another peer (helper) that we have a total of three now, the communication and initialization logic between each peer shall be dealt with differently.
  - we need three pairs of bidirectional RPC now between `main`, `helper` and `renderer`
  - `main` and `helper` shall be initialized earlier than `renderer` and will pass the api meta to `renderer`
  - both `main` and `helper` will expose APIs to `renderer`, but with different scopes (mostly, helper cannot invoke electron related APIs)
- there is no good solution for PUSH notifications based on JSON-RPC/async-call-rpc yet. We have to hack it around. See `rpc.postEvent`
- types are mostly ignored. We shall revisit it later.

The general design:
image


himself65 (approved):

himself65:
LGTM

PR #2757, opened 2023-06-12
feat: new setting entry

QiShaoXuan:


apps/web/src/components/affine/setting-modal/style.css.ts
```diff
@@ -0,0 +1,33 @@
...
+  marginTop: '-80px',
+  fontSize: 'var(--affine-font-sm)',
+  display: 'flex',
+});
+globalStyle(`${settingContent} .footer > svg`, {
+  fontSize: 'var(--affine-font-base)',
+  color: 'var(--affine-icon-color)',
+  marginRight: '12px',
+  marginTop: '2px',
+});
```
lawvs, line 33: I don't think `globalStyle` should be abused.

QiShaoXuan: I don't think they are abused, they are use in a custom limited scope and won't pollute other component

PR #2764, opened 2023-06-13
fix: add @typescript-eslint/no-floating-promises rule

Brooooooklyn:


Himself65 (approved):

apps/web/src/pages/workspace/[workspaceId]/setting.tsx
```diff
@@ -43,56 +43,60 @@ function useTabRouterSync(
...
+        })
+        .catch(err => {
+          console.error(err);
+        });
     },
     [router, setCurrentTab]
   );
-  useTabRouterSync(router, currentTab, setCurrentTab);
+
+  useTabRouterSync(router, currentTab, setCurrentTab).catch(err => {
```
Comment (Brooooooklyn) [inline apps/web/src/pages/workspace/[workspaceId]/setting.tsx:97]: @Himself65 this async function is called in render function, which is really weird

packages/component/src/components/page-list/operation-menu-items/export.tsx
```diff
@@ -123,7 +130,9 @@ const ExportToMarkdownMenuItem = ({
     if (!contentParserRef.current) {
       contentParserRef.current = new ContentParser(currentEditor.page);
     }
-    contentParserRef.current.exportMarkdown();
+    contentParserRef.current.exportMarkdown().catch(err => {
+      console.error(err);
+    });
     onSelect?.({ type: 'markdown' });
```
Brooooooklyn, line 136: should this `onSelect` be called in the `contentParserRef.current.exportMarkdown().then`?

apps/web/src/pages/workspace/[workspaceId]/setting.tsx
Comment (himself65) [reply apps/web/src/pages/workspace/[workspaceId]/setting.tsx]: I would like to wrap it into swr I think

packages/component/src/components/page-list/operation-menu-items/export.tsx
himself65: it will block the ui since export markdown is a heavy computing fn

PR #2773, opened 2023-06-14
feat(server): auth server

Brooooooklyn:
Auth server

Prisma schema is from https://authjs.dev/reference/adapter/prisma

GraphQL FileUpload reference: https://gqlgen.com/reference/file-upload/


apps/server/src/modules/storage/fs.ts
```diff
@@ -0,0 +1,32 @@
...
+  constructor(private readonly config: Config) {}
+
+  async writeFile(key: string, file: FileUpload) {
+    const dest = this.config.objectStorage.fs.path;
+    await mkdir(dest, { recursive: true });
+    let crc = 0;
+    let buffer = Buffer.alloc(0);
+    await finished(
+      file.createReadStream().on('data', chunk => {
+        crc = crc32(chunk, crc);
```
darkskygit, line 22: key is sha256 generate by content, i think no extra crc32 is needed?

Brooooooklyn: It's a file with a long hash; not a sub dir, so I think it's okay for self-host users.

himself65 (approved):
LGTM

PR #2777, opened 2023-06-14
feat: export page as file

### LINKED ISSUE #2778: Export page to pdf/png in affine
The exporting feature is implemented here for sevaral reasons:

First,  exporting PDF can be done using the ‘webContents.printToPDF’ method  in electron, which is difficult in js.

Second, the use of another instance does not affect the current window instance and does not block the user's operation.

Thirdly, there is no way to specify the download directory in advance in js.

Finally,  prepare for the future implementation of exporting the entire workspace.

image

Conclusion of the discussion with @pengx17 :
There are imperfections in the current technical solution, so for the time being we are going to deal with a simple solution.

zuoxiaodong0815:
close #2778


apps/electron/src/main/export/preload.ts
pengx17, line 1: Can you check https://github.com/toeverything/AFFiNE/blob/master/apps/electron/src/preload/affine-apis.ts can see how did the existing Electron code manage RPC?

apps/electron/src/main/export/export.ts
```diff
@@ -0,0 +1,297 @@
...
+        if (err) {
+          console.log(err);
+        } else {
+          console.log('PDF Generated Successfully');
+        }
+      });
+      return filePath;
+    } else {
+      const finalFilePath = await win.webContents.executeJavaScript(`
+        const savePageToPng = async () => {
```
pengx17, line 170: The name of the function is should be savePageToPDF?

packages/component/src/components/page-list/operation-menu-items/export.tsx
```diff
@@ -89,33 +101,57 @@ const ExportToHtmlMenuItem = ({
...
+        currentWorkspaceId || '',
+        currentPageId || '',
+        (currentEditor.page.root as PageBlockModel).title.toString(),
+        currentEditor.mode
+      )
+      .then(result => {
+        if (result !== undefined) {
+          return;
+        }
+        return contentParser.exportPng();
```
pengx17, line 131: Does this mean this feature can work without creating a new window? If if it is the case, I think we can just do it purely in here instead of creating a window?

pengx17 (changes requested):
I went over the implementation and it seems the export PNG workflow is like this:
image

think the core logic here is using `transPageToCanvas` and call toDataUrl to convert to png.
So what is the necessity of creating a window to export the png? I

apps/electron/src/main/export/export.ts
```diff
@@ -0,0 +1,297 @@
...
+                resolve('true');
+              }
+              count++;
+              if (count > 10 * 60) {
+                clearInterval(checkReactRender);
+                resolve('false');
+              }
+            }, 100);
+          })
+        `);
```
himself65: Dont run plain text

zuoxiaodong0815:
This technical solution is still a bit problematic and has been discarded for the time being.

PR #2784, opened 2023-06-15
fix: electron cannot be started in Windows

JimmFly:


apps/electron/package.json
```diff
@@ -67,6 +67,7 @@
     "lodash-es": "^4.17.21",
     "nanoid": "^4.0.2",
     "rxjs": "^7.8.1",
+    "which": "^3.0.1",
```
himself65: why you need this?

Himself65 (changes requested):

apps/electron/package.json
JimmFly: I added this in order for spawn to find yarn. I've added a screenshot of the error I'm getting in the description of the pr.

himself65: this package should in dev deps

Himself65 (approved):

PR #2808, opened 2023-06-16
feat(storage): binding jwst storage to node

forehalo:
We port jwst storage as workspace persistence method, so we can read/write any storage backend later on, for example db or s3.


Brooooooklyn (approved):

Brooooooklyn (approved):

himself65 (approved):

himself65:
You dont have to update to latest branch manually. Merge group will handle it automatically

Issue #2798, opened 2023-06-16
SQLite provider should support subdocument​


pengx17:
SQLite provider will not work if blocksuite uses subdocs

Related upstream
- https://github.com/toeverything/blocksuite/pull/2948

```[tasklist]
### Tasks
- [x] support download
- [x] support syncing when first connected / add / removed pages
- [x] support persisting subdoc in .affine / .
- [x] optimize DB persisting
```

himself65:
Any update?

pengx17:
Should have been fixed in https://github.com/toeverything/AFFiNE/pull/2991

PR #2816, opened 2023-06-17
feat: sqlite subdocument

pengx17:
- [x] idb sub document sync issues
- [x] supporting sqlite subdoc (fix https://github.com/toeverything/AFFiNE/issues/2798)
- [x] native sql module should support more db operation

Some issues
- upstream blocksuite `Page.waitForLoaded` only guaranteed that the subdoc is added to the parent Doc, but it is not necessarily loaded from the persistent storage (such as IDB).
- in this PR we will use `shoudLoad` & `whenloaded` etc to load sub doc that is being flagged as to be loaded.


himself65 (changes requested):
You shouldn't upgrade blocksuite version

himself65:
I suggest you split it to a package `y-sqlite` and have `web` and `node` side

pengx17:
> I suggest you split it to a package `y-sqlite` and have `web` and `node` side

The node part is actually part of `@affine/native` binding. Unlike idb that the provider has direct access to DB APIs, the web part is calling DB APIs via RPC. I think one possible solution is to abstract the DB interface and have a provider connect to DB abstraction and implement it for both IDB and SQLite, instead of direct DB access.
Anyway the provider for SQLite may not be that generic and useful for project outside AFFiNE's case I guess, but we can revisit it later.

pengx17:
It seems e2e in electron has some stale cache
> 02:06:48.228 (helper) › getAllUpdates TypeError: this.db.getAllUpdates is not a function +1ms

himself65:
It only runs 9s, looks indeed

pengx17:
@Brooooooklyn please help me review the Rust part

Brooooooklyn (approved):

packages/native/index.d.ts
```diff
@@ -32,6 +32,11 @@ export interface UpdateRow {
   id: number;
   timestamp: Date;
   data: Buffer;
+  docId?: string;
+}
+export interface InsertRow {
+  docId?: string;
```
himself65, line 38: remove `?` ?

apps/electron/src/helper/db/base-db-adapter.ts
```diff
@@ -79,21 +79,34 @@ export abstract class BaseSQLiteAdapter {
     }
   }
 
-  async getUpdates() {
+  async getUpdates(docId?: string) {
```
himself65, line 82: why doc id could be `undefined`?

pengx17: for convention rows without doc_id is considered the root doc's update

packages/native/index.d.ts
pengx17: this is actually intended in current design.

apps/electron/src/helper/db/base-db-adapter.ts
pengx17: i think this may not be a good design and make the implementation complicated. i can change it to always populate the db in next canary version when we consider migration, is this ok?

himself65 (approved):

PR #2820, opened 2023-06-19
feat: add subdoc migration script

Saul-Mirone:


packages/env/src/blocksuite/subdoc-migration.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export function migrateToSubdoc(doc: Y.Doc): Y.Doc {
+  const needMigration = doc.get('space:meta') instanceof Y.Map;
+  if (!needMigration) {
+    return doc;
+  }
+  const output = new Y.Doc();
+  migrateMeta(doc, output);
+  migrateBlocks(doc, output);
+  return output;
```
himself65: Can we use the original doc?

Saul-Mirone: Seems we can't because we cannot remove a property from a ydoc.

himself65 (changes requested):
I need to add a test case to make sure this works correct

packages/env/src/blocksuite/subdoc-migration.ts
```diff
@@ -0,0 +1,77 @@
...
+      const version = originalVersions.get(flavour);
+      if (version !== undefined) {
+        runBlockMigration(flavour, blockData, version);
+      }
+    });
+  });
+}
+
+export function migrateToSubdoc(doc: Y.Doc): Y.Doc {
+  const needMigration = doc.get('space:meta') instanceof Y.Map;
```
himself65: Wrong logic

himself65 (changes requested):

himself65 (approved):

PR #2831, opened 2023-06-20
feat: support subdoc for WebSocketProvider

thorseraq:


himself65:
We are using socket.io now

Closed in https://github.com/toeverything/AFFiNE/pull/2881

PR #2836, opened 2023-06-21
feat(docs): add `@affine/docs` package

himself65:


himself65:
https://github.com/toeverything/AFFiNE/pull/2849

PR #2860, opened 2023-06-26
build: docker images

Brooooooklyn:


.prettierignore
```diff
@@ -2,3 +2,4 @@ pnpm-lock.yaml
 target
 lib
 test-results
+packages/graphql/src/graphql/index.ts
```
himself65, line 5: what is this

himself65 (changes requested):

.prettierignore
Brooooooklyn: graphql-codegen generated file

himself65: but you actually emit that file in source code

himself65: https://github.com/toeverything/AFFiNE/pull/2860/files#diff-cba3e7481bb9a82af6806d6bf71bfc5624df5000ac660e73a3bde48904d70046

darkskygit (approved):

Issue #2793, opened 2023-06-15
Add scrollbar to page, all pages and other necessary places


xell:
[internal design](https://www.figma.com/file/hVxJQtEqyv5xw1AbCzc50M/AFFiNE-2023?type=design&node-id=5380-113130&t=ruWkAXuJJ8E5mXDF-4)

PR #2869, opened 2023-06-27
fix: incorrect scrollbar position when opening full width layout

JimmFly:
before:
截屏2023-06-27 下午4 43 47

after:
截屏2023-06-27 下午4 44 07


JimmFly:
Supplement to #2793

himself65 (approved):

PR #2870, opened 2023-06-27
feat: add affine cloud helm config

darkskygit:


packages/graphql/src/graphql/index.ts
```diff
@@ -1,10 +1,10 @@
 /* do not manipulate this file manually. */
```
Brooooooklyn, line 1: rebase and re-generate this file

.github/helm/affine-cloud/templates/stateful-set.yaml
```diff
@@ -0,0 +1,65 @@
+apiVersion: apps/v1
```
Brooooooklyn, line 1: Let's create pg by the official postgres helm charts: https://artifacthub.io/packages/helm/bitnami/postgresql

.github/helm/affine-cloud/templates/secret.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: affine-cloud-secret
+type: Opaque
```
Brooooooklyn, line 5: And this will be managed by this charts: https://artifacthub.io/packages/helm/bitnami/postgresql

.github/helm/affine-cloud/templates/deployment.yaml
```diff
@@ -0,0 +1,41 @@
...
+      labels:
+        app: affine-cloud
+    spec:
+      restartPolicy: Always
+      containers:
+      - name: affine-cloud
+        image: "ghcr.io/toeverything/cloud-self-hosted:{{ .Values.affineCloud.tag | default .Chart.AppVersion }}"
+        env:
+        - name: DATABASE_URL
+          value: "postgresql://$(POSTGRES_USER):$(POSTGRES_PASSWORD)@postgres-service/affine"
```
Brooooooklyn, line 25: database name should also be a configable value

Brooooooklyn, line 25: I'd like to make the whole `DATABASE_URL` configurable by the helm `values`; and make `postgresql://$(POSTGRES_USER):$(POSTGRES_PASSWORD)@postgres-service/affine` as the default value

.github/helm/affine-cloud/templates/services.yaml
```diff
@@ -0,0 +1,25 @@
...
+  type: NodePort
+  selector:
+    app: affine-cloud
+  ports:
+  - name: affine-cloud
+    protocol: TCP
+    port: 3000
+    targetPort: 3000
+---
+apiVersion: v1
```
Brooooooklyn, line 15: this could be removed if we are using postgresql charts to create pg instance

Brooooooklyn (approved):

VictorGaiva:
I have a Postgres Operator on My cluster, with backups already configured. I would like to be able to tell affine to and existing Postgres Server, instead of starting its own.

darkskygit:
> I have a Postgres Operator on My cluster, with backups already configured. I would like to be able to tell Affine Helm Application to use an existing Postgres Server, instead of starting its own.

This helm configuration is used to deploy the old server. You can point the database connection to your database by setting `affineCloud.databaseUrl`.

Our server has been rewritten with nodejs, and the corresponding helm configuration is here: https://github.com/toeverything/AFFiNE/tree/master/.github/helm/affine

PR #2872, opened 2023-06-27
fix: preloading

Saul-Mirone:


himself65:
Don't remove that file or all type check will failure

Saul-Mirone:
> Don't remove that file or all type check will failure

Remove what file?

PR #2879, opened 2023-06-27
fix: remove duplicated fetch desktop workspace

himself65:


pengx17 (commented):
As we discussed offline, the electron side's workspace list fetch in `rootWorkspacesMetadataAtom` is mandatory right now. But I think since you are refactoring rootWorkspacesMetadataAtom at the same time, maybe we could keep it and rethink about it in https://github.com/toeverything/AFFiNE/pull/2882

himself65:
see https://github.com/toeverything/AFFiNE/pull/2882

PR #2881, opened 2023-06-27
feat: affine-cloud sync server & SocketIO provider

thorseraq:
- [x] sync server based on Socket.IO
- [x] corresponding sync provider
- [x] supportng subdoc
- [x] storage, collaboration test
- [x] start affine server to pass test in CI (ignored, expected to test as e2e)
- [ ] auth


apps/server/src/modules/sync/events/workspace.ts
```diff
@@ -0,0 +1,69 @@
...
+    return docs;
+  }
+
+  async getDocFromGuid(guid: string): Promise<Doc> {
+    let update;
+    try {
+      update = await this.storage.load(guid);
+    } catch (_) {
+      update = createWorkspace();
+      await this.storage.syncWithGuid(guid, Buffer.from(update));
```
forehalo, line 47: first parameter of `syncWithGuid` is workspaceId, not doc guid

forehalo, line 47: first parameter of `syncWithGuid` is workspaceId, not doc guid

himself65 (changes requested):
test needed

apps/server/src/modules/sync/events/events.gateway.ts
```diff
@@ -0,0 +1,81 @@
...
+export class EventsGateway {
+  constructor(
+    private readonly storageService: WorkspaceService,
+    private readonly storage: Storage
+  ) {}
+
+  @WebSocketServer()
+  server: any;
+
+  @SubscribeMessage('client-handshake')
```
himself65, line 28: Where is the permission check?

thorseraq: Later it will come!

apps/server/src/modules/sync/events/events.gateway.ts
```diff
@@ -0,0 +1,83 @@
...
+  @SubscribeMessage('client-update')
+  async handleClientUpdate(
+    @MessageBody()
+    message: {
+      workspaceId: string;
+      guid: string;
+      update: string;
+    }
+  ) {
+    const update = base64ToUint8Array(message.update);
```
Brooooooklyn, line 53: What's the purpose to convert `update` like this instead of just a simple `Buffer.from(message.update, 'base64')?

`base64 string -> Buffer -> Uint8Array -> Buffer`, the data is copied 3 times

Brooooooklyn: btw, `Buffer` is extended from `Uint8Array`, you can pass `Buffer` into any function that receives `Uint8Array`

thorseraq: There's no particular purpose, I haven't noticed the cost, it's inlined now

PR #2882, opened 2023-06-27
refactor: rootWorkspacesMetadataAtom loading logic

himself65:
`rootWorkspacesMetadataAtom` is now a async atom


apps/web/src/layouts/workspace-layout.tsx
```diff
@@ -335,31 +255,35 @@ export const WorkspaceLayoutInner: FC<PropsWithChildren> = ({ children }) => {
...
   const { jumpToPage } = useRouterHelper(router);
 
-  // fixme(himself65):
-  //  we should move the page into jotai atom since it's an async value
-
   //#region init workspace
-  if (currentWorkspace.blockSuiteWorkspace.isEmpty) {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-expect-error
+  if (currentWorkspace.blockSuiteWorkspace.meta._proxy.isEmpty !== true) {
```
himself65, line 261: This is a hot patch, because `workspace.isEmpty` is never true for some reason

packages/workspace/src/atom.ts
```diff
@@ -1,28 +1,54 @@
...
+
+export type RootWorkspaceMetadataV1 = z.infer<
+  typeof rootWorkspaceMetadataV1Schema
+>;
 
 export type RootWorkspaceMetadata =
   | RootWorkspaceMetadataV1
   | RootWorkspaceMetadataV2;
 
+export const workspaceAdaptersAtom = atom<
```
pengx17, line 34: `workspaceAdaptersAtom` is never changed. Why do we want it to be an atom?

himself65: We haven’t but we will.

I might change it to a mock apis in the test environment.

PR #2913, opened 2023-06-28
fix: updater color updates

pengx17:
Keeping up with the latest color scheme
image

### DIFF
--- a/packages/component/src/components/app-sidebar/app-updater-button/index.css.ts
+++ b/packages/component/src/components/app-sidebar/app-updater-button/index.css.ts
@@ -36,7 +36,7 @@ export const root = style({
     },
   },
   vars: {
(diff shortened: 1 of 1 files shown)


packages/component/src/components/app-sidebar/app-updater-button/index.css.ts
```diff
@@ -206,7 +206,7 @@ export const halo = style({
     willChange: 'filter, transform',
     transform: 'translateY(100%) scale(0.6)',
     background:
-      'radial-gradient(ellipse 60% 80% at bottom, rgba(50, 26, 206, 0.35), transparent)',
+      'radial-gradient(ellipse 60% 80% at bottom, rgba(30, 150, 235, 0.35), transparent)',
```
himself65, line 209: can we use global variable for this color?

PR #2916, opened 2023-06-29
chore: update translation

JimmFly:


packages/component/src/components/page-list/filter/vars.tsx
```diff
@@ -48,16 +50,19 @@ export const CreateFilterMenu = ({
...
   );
 };
-
 export const VariableSelect = ({
   onSelect,
 }: {
   selected: Filter[];
   onSelect: (value: Filter) => void;
 }) => {
+  const variableName = useFilterName;
```
himself65: This is a hook, don't use it like a normal variable

himself65 (approved):

PR #2924, opened 2023-06-29
feat!: affine cloud support

himself65:


.github/workflows/build.yml
```diff
@@ -413,7 +413,7 @@ jobs:
           fail_ci_if_error: false
 
   build-docker:
-    if: github.ref == 'refs/heads/master'
+    # if: github.ref == 'refs/heads/master'
```
himself65: uncomment this before merging

packages/workspace/src/affine/shared.ts
```diff
@@ -41,7 +41,7 @@ if (!globalThis.AFFINE_APIS) {
   };
   const loginMockUser2 = async () => {
     const user2 = await import('@affine-test/fixtures/built-in-user2.json');
-    const data = await fetch(prefixUrl + 'api/user/token', {
+    const data = await fetch(prefixUrl + '/api/user/token', {
```
himself65: change `prefixUrl` to an atom

apps/server/src/modules/auth/next-auth.controller.ts
```diff
@@ -29,15 +29,42 @@ export class NextAuthController {
...
+          },
+          from: config.auth.email.sender,
+        }),
+      ],
+      // @ts-expect-error Third part library type mismatch
+      adapter: prismaAdapter,
+      debug: !config.prod,
       // @ts-expect-error Third part library type mismatch
-      adapter: PrismaAdapter(prisma),
+      logger: console,
```
Brooooooklyn, line 61: remove it before merge

PR #2957, opened 2023-07-01
chore: replace 'eslint-plugin-import'

himself65:
Upstream: https://github.com/import-js/eslint-plugin-import/pull/2447#issuecomment-1117384140


Brooooooklyn:
Great work

PR #2984, opened 2023-07-03
test: image preview e2e

danielchim:
Now it should able to cover all the features in the image-preview component.


himself65 (changes requested):
Please fix the e2e error

himself65:
I have figured out the problem: the nx cache hit the wrong version because we have an incorrect configuration. I've fixed this in https://github.com/toeverything/AFFiNE/pull/3079

himself65 (approved):

himself65:
See https://github.com/toeverything/AFFiNE/pull/3080

PR #2991, opened 2023-07-03
fix: sqlite provider import sub doc db file

pengx17:
In the new migration workflow, workspace persisted in the electron cache path will be removed and re-created as well via `bootstrap`.

One missing part is that how to due with import the old DB. In this pr, we prevent user from loading old format db files when validating.


pengx17:
@Brooooooklyn @himself65 please take a look if this approach is acceptable.

.github/workflows/build.yml
```diff
@@ -352,14 +352,8 @@ jobs:
...
           NATIVE_TEST: 'true'
 
-      - name: Build Infra
-        run: yarn run build:infra
-
-      - name: Build Plugins
-        run: yarn run build:plugins
-
       - name: Build Desktop Layers
-        run: yarn workspace @affine/electron build
```
himself65: replace all ci actions

himself65 (changes requested):
Is it possible to test the migration in e2e? Moving a `.affine` file into fixtures is acceptable.

himself65 (approved):

himself65:
hope everything works!

Issue #1968, opened 2023-04-16
New core component `Notification Center`


himself65:
Upstream: https://github.com/toeverything/AFFiNE/issues/1870

### Notification Center

Permission: router, current workspace, global event

collection of toast

### Toast (Pure UI)

single message

// TODO

PR #3011, opened 2023-07-04
feat(electron): move preload to infra

himself65:


pengx17:
May I ask what's the purpose of moving preload to infra?

himself65:
> May I ask what's the purpose of moving preload to infra?

We might not only have one distribution in the future(`preload/mobile`, `preload/mock`), and the current electron has a `env.d.ts` with the namespace, which is not supported to be there, and it out of the typescript module.

And I want to add a mock `desktop api`(a fake `window.apis` `window.platform`) if I have time so that we can test the logic purely in web e2e.


Also, forget to say that this is also a goal of the notification center (https://github.com/toeverything/AFFiNE/issues/1968), which moves the handler manager to the top level and adds `event emitter` API could make things a little step forward

packages/infra/src/type.ts
```diff
@@ -0,0 +1,161 @@
...
+  Namespace extends string,
+  Handlers extends Record<string, PrimitiveHandlers>
+> {
+  static instance: HandlerManager<string, Record<string, PrimitiveHandlers>>;
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  private _app: App<Namespace, Handlers> = null!;
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  private _namespace: Namespace = null!;
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  private _handlers: Handlers = null!;
```
himself65: not finished yet

packages/infra/preload/electron.js
```diff
@@ -0,0 +1,3 @@
+/* eslint-disable */
+/// <reference types="../dist/preload/electron.d.ts" />
+export * from '../dist/preload/electron.js';
```
himself65, line 3: `"moduleResolution": "Node"` compatibility

pengx17 (approved):

himself65:
> > May I ask what's the purpose of moving preload to infra?
> 
> We might not only have one distribution in the future(`preload/mobile`, `preload/mock`), and the current electron has a `env.d.ts` with the namespace, which is not supported to be there, and it out of the typescript module.
> 
> And I want to add a mock `desktop api`(a fake `window.apis` `window.platform`) if I have time so that we can test the logic purely in web e2e.
> 
> Also, forget to say that this is also a goal of the notification center (#1968), which moves the handler manager to the top level and adds `event emitter` API could make things a little step forward

Like `preload/web-polyfill` to make a fake desktop experience in the browser, this could be the plugin system that happened in the web browser, like vscode

PR #3072, opened 2023-07-06
fix: cleanup workspace when switch setting panel

himself65:
@QiShaoXuan Don't use `?` if you already know the default value. And set the workspace to null to avoid UI lagging


QiShaoXuan:
> @QiShaoXuan Don't use `?` if you already know the default value. And set the workspace to null to avoid UI lagging

get it～

PR #3076, opened 2023-07-06
feat: add subpage apis

darkskygit:


himself65 (approved):
How to download a subpage(or, we say a subdoc)?

darkskygit:
> How to download a subpage(or, we say a subdoc)?

include next pr about storage, still in developing because some database glitch

PR #3084, opened 2023-07-07
chore: add new item for share component

### LINKED ISSUE #3066: Two export menus should be identical
image

image

JimmFly:
Close #3066


himself65 (commented):
Can we abstract share menu into one component?

himself65 (changes requested):
Let's abstract this component into a **single one** so that the UI will be consistent.

Please do that

image

himself65 (approved):

Issue #3068, opened 2023-07-06
Workspace Settings in quick search window should link to the corresponding location of the New Settings


xell:
![image](https://github.com/toeverything/AFFiNE/assets/132558/505a75d3-15d3-48cb-a44c-05c917702bc7)

joebeijing:
And it's name should be "Settings" instead.

PR #3089, opened 2023-07-07
feat: modify setting modal entry in quick search modal

QiShaoXuan:
fixed: https://github.com/toeverything/AFFiNE/issues/3068


himself65 (changes requested):
Please merge to the master.

Now we use `workspaceId` in the props and use `useStaticWorkspace` for blocksuite data
and `useWorkspace` for flavour-related data.

Also, please add a suspense boundary into the setting panel to make sure the skeleton is correct

Related: https://github.com/toeverything/AFFiNE/issues/3095

apps/web/src/shared/index.ts
```diff
@@ -47,9 +47,9 @@ export const pathGenerator = {
 };
 
 export const publicPathGenerator = {
-  all: workspaceId => `/public-workspace/${workspaceId}/all`,
-  trash: workspaceId => `/public-workspace/${workspaceId}/trash`,
-  shared: workspaceId => `/public-workspace/${workspaceId}/shared`,
+  all: workspaceId => `/share/${workspaceId}/all`,
+  trash: workspaceId => `/share/${workspaceId}/trash`,
+  shared: workspaceId => `/share/${workspaceId}/shared`,
```
himself65: I think we should remove `trash` and `shared` here? because we don't have such that?

himself65 (approved):

PR #3094, opened 2023-07-07
fix: a possible issue on electron flaky test

pengx17:
I suspect that there could be race condition that writes to workspace meta on initialization. So maybe adding a timeout could resolve the issur.


himself65 (commented):
What is the difference? Could you please give us a brief description?

pengx17:
> What is the difference? Could you please give us a brief description?

Sure. updated the summary

PR #3097, opened 2023-07-07
fix: page mode shortcut

danielchim:


apps/web/src/components/blocksuite/workspace-header/editor-mode-switch/index.tsx
```diff
@@ -51,8 +55,28 @@ export const EditorModeSwitch = ({
     return () =>
       document.removeEventListener('keydown', keydown, { capture: true });
   }, [setSetting, t]);
+
+  const TooltipContent = () => {
```
himself65: please move it out side and make it to a new component instead of in the function context

apps/web/src/components/blocksuite/workspace-header/editor-mode-switch/index.tsx
```diff
@@ -51,8 +55,28 @@ export const EditorModeSwitch = ({
...
+
   return (
-    <Tooltip content={'Switch ⌘ + S'}>
+    <Tooltip
+      content={
+        (<TooltipContent />) as ReactElement<
+          any,
+          string | JSXElementConstructor<any>
+        > &
+          string
```
himself65: I don't think this is the correct type

himself65 (changes requested):

apps/web/src/hooks/affine/use-shortcuts.ts
```diff
@@ -123,7 +123,8 @@ export const useMacPageKeyboardShortcuts = (): ShortcutTip => {
       [t['Increase indent']()]: 'Tab',
       [t['Reduce indent']()]: '⇧+Tab',
       [t['Group as Database']()]: '⌘ + G',
-      ['Switch']: '⌘ + S',
+      // TODO: i18N
+      ['Switch']: '⌥ + S',
```
himself65: Please update i18n also

himself65 (changes requested):

himself65 (approved):

PR #3105, opened 2023-07-08
fix: disable move db by default

pengx17:


apps/web/preset.config.mjs
```diff
@@ -28,6 +28,7 @@ const buildPreset = {
     enableNewSettingModal: true,
     enableNewSettingUnstableApi: false,
     enableSQLiteProvider: true,
+    enableMoveDatabase: false,
```
himself65, line 31: there is need another one, under line L49

himself65 (approved):

PR #3108, opened 2023-07-08
fix: user info lack

darkskygit:


himself65 (approved):

darkskygit:
seems break storage apis, will be fix today
merge first

PR #3115, opened 2023-07-09
fix: add DB migration to add workspace

pengx17:
This pr allows the user to add a old db file without sub docs. When loading an old db, it will check if it had doc_id column. If not, it will firstly clone the source db to a tmp folder, migrate it, persist to db file and then use the new db file path as the new db source.


apps/electron/src/helper/db/migration.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const rows = await db.getAllUpdates();
+  const originalDoc = new Y.Doc();
+
+  // 1. apply all updates to the root doc
+  rows.forEach(row => {
+    Y.applyUpdate(originalDoc, row.data);
+  });
+
+  console.log(originalDoc.toJSON());
```
himself65: No console

apps/electron/tsconfig.json
```diff
@@ -11,7 +11,8 @@
     "outDir": "dist",
     "moduleResolution": "node",
     "resolveJsonModule": true,
-    "noImplicitOverride": true
+    "noImplicitOverride": true,
+    "verbatimModuleSyntax": false
```
himself65: Why

packages/native/index.d.ts
```diff
@@ -38,6 +38,12 @@ export interface InsertRow {
   docId?: string;
   data: Uint8Array;
 }
+export const enum ValidationResult {
```
himself65: Remove const here could helpful for removing compiler options

himself65 (changes requested):

packages/native/index.d.ts
pengx17: This is auto generated. Not sure if there is a better option 😳

apps/electron/tsconfig.json
pengx17: verbatimModuleSyntax cannot be used with `export const` from an external module.

packages/native/index.d.ts
himself65: Related: https://github.com/napi-rs/napi-rs/issues/1345

you need `--const-enum false`

himself65 (changes requested):

pengx17:
> Related: https://github.com/napi-rs/napi-rs/issues/1345
> you need --const-enum false

Looks like the v2 version we are using not honoring ts.constEnum option in package.json. I proposed a PR upstream: https://github.com/napi-rs/napi-rs/pull/1647

himself65 (approved):

PR #3116, opened 2023-07-10
fix: the image lost after exporting

### LINKED ISSUE #3122: After exporting Bookmark as an image, the thumbnail and icon are not visible
CleanShot 2023-07-04 at 17 12 34@2x
The two elements pointed by the arrow are blank after exporting.

CleanShot 2023-07-04 at 17 13 21@2x

JasperStardream:
close #3122


Brooooooklyn:
@zuoxiaodong0815 where is the source code in the https://proxyimage.toeverything.workers.dev? It should be committed with this pull request.

zuoxiaodong0815:
Ok, I got it. l will see how to add workers to the affine repository.

himself65 (changes requested):
Please add the source code to a new package like `packages/worker` to make sure every part is open source.

If you don't know how to integrate with affine-web, I will handle that

zuoxiaodong0815:
Because adding a new pr(https://github.com/toeverything/AFFiNE/pull/3150), close this pr.

PR #3117, opened 2023-07-10
fix: font style setting only control editor's font

QiShaoXuan:


apps/web/src/components/page-detail-editor.tsx
```diff
@@ -69,12 +69,21 @@ const EditorWrapper = memo(function EditorWrapper({
   const [appSettings] = useAppSetting();
 
   assertExists(meta);
-
+  const value = useMemo(() => {
+    return fontStyleOptions.find(option => option.key === appSettings.fontStyle)
```
himself65: I think we should assertExist here since it's a enum

himself65 (approved):

PR #3118, opened 2023-07-10
refactor: replace yrs with yjs

thorseraq:


packages/storage/Cargo.toml
```diff
@@ -10,8 +10,8 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
-jwst = { git = "https://github.com/toeverything/OctoBase.git", branch = "master" }
-jwst-storage = { git = "https://github.com/toeverything/OctoBase.git", branch = "master" }
+jwst = { git = "https://github.com/toeverything/OctoBase.git", rev = "da3824e" }
```
thorseraq, line 13: refers: https://github.com/toeverything/OctoBase/commit/da3824e419dbf2cf514ddf68402e8f211f2564f8

apps/server/src/modules/sync/events/workspace.ts
```diff
@@ -44,14 +44,18 @@ export class WorkspaceService {
...
-      // TODO load method return null if doc doesn't exist, error throwing is not needed.
-      const update = await this.storage.load(guid);
-      update && Y.applyUpdate(doc, update);
-      return doc;
+      updates.forEach(update => {
+        Y.applyUpdate(doc, update);
+      });
     } catch (e) {
-      return null;
+      console.error(e);
```
himself65: I think we should return null here if update broke the yjs doc?

thorseraq: emmmm, make sense, how do you think of this way?

[`d525500` (#3118)](https://github.com/toeverything/AFFiNE/pull/3118/commits/d525500187ba41d8dd401e0389bcccb66c70de32)

himself65: I'm not sure how it is possible that an update cause an error.

himself65 (approved):

Issue #3021, opened 2023-07-05
cloud storage api with blocksuite


himself65:
Reference: https://github.com/toeverything/blocksuite/blob/master/packages/store/src/persistence/blob/mock-server-storage.ts

Use Affine Cloud to implement the corresponding Storage API, use Import {fetcher} from '@AFFINE/GRAPHQL'

JimmFly:
Is the crud of the blob ready? @darkskygit

darkskygit:
> Is the crud of the blob ready? @darkskygit

https://github.com/toeverything/AFFiNE/blob/himself65/0629-login/packages/graphql/src/graphql/index.ts

darkskygit:
all apis ready

PR #3125, opened 2023-07-10
feat: add cloud blob storage api

himself65:
Fixes: https://github.com/toeverything/AFFiNE/issues/3021


packages/workspace/src/blob/cloud-blob-storage.ts
```diff
@@ -0,0 +1,47 @@
...
+} from '@affine/graphql';
+import type { BlobStorage } from '@blocksuite/store';
+
+import { fetcher } from '../affine/gql';
+
+export const createCloudBlobStorage = (workspaceId: string): BlobStorage => {
+  return {
+    crud: {
+      get: async () => {
+        throw new Error('Not implemented');
```
himself65: @darkskygit

himself65:
See https://github.com/toeverything/AFFiNE/pull/3163

PR #3136, opened 2023-07-10
fix: try redirecting api requests in electron

pengx17:
This PR tries to use `protocol.handle` to redirect `/api` and `/graphql` requests to `CLOUD_URL`.
However the attempt was blocked:
- when using `file://` scheme, it seems we cannot overwrite it with `protocol.handle`. see https://github.com/electron/electron/issues/39031
- when using `assets://` scheme, the message from render to helper cannot be intercepted.

In either way, the proposal seems to be stuck at the moment. Alternatively and maybe more reasonably, we shall add prefix urls to `fetch` usage in the renderer, instead of calling `/api/xxx` instead.

### DIFF
--- a/apps/electron/src/main/updater/electron-updater.ts
+++ b/apps/electron/src/main/updater/electron-updater.ts
@@ -26,10 +26,13 @@ export const quitAndInstall = async () => {
 
 let lastCheckTime = 0;
 export const checkForUpdatesAndNotify = async (force = true) => {
+  logger.info('checkForUpdatesAndNotify', autoUpdater);
   // check every 30 minutes (1800 seconds) at most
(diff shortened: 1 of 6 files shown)


apps/electron/src/main/updater/electron-updater.ts
```diff
@@ -99,6 +102,8 @@ export const registerUpdater = async () => {
   });
   autoUpdater.forceDevUpdateConfig = isDev;
 
+  await checkForUpdatesAndNotify();
+
```
himself65, line 106: Unrelated to this PR

himself65:
Closed in https://github.com/toeverything/AFFiNE/pull/3178

PR #3150, opened 2023-07-11
fix: the image lost after exporting

### LINKED ISSUE #3122: After exporting Bookmark as an image, the thumbnail and icon are not visible

zuoxiaodong0815:
close #3122


.github/workflows/workers.yml
```diff
@@ -0,0 +1,18 @@
+name: Deploy
+
+on:
```
Brooooooklyn, line 3: We can add a paths filter to reduce this workflow running times: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

himself65 (approved):

PR #3160, opened 2023-07-11
feat(affine-cloud): add metrics

thorseraq:


apps/server/src/modules/workspaces/controller.ts
```diff
@@ -33,6 +35,11 @@ export class WorkspacesController {
     @Param('name') name: string,
     @Res() res: Response
   ) {
+    this.metric.restRequest(1, {
+      method: 'GET',
+      path: '/api/workspace/:id/blobs/:name',
+      job: 'get workspace blob',
```
forehalo, line 41: we know what each controller does, the job field seems useless

thorseraq: Updated, haven't notice that!

forehalo (approved):

apps/server/src/metrics/metrics.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable, OnModuleDestroy } from '@nestjs/common';
+import { register } from 'prom-client';
+
+import { metricsCreator } from './utils';
+
+@Injectable()
+export class Metrics implements OnModuleDestroy {
+  constructor() {}
```
Brooooooklyn, line 8: useless

PR #3176, opened 2023-07-12
fix: enhancing the security of image proxy

### LINKED ISSUE #3173: Code enhancement in #3150
1. https://github.com/toeverything/AFFiNE/pull/3150/files#diff-d3601566873293cfe4056f0478234be375e66280ceea8475258cb8ab886255fbR42 the `*` should be the origin for security reason
2. https://github.com/toeverything/AFFiNE/pull/3150/files#diff-d3601566873293cfe4056f0478234be375e66280ceea8475258cb8ab886255fbR50 the `||` should be `??`

zuoxiaodong0815:
close #3173

### DIFF
--- a/packages/workers/src/index.ts
+++ b/packages/workers/src/index.ts
@@ -39,7 +39,11 @@ async function proxyImage(request: Request): Promise<Response> {
 
   const response = await fetch(imageRequest);
   const modifiedResponse = new Response(response.body);
-  modifiedResponse.headers.set('Access-Control-Allow-Origin', '*');
+
+  modifiedResponse.headers.set(
+    'Access-Control-Allow-Origin',
+    request.headers.get('Origin') ?? 'false'
+  );
   modifiedResponse.headers.set('Access-Control-Allow-Methods', 'GET');
 
   return modifiedResponse;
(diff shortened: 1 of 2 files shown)


himself65 (approved):

packages/workers/src/index.ts
```diff
@@ -39,15 +39,19 @@ async function proxyImage(request: Request): Promise<Response> {
 
   const response = await fetch(imageRequest);
   const modifiedResponse = new Response(response.body);
-  modifiedResponse.headers.set('Access-Control-Allow-Origin', '*');
+
+  modifiedResponse.headers.set(
+    'Access-Control-Allow-Origin',
+    request.headers.get('Origin') ?? 'false'
```
Brooooooklyn, line 45: `'false'` is an invalid value for `Access-Control-Allow-Origin` header, set it to `null` or put it in if the expression: 

```
const origin = request.headers.get('Origin')
if (origin) {
    modifiedResponse.headers.set('Access-Control-Allow-Origin', origin)
}
```

btw the `request.headers.get('Origin')` could be shared between `handler.fetch` and `proxyImage` functions so you don't need to get it from `request.headers` twice

Brooooooklyn (changes requested):
See comments

packages/workers/src/index.ts
Brooooooklyn: See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

PR #3177, opened 2023-07-12
feat: upgrade dependencies

Brooooooklyn:


himself65 (changes requested):
dont upgrade waku

himself65 (changes requested):
do not upgrade swr

https://github.com/dai-shi/waku/issues/96

Issue #3166, opened 2023-07-11
The effect of full width layout toggle is inverted


xell:
https://github.com/toeverything/AFFiNE/assets/132558/e7d17a0e-a238-4d32-9b58-1cd5a25130ea

himself65:
Landed in https://github.com/toeverything/AFFiNE/commit/73a47f31a7edc5bd85df2a073b2a05aa6f39b4f9

PR #3195, opened 2023-07-12
fix: full screen setting not work

QiShaoXuan:
https://github.com/toeverything/AFFiNE/issues/3166


himself65 (changes requested):
do not only update one package

himself65:
see https://github.com/toeverything/AFFiNE/pull/3196

Issue #3203, opened 2023-07-12
Stale page description state in all pages list after page content update


Yanxi0202:
version latest version 
I use windows.

https://github.com/toeverything/blocksuite/assets/134897361/361db986-86ac-49b6-b09e-26fc417a0377

xell:
Note: wait for another solution mentioned in https://github.com/toeverything/AFFiNE/pull/3266

PR #3205, opened 2023-07-13
fix: page preview missing

himself65:
Fixes: https://github.com/toeverything/AFFiNE/issues/3203


pengx17:
This could be superseded by https://github.com/toeverything/AFFiNE/pull/3266

Issue #3211, opened 2023-07-13
Collections are persistent across workspaces


ShortCipher5:
Maybe this is by design currently but collections are persistent across workspaces.

I think collections should be unique for workspaces.

If I'm creating a "School" workspace and have tags such as "Maths" and "English" I don't want those collections to show or be pinned in my "Work" workspace.

https://github.com/toeverything/AFFiNE-Community/assets/4605025/e9604064-5110-476f-ae18-60a890b95184

PR #3213, opened 2023-07-13
fix: collections should be unique for workspaces

zzj3720:
close: #3211


apps/web/src/components/blocksuite/block-suite-page-list/index.tsx
```diff
@@ -98,6 +99,7 @@ export const BlockSuitePageList: React.FC<BlockSuitePageListProps> = ({
   collection,
 }) => {
   const pageMetas = useBlockSuitePageMeta(blockSuiteWorkspace);
+  const [currentWorkspace] = useCurrentWorkspace();
```
himself65: `blockSuiteWorkspace.id`

himself65 (changes requested):

himself65 (approved):

PR #3229, opened 2023-07-14
fix: internal build updater

pengx17:
Looks like `0.0.0-20230714031808-0ca9f2c` is not a valid semversion for electron-updater.
It requires it to be something like `0.0.0-channel.build-number`. The channel in the name is crucial here. 

https://github.com/electron-userland/electron-builder/blob/84ed3ff123b5ae92cd3350d64677434f9b397b76/packages/electron-updater/src/providers/GitHubProvider.ts#L59-L94


himself65 (approved):
This is cool. Thanks

PR #3257, opened 2023-07-17
feat(plugin-infra): new plugin system

himself65:
1. All plugins will be generated in `public/plugins`


himself65:
See https://github.com/toeverything/AFFiNE/pull/3323

PR #3259, opened 2023-07-17
feat: move plugins config to setting

### LINKED ISSUE #3200: move plugin config panel into setting modal
Move https://github.com/toeverything/AFFiNE/blob/master/apps/web/src/pages/plugins.tsx into setting modal

JimmFly:
Close #3200


himself65 (commented):
could you please also delete pluign page

PR #3267, opened 2023-07-17
refactor!: remove `next.js`

himself65:
`@affine/core` will be the replacement of the `@affine/web` that uses self-configured webpack and loading logic.

This will benefit the router logic and plugin infra.

---
1. Load the bootstrap (desktop/web/mock)
2. Check if the first load, create the first demo workspace
3. init UI
4. Register plugins
---

Refs:
- https://github.com/DimensionDev/Maskbook/blob/develop/packages/mask/.webpack/config.ts
- https://github.com/perfsee/perfsee/blob/main/tools/webpack/webpack.config.ts


apps/core/.webpack/config.ts
```diff
@@ -0,0 +1,179 @@
...
+  '@toeverything/y-indexeddb',
+];
+
+const isProduction = () => process.env.NODE_ENV === 'production';
+
+const OptimizeOptionOptions: () => webpack.Configuration['optimization'] =
+  () => ({
+    minimize: isProduction(),
+    minimizer: [
+      new TerserPlugin({
```
Brooooooklyn, line 35: use SWC minifier https://github.com/webpack-contrib/terser-webpack-plugin#swc

apps/core/.webpack/config.ts
```diff
@@ -0,0 +1,179 @@
...
+    context: rootPath,
+    output: {
+      path: join(rootPath, 'dist'),
+      publicPath,
+    },
+
+    mode: isProduction() ? 'production' : 'development',
+
+    devtool: isProduction()
+      ? 'hidden-nosources-source-map'
```
Brooooooklyn, line 91: Should it be configurable? like debug source codes in electron app

apps/core/.webpack/config.ts
```diff
@@ -0,0 +1,179 @@
...
+      rules: [
+        {
+          test: /\.m?js?$/,
+          resolve: {
+            fullySpecified: false,
+          },
+        },
+        {
+          include: require.resolve('serialize-javascript'),
+          sideEffects: false,
```
Brooooooklyn, line 108: remove it

apps/core/.webpack/config.ts
himself65: Working on it

PR #3313, opened 2023-07-19
feat: load side effect from worker

himself65:


himself65:
stale PR, closing for now

PR #3326, opened 2023-07-20
build: asar support

pengx17:
blocked by https://github.com/electron/electron/issues/38957
we have plugin-infra dynamic loaded and this will cause issues in asar build

image


himself65 (commented):
plugin-infra will use cjs module in the future

himself65 (changes requested):
still now working

himself65:
We have removed the usage of esm module, closing

himself65:
no, shouldn't close, I will continue on this PR

Issue #3029, opened 2023-07-05
Page (sub doc) Lazy Load


pengx17:
pages/subdocuments are loaded all at once for every workspace and this could take a long time. We may want to delay its loading on demand.

The proposed diagram:

![image](https://github.com/toeverything/AFFiNE/assets/584378/d990e1dd-6e38-4dee-b1b7-5d0d6577df92)

The key is to rely on React hook lifecycle to determine whether or not a page or part of a page is being used.
On connection, we will find the doc in the doc tree, increase its reference counter.
On unmount (React hook cleanup in `useEffect`), call `provider.disconnect(guid)`. This time we will decrease the doc by 1.

If the reference counter equals to 1, we will setup the connection:
- sync doc with the datasource. i.e., get the full doc from remote, apply it to local and then send the diff to remote
- setup doc's `update` event listener. We may not need to setup other events any longer
- setup remote's update event. (this may not be required if the remote does not send such events)

If the reference counter equals to 0, we will cleanup the listeners.

---

Above is outdated. New design is updated below:

![image](https://github.com/toeverything/AFFiNE/assets/584378/2c23f9bc-5067-4383-8ee0-5b52067b7bc3)

```[tasklist]
### Tasks
- [ ] https://github.com/toeverything/AFFiNE/issues/3230
- [ ] https://github.com/toeverything/AFFiNE/issues/2914
```

pengx17:
![image](https://github.com/toeverything/AFFiNE/assets/584378/d6f41a08-252c-4787-8546-f8b06a22dbcd)
Before & after. Using dumped AFFiNE data.

PR #3330, opened 2023-07-20
perf: lazy doc provider factory

pengx17:
Add a new module to share it to other doc providers in AFFiNE.

This module formalize a common doc lazy load pattern, defined in https://github.com/toeverything/AFFiNE/issues/3029.

To use a lazy doc provider, firstly you need to implement the interface for `DatasourceDocAdapter`. Then, you can create on Lazy doc provider (an instance of PassiveDocProvider) with `createLazyProvider(doc, datasource)`.

Implementations for idb/sqlite is on the way (could be in anther PR).


himself65 (approved):
LGTM

PR #3341, opened 2023-07-21
fix(core): page full refresh

himself65:


himself65 (commented):
@pengx17 could you please fix the desktop test failure?

pengx17:
It seems that after the router loader changes, there could be some changes made to y.Doc between sqlite download & passive sqlite connection. For now I added a workaround to sync the diff from y.Doc to sqlite.
This would be a workaround and will be removed after the lazy load provider is up for Sqlite.

PR #3349, opened 2023-07-23
fix: share WorkspaceLayout for workspace routes

pengx17:
On route change under `/workspace`, every route content will restart the workspace disconnection/connection again and is a waste and may cause issues if provider is not well implemented.
This PR fix the issue by moving WorkspaceLayout to a common layer to /:pageId, /all and /trash


pengx17:
Just noticed that https://github.com/toeverything/AFFiNE/pull/3341 implements the same behavior.

PR #3370, opened 2023-07-25
feat(affine-cloud): traceId, serve timing for GraphQL, global timing log

thorseraq:
1. Add requestId in header for business requests in format `x-request-id: N35V6Yg50BMse-yZ8pn6k`
2. implement https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing for cloud server, example:

	### GraphQL Requests (currently more fine-grained)
	image
	
	### Other Requests
	image

3. Log graphQL requests with requestId(When error encountered, more details will be printed)
	
	image


packages/graphql/src/fetcher.ts
```diff
@@ -1,6 +1,7 @@
 import type { ExecutionResult } from 'graphql';
 import { GraphQLError } from 'graphql';
 import { isNil, isObject, merge } from 'lodash-es';
+import { v4 as uuidv4 } from 'uuid';
```
Brooooooklyn, line 4: https://bundlephobia.com/package/nanoid@4.0.2 is smaller

Brooooooklyn (approved):

PR #3371, opened 2023-07-25
refactor: lazy provider for cloud

forehalo:


packages/workspace/src/affine/index.ts
```diff
@@ -0,0 +1,161 @@
...
+          callback(message.guid, base64ToUint8Array(message.update));
+        }
+      };
+      socket.on('server-update', onUpdate);
+      const destroyAwareness = setupAffineAwareness(socket, rootDoc, awareness);
+
+      return () => {
+        socket.off('server-update', onUpdate);
+        destroyAwareness();
+        // socket.disconnect()
```
thorseraq, line 86: just in case, are we good to keep this?
 https://github.com/toeverything/AFFiNE/blob/fe32eccc03f5c3f8ca899afa309f3fe53b0c10b9/packages/workspace/src/affine/sync-socket-io/index.ts#L192

then use `destroyAwareness().then(() => socket.disconnect())`

packages/workspace/src/providers/index.ts
```diff
@@ -32,30 +33,11 @@ const createAffineSocketIOProvider: DocProviderCreator = (
...
-      provider.connect();
-    },
-    disconnect: () => {
-      provider.disconnect();
-    },
-    cleanup: () => {
-      provider.destroy();
-    },
-  } satisfies AffineSocketIOProvider;
+    ...createLazyProvider(doc, dataSource),
```
pengx17, line 39: For IDB, I am a little worried that the provider may has issues on frequent connect/disconnect for the same provider  instance and the cleanup logic may conflicts with the newly connection logic, thus what I did is to create a new instance on re-connection.

forehalo: let's see what will happen then

pengx17:
I haven’t checked through but can I ask if the new provider also has an active download provider that
will download all pages once sync? 
This is a requirement from the product side.

forehalo:
> I haven’t checked through but can I ask if the new provider also has an active download provider that will download all pages once sync? This is a requirement from the product side.

The `sync` event is actually `synced`, not used to tell provider that should load any doc, but used to tell doc holder that the doc is up to date.

For instance, in `apps/core/src/adapters/cloud/crud.ts#L62`, we wait for the subdoc content loaded so we can apply the content to the new subdoc, and sync them to the newdoc providers:

```
subdoc.load();
await subdoc.whenLoaded;
//           ^^^^^^^^^^

newBLockSuiteWorkspace.doc.subdocs.forEach(newSubdoc => {
  if (newSubdoc.guid === subdoc.guid) {
    Y.applyUpdate(newSubdoc, Y.encodeStateAsUpdate(subdoc));
  }
});
```

the progress would be: `doc.load => provider.subdocHandler => await sync doc => doc.emit('sync') => whenLoaded.resolve()`

pengx17:
> the progress would be: doc.load => provider.subdocHandler => await sync doc => doc.emit('sync') => whenLoaded.resolve()

I see. This will effectively mark all subdoc as should load and making them eagerly loaded later but not loading them on demand.

pengx17 (approved):

Issue #3272, opened 2023-07-17
Support installation packages of different versions such as .deb and .rpm under Linux


houjoe0829:
- Need to distinguish between 64-bit and 32-bit
- Provide option of "Compile from source"
- Suggest aligning with Dropbox's Linux installation method  https://www.dropbox.com/install-linux 
- Need to update the UI of the AFFiNE download page.

SolarDesalination:
I think using Snap, AppImage, or Flatpak would be better then Deb or RPM since it seems that's where the state of development is headed for linux package managers. Also would suggest packaging in Linglong for Deepin/China UOS market.

pengx17:
We have merged https://github.com/toeverything/AFFiNE/pull/3401 which adds appimage support in version 0.8

PR #3401, opened 2023-07-27
build: add AppImage build

pengx17:
Tried bundle AppImage build on linux arm64 and it works.
related to https://github.com/toeverything/AFFiNE/issues/3272


himself65 (approved):
LGTM

himself65 (changes requested):
update nightly-build.yml also

himself65 (approved):

PR #3405, opened 2023-07-27
fix: unable to add a second collection

JimmFly:
After creating a collection, it cannot be created again, and the click will change to modify the collection. This is not expected behavior.


himself65 (approved):
I think the whole implementation of `EditCollectionModel` is incorrect, you shouldn't depend on react re-render

@zzj3720

Issue #3412, opened 2023-07-27
Delete the last page in a workspace crashed​


teobler:
### What happened?

![bugs](https://github.com/toeverything/AFFiNE/assets/26424691/e619c4fd-0d7a-43a5-82a8-c2de911953c0)

### Distribution version

macOS x64 (Intel)

### What browsers are you seeing the problem on if you're using web version?

Chrome

### Relevant log output

```shell
Uncaught runtime errors:

ERROR
Cannot read properties of undefined (reading 'tags')
TypeError: Cannot read properties of undefined (reading 'tags')
  at get tags [as tags] (webpack-internal:///./node_modules/@blocksuite/blocks/dist/page-block/default/meta-data/meta-data.js:140:31)
  at PageMetaData.renderTagsInline (webpack-internal:///./node_modules/@blocksuite/blocks/dist/page-block/default/meta-data/meta-data.js:52:31)
  at PageMetaData.render (webpack-internal:///./node_modules/@blocksuite/blocks/dist/page-block/default/meta-data/meta-data.js:159:51)
  at PageMetaData.update (webpack-internal:///./node_modules/lit-element/development/lit-element.js:127:28)
  at PageMetaData.performUpdate (webpack-internal:///./node_modules/@lit/reactive-element/development/reactive-element.js:825:22)
  at PageMetaData.scheduleUpdate (webpack-internal:///./node_modules/@lit/reactive-element/development/reactive-element.js:760:21)
  at PageMetaData.__enqueueUpdate (webpack-internal:///./node_modules/@lit/reactive-element/development/reactive-element.js:733:29)
```

### Anything else?

After the last page be deleted, seems we do not set **currentPageId** to **null** 
`assertExists(pageMeta)` checking cannot pass in `BlockSuiteEditorHeader`

```
Uncaught runtime errors:

ERROR
val does not exist
  at assertExists (webpack-internal:///./node_modules/@blocksuite/global/dist/utils.js:37:15)
  at BlockSuiteEditorHeader (webpack-internal:///./apps/core/src/components/blocksuite/workspace-header/index.tsx:60:76)
  at renderWithHooks (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:16305:18)
  at updateFunctionComponent (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:19583:20)
  at beginWork (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:21596:16)
[...]

Issue #3359, opened 2023-07-24
Improve e2e coverage on page deletion


himself65:
```[tasklist]
### E2E Improvement
- [ ] page delete -> it should be deleted
- [ ] page delete -> refresh page -> it should be disappear
- [ ] page delete -> create new page -> refresh page -> new page should be appear -> old page should be disappear
- [ ] delete multiple pages -> create multiple pages -> refresh
```

teobler:
Doing some testing related stuff spike around this issue and #3198 

1. during coding, [#3412](https://github.com/toeverything/AFFiNE/issues/3412) happened, will block this issue.
2. need more context about 'logic abstraction' in #3198, something like build a page level model/class for holding navigation/click logic?

Issue #3198, opened 2023-07-12
e2e test logic abstraction


himself65:
For now, we have many duplicated e2e test logic.  Like editor initiation, switch workspace, change settings...

I want to make things easier to write a new test instead of opening `dom inspector` and see what I should do next

EYHN:
It's better now

PR #3414, opened 2023-07-27
test: add e2e test case for page delete(#3359), blocked by #3412

teobler:
related to #3359, and want to do some improvement during development for #3198 

but this PR currently be blocked by #3412 

any feedback is welcome!


himself65 (commented):
Seems like you both are working on the same thing

https://github.com/toeverything/AFFiNE/pull/3416 by @Kaneki-Ken-7

himself65:
Please wait for https://github.com/toeverything/AFFiNE/pull/3419

teobler:
> Seems like you both are working on the same thing
> 
> #3416 by @Kaneki-Ken-7

Hi @himself65, okay, I aborted my PR then, but something I noticed during development, not sure if valuable, **just ignore me if anything wrong**:face_in_clouds:, since I do not have enough context at this time:sweat_smile::

### Testing Tool
AFFiNE is using e2e test as primary quality maintenance tool, according to 'Test Pyramid', the **cost** of e2e test could be much more than unit test.

The cost here will be the effort for developing and maintaining(you need to know each element's name / position / test-id, etc.).

And more importantly, TIME. One test case of e2e testing will cost as least 5s - 8s to init env, running and shot down, compared to unit testing, this time is 0.xs - 2s by using Jest, if using vitest, this time should be 'ms' level. This will be a pain point in the future when project growing bigger and bigger.

In one of my experienced project, 1500+ unit testing cases(Jest) will take around 12mins, but 200-300 Cypress e2e testing cases will take around 25mins. I do think AFFiNE will be a big project in the future, if we all in e2e testing we will lose fast feedback from testing.

Just like after someone modified some code in codebase, he can easily running related unit testing in a few seconds locally to ensure do not break any other functionality.

But that doesn't mean we don't need e2e testing anymore, in 'Test Pyramid', we need to use unit testing as primary tool to ensure each component/page is working well, but we still need some of e2e testing to ensure the 'connection' of them is working well as well. We need to combine these two kind of testing to ensure whole project is working well.

### Testing Hook Functionality
Another thing I noticed is that all testing cases have some same logic like `openHomePage` / `waitEditorLoad`. These kind of things can be placed in Testing hook like this:
```typescript
test.beforeEach(async ({ page }) => {
    await openHomePage(page);
    await waitEditorLoad(page);
});
```

One step further, we can use `test.description` to group related testing cases for better readability and maintainability:
```typescript
test.describe('page delete in current page', () => {
  let homePageId: string;

  test.beforeEach(async ({ page }) => {
    homePageId = await openHomePage(page);
    await waitEditorLoad(page);
  });

  test('should delete current page successfully', async ({ page }) => {}

  // other cases
}

test.describe('page delete in all list page', () => {
  let homePageId: string;

  test.beforeEach(async ({ page }) => {
    homePageId = await openHomePage(page);
    await waitEditorLoad(page);
  });

  test('should delete the first page successfully', async ({ page }) => {}

  // other cases
}
```

### Testing Assertion Issue
Many testing cases are using `toBeUndefined` assertion, but this assertion will be useless if not correct. here is a example:

in testing file `local-first-delete-page.spec.ts`, there is a assertion like this:
```typescript
expect(
    page.getByText(
      'Tips: Click Add to Favorites/Trash and the page will appear here.'
    )
  ).not.toBeUndefined();
```
But if you use playwright ui mode to run this case, you will see this assertion is useless, it will pass all the time. That means, this testing case is a fake case, since it will pass no matter what happend. And this action does not delete the corresponding page:
```typescript
// permanently delete it
  await page
    .getByTestId('more-actions-' + newPageId)
    .getByRole('button')
    .nth(1)
    .click();
  await page.getByText('Delete permanently?').dblclick();
```
tests

But I don't think this is anyone's mistake, just a 'trap' from library, I would suggest do not use this 'undefined' assertion in the future, can use something like `toBeInViewport` instead.

### In The End
I'd like to say it again(叠甲 hahaha), I'm not blaming anyone here. Just giving some of my thought about project from community side. Hopefully you guys can make a great tool:hugs:.

Issue #3335, opened 2023-07-20
plugin infra data storage API


himself65:
Remove these code

https://github.com/toeverything/AFFiNE/blob/19055baa4902dbe60456388b8bdca149b00543af/apps/core/src/bootstrap/register-plugins.ts#L95-L106

PR #3447, opened 2023-07-28
feat: plugin infra indexdb

danielchim:
Replacing #3335's issue with our own API


himself65 (changes requested):
could you please use https://github.com/jakearchibald/idb-keyval/blob/main/custom-stores.md

PR #3474, opened 2023-07-31
chore: clean useless dependency injection

thorseraq:


Brooooooklyn:
Close in https://github.com/toeverything/AFFiNE/pull/3475

Issue #3454, opened 2023-07-29
Login button UI incorrect


himself65:
![image](https://github.com/toeverything/AFFiNE/assets/14026360/cbe325db-b0cf-479a-86d5-407b2b65703b)

Kaneki-Ken-7:
what is the correct design of this button? @himself65

himself65:
see https://www.figma.com/file/hVxJQtEqyv5xw1AbCzc50M/%F0%9F%97%92%EF%B8%8F-AFFiNE-2023?type=design&node-id=2-1563&mode=design&t=aPmDZH7yacDZUSaT-0

Kaneki-Ken-7:
@himself65 can you check my PR whether its the right UI or not?

PR #3477, opened 2023-07-31
fix: error style

QiShaoXuan:
fixed: https://github.com/toeverything/AFFiNE/issues/3454

### DIFF
--- a/apps/core/src/components/affine/setting-modal/account-setting/index.tsx
+++ b/apps/core/src/components/affine/setting-modal/account-setting/index.tsx
@@ -1,4 +1,4 @@
-import { Button, FlexWrapper, IconButton, Input } from '@affine/component';
+import { FlexWrapper, IconButton, Input } from '@affine/component';
 import {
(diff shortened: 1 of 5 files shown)


apps/core/src/components/affine/setting-modal/account-setting/index.tsx
```diff
@@ -118,32 +118,32 @@ export const AccountSetting = () => {
...
-        disabled={true}
-      >
-        <Button
-          onClick={() => {
-            toast('Function coming soon');
-          }}
-        >
-          {t['com.affine.settings.password.action']()}
-        </Button>
-      </SettingRow>
```
himself65, line 146: You comment all of this? Sure?

QiShaoXuan: These function is not support in this version，and disable style not work well， They will come back with the new sign in function

himself65 (approved):

Issue #3433, opened 2023-07-28
Edgeless mode font loading problem


JimmFly:
### What happened?

When I open the client for the first time, the font will change when I create text in edgeless mode and perform a bold operation. After several attempts it returned to normal.

https://github.com/toeverything/AFFiNE/assets/102217452/396440a8-1f7f-4848-afd0-812ce856374c

### Distribution version

macOS ARM 64 (Apple Silicon)

### What browsers are you seeing the problem on if you're using web version?

Chrome

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

Kaneki-Ken-7:
@JimmFly can you check the PR ?

JimmFly:
> 

Wow, thanks for reminding me, I'll check it out.

JimmFly:
> @JimmFly can you check the PR ?

Wow, thanks for reminding me, I'll check it out.

PR #3482, opened 2023-07-31
fix: fixed Edgeless font loading problem

Kaneki-Ken-7:
Fixes : #3433


himself65 (changes requested):
First, this pr doesn't fix the problem. 

Second, it also doesn't fit the rule of React, it should be something like

```tsx
const loadPromise = document.fonts.load('16px Kalam')

const App = () => {
  // ...
  use(loadPromise)
  // ...
}
```

PR #3504, opened 2023-08-01
feat: support sign with password & user info operation

QiShaoXuan:
Still some detail need implement: https://github.com/toeverything/AFFiNE/issues/3556


apps/server/src/modules/auth/guard.ts
```diff
@@ -92,8 +92,10 @@ class AuthGuard implements CanActivate {
         return false;
       }
 
-      // @ts-expect-error body is user here
-      req.user = body.user;
+      req.user = await this.prisma.user.findUnique({
```
Brooooooklyn, line 95: duplicate database query here

apps/server/src/modules/auth/next-auth-options.ts
```diff
@@ -217,10 +219,17 @@ export const NextAuthOptionsProvider: FactoryProvider<NextAuthOptions> = {
             session.user.id = user.id;
             // @ts-expect-error Third part library type mismatch
             session.user.image = user.image ?? user.avatarUrl;
+            // @ts-expect-error Third part library type mismatch
+            session.user.hasPassword = Boolean(user.password);
           } else {
             // technically the sub should be the same as id
             // @ts-expect-error Third part library type mismatch
             session.user.id = token.sub;
+            const databaseUser = await prisma.user.findUnique({
```
Brooooooklyn, line 228: duplicate database query here

Brooooooklyn: `user` in this function params should be the same with the `prisma.user.findUnique`

apps/server/src/modules/auth/guard.ts
darkskygit: user session was filter by next-auth, we dont has enough information

packages/component/src/components/auth-components/change-email-page.tsx
```diff
@@ -0,0 +1,89 @@
...
+type User = {
+  id: string;
+  name: string;
+  email: string;
+  image: string;
+};
+function validateEmail(email: string) {
+  return new RegExp(
+    /^(?:(?:[^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@(?:(?:\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|((?:[a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
+  ).test(email);
```
darkskygit, line 17: extract regex init as a top level const

himself65 (approved):
In the last meeting, we saw some problem that if you have 2 tabs and you logout the account in the second one, the first tab will crash if you are opening account profile.

Please fix this issue

QiShaoXuan:
> In the last meeting, we saw some problem that if you have 2 tabs and you logout the account in the second one, the first tab will crash if you are opening account profile.
> 
> Please fix this issue

Fixed in https://github.com/toeverything/AFFiNE/pull/3504/commits/9768d2c4f92c0e42d998c687dc0b9ce231ef3ea9

Issue #3410, opened 2023-07-27
"Drag page to sidebar" does not work


PaxtonXuan:
https://github.com/toeverything/AFFiNE/assets/134894293/9fcd7280-c391-4f8e-a9c9-e1fa0b7e53b3

PaxtonXuan:
https://github.com/toeverything/AFFiNE/assets/134894293/128a5ad8-597d-47e6-8df0-fb0e7436246d

houjoe0829:
As our expectation:
1. Could drag and add to Favourites.
2. Could adjust the order in Favourites.

teobler:
to support sortable list, there are several ways:
1. just like collection pages, we can save favorite pages as an array to idb
2. add an array or related data structure to `workspace` to support sorted list
3. add two fields like `prevFavoritePage, nextFavoritePage` to make pageMeta a linked list

not sure if there is more option we can choose. cc: @himself65

himself65:
> 

I believe this should be saved in the CRDT to sync to all clients. So you might need to know how to use `workspace.meta`

teobler:
Sure, this is what I proposed in option 2, then I will create an `array` in `workspace.meta` maybe named `favoritePages`

Garfield550:
@houjoe0829 Is this "drag and drop design" still effective?

<img width="1215" alt="image" src="https://github.com/toeverything/AFFiNE/assets/3471836/9aad6c97-3051-467f-bca0-e9fb9e3d6d6a">

houjoe0829:
> @houjoe0829 Is this "drag and drop design" still effective?
> 
> <img alt="image" width="1215" src="https://user-images.githubusercontent.com/3471836/266239204-9aad6c97-3051-467f-bca0-e9fb9e3d6d6a.png">

Thank you for your attention. This design draft has expired and requires @xell  to update the new design for dragging to the sidebar.

Garfield550:
> > @houjoe0829 Is this "drag and drop design" still effective?
> > <img alt="image" width="1215" src="https://user-images.githubusercontent.com/3471836/266239204-9aad6c97-3051-467f-bca0-e9fb9e3d6d6a.png">
> 
> Thank you for your attention. This design draft has expired and requires @xell to update the new design for dragging to the sidebar.

Okay, this issue is currently blocked. I will start working after the new design is completed.

PR #3509, opened 2023-08-01
feat: support drag page from all page list #3410

teobler:
> As our expectation:
> 
> 1. Could drag and add to Favourites.
> 2. Could adjust the order in Favourites.

related to #3410, in progress in my free time, if it urgent, feel free to modify and merge it.

- [x] Could drag and add to Favourites.
- [ ] Could adjust the order in Favourites.
    - [ ] support sortable from all page list
    - [ ] support sortable from favourites


himself65 (commented):
@Garfield550 will handle this when avaliable

teobler:
> @Garfield550 will handle this when avaliable

**Context sharing:**
During previous development, I discussed the solution with @zzj3720 & @doodlewind on Telegram.

This function need to sync between each platform like browser & desktop for single user, but the favorite pages should not sync between users, they should be user specific. That's means we need to get this down with cloud supporting.

And @doodlewind suggested we can create a data structure like awareness data in yjs, and make it persistent. I paused the development since I have no context about cloud data.

More info can refer here: https://t.me/AffineDevCN/4567

But if you guys are already at the same page, just ignore me then.

CC: @himself65 @Garfield550

EYHN:
Some functions have been implemented in https://github.com/toeverything/AFFiNE/pull/6302

PR #3511, opened 2023-08-01
feat: change useCurrentUser to useUserAtom

QiShaoXuan:
The old `useCurrentUser` hook is a wrapper based on next-auth. Now it is replaced with a custom `userAtom` to serve as the central handler for user information, for the following reasons:

1. next-auth cannot fetch custom user data, such as hasPassword.
2. next-auth cannot refresh the user status when not logged in (the page must be reloaded).


apps/server/src/modules/users/resolver.ts
```diff
@@ -108,6 +108,41 @@ export class UserResolver {
     });
   }
 
+  @Mutation(() => UserType, {
+    name: 'updateUser',
+    description: 'Upload user avatar',
+  })
+  async updateUser(
```
Brooooooklyn, line 115: This API is already implemented in `updateSession`

Issue #3532, opened 2023-08-03
Wrong padding in the Switch button of Page/Edgeless


xell:
The left and right padding should be the same as top and bottom. 

![image](https://github.com/toeverything/AFFiNE/assets/132558/50459d74-387d-45c4-8ed1-db970614768a)

Here's the design:

<img width="383" alt="CleanShot 2023-08-03 at 10 01 49@2x" src="https://github.com/toeverything/AFFiNE/assets/132558/7f695bfc-762a-45f2-a113-aed307cd0136">

Kaneki-Ken-7:
@xell I have made the changes. can you check the PR?

xell:
> @xell I have made the changes. can you check the PR?

Thank you for the contribution. Our developers will look into it.

PR #3542, opened 2023-08-03
fix: padding in the Switch button of Page/Edgeless

Kaneki-Ken-7:
Fixes : #3532

added 4px padding and 8px gap.


JimmFly (approved):
Much obliged for your contribution, splendid work indeed.

PR #3543, opened 2023-08-03
fix: editor-mode-switch animation should only run once

Garfield550:
https://github.com/toeverything/AFFiNE/assets/3471836/77d10598-e6f7-4c3b-b117-db83ef259a4d

image


himself65 (commented):
You don't need to rebase every time when the master is updated. We have the merge queue to make sure it's always the latest

lawvs (approved):
Thanks!

PR #3555, opened 2023-08-03
feat: should hide downloadtip when it had been closed

tzhangchi:
https://github.com/toeverything/AFFiNE/assets/5910926/ebfe137b-32dc-49da-9275-5deacbe4c1d2


himself65 (changes requested):
This is by design

tzhangchi:
> This is by design

I change the  product design for affine.fail testing .If After affine.fail is officially launched, this method of client download and drainage is not required

PR #3590, opened 2023-08-05
fix: disable button when value is empty

### LINKED ISSUE #3534: "Save" button should be disabled when there's no name given to the New Collection
When the name filed is blank, i.e. before the user entering a name for the New Collection, the visual appearance and function of the Save button should be disabled. Currently, it's clickable with no effect.

image

Wzb3422:
Fixes #3534 

This pull request resolves the [issue#3534](https://github.com/toeverything/AFFiNE/issues/3534)


himself65 (changes requested):
useMemo here

packages/component/src/components/page-list/view/create-collection.tsx
```diff
@@ -274,9 +274,9 @@ export const EditCollection = ({
           size="large"
           data-testid="save-collection"
           type="primary"
-          disabled={isNameEmpty(value)}
+          disabled={isNameEmpty}
           onClick={() => {
```
himself65: Could you please use useCallback here? Thx

Wzb3422:
> useMemo here

Sure. `useMemo` does a better job in memorization. I've refactored my code with `useMemo`.

packages/component/src/components/page-list/view/create-collection.tsx
Wzb3422: Done. Refactored this implementation with `useCallback`.

himself65 (approved):

PR #3591, opened 2023-08-05
feat: add local blocksuite debug support

fourdim:


apps/core/.webpack/config.ts
```diff
@@ -121,6 +123,65 @@ export const createConfiguration: (
...
+      modules: [
+        // Root directory's node_modules takes precedence over other node_modules.
+        resolve(rootPath, '../../', 'node_modules'),
+        'node_modules',
+      ],
+      fallback:
+        blocksuiteBaseDir === undefined
+          ? undefined
+          : {
+              events: require.resolve('events'),
```
himself65: why we have this

himself65 (changes requested):

apps/core/.webpack/config.ts
fourdim: Module `events` is introduced in blocksuite, I have removed it by `event: false`

apps/core/.webpack/config.ts
```diff
@@ -121,6 +123,60 @@ export const createConfiguration: (
...
+                blocksuiteBaseDir,
+                'packages',
+                'phasor'
+              ),
+              '@blocksuite/store/workspace/migration/migrate-block': resolve(
+                blocksuiteBaseDir,
+                'packages',
+                'store',
+                'src/workspace/migration/migrate-block'
+              ),
```
fourdim, line 168: If bump blocksuite to https://github.com/toeverything/blocksuite/commit/36f1cf5de478d20dda9c4da41bbad1cdac5de439 and use the new migration api, these lines can be removed.

himself65 (approved):

PR #3645, opened 2023-08-09
feat: support get datasource status

himself65:
We need a way to display the status of each data source

Especially in the cloud data source, there will have network pending and server error. There must have a way to block the UI


packages/y-indexeddb/src/provider.ts
```diff
@@ -29,12 +30,24 @@ const createDatasource = ({
...
+    callbackSet.forEach(cb => cb());
+  };
+
   const dbPromise = openDB<BlockSuiteBinaryDB>(dbName, dbVersion, {
     upgrade: upgradeDB,
   });
   const adapter = {
     queryDocState: async (guid, options) => {
       try {
+        changeStatus({
```
pengx17, line 48: I am not sure but it seems the way setting status may has race condition issues

packages/y-provider/src/types.ts
```diff
@@ -1,3 +1,18 @@
+export type Status =
```
himself65, line 1: Can we use `xstate` to handle this?

packages/y-indexeddb/src/provider.ts
himself65: I think you are right. we need a mutex lock here

himself65: After I tried, I think it's not necessary to add a mutex here

himself65: I simulate a stack, it works great

packages/y-provider/src/types.ts
```diff
@@ -1,4 +1,24 @@
...
+  | {
+      type: 'syncing';
+    }
+  | {
+      type: 'synced';
+    }
+  | {
+      type: 'error';
+      error: Error;
+    };
```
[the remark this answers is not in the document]
himself65: I think the `Status` here is just a signal for the UI. In the design part, we don't have such demand on display the sync details to the user

PR #3664, opened 2023-08-10
fix: remove workspace.list in CRUD

pengx17:
This will fix client cloud syncing issue


packages/workspace/src/local/crud.ts
```diff
@@ -98,26 +98,10 @@ export const CRUD: WorkspaceCRUD<WorkspaceFlavour.LOCAL> = {
...
+      typeof schema
+    >;
 
-    // workspaces in desktop
-    if (
-      window.apis &&
-      environment.isDesktop &&
-      runtimeConfig.enableSQLiteProvider
-    ) {
-      const desktopIds = (await window.apis.workspace.list()).map(([id]) => id);
```
pengx17, line 113: turns out this is not necessary

Issue #3602, opened 2023-08-07
add the content menu function for title side


hwangdev97:
### Description

![shot-2023-08-07-16 19 26@2x](https://github.com/toeverything/AFFiNE/assets/24713927/f696220f-48b2-4485-89f0-ff7b2adcd388)

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

himself65:
@doodlewind BTW there's an existing error when you double-click the title in the latest Canary

JimmFly:
The function of adding tags requires upstream support. /cc @zzj3720

QiShaoXuan:
@xell Now only the function of Add Tag is missing. I think we can open a new issue for recording and assign it to @zzj3720

xell:
> @xell Now only the function of Add Tag is missing. I think we can open a new issue for recording and assign it to @zzj3720

@houjoe0829 If the "Add Tag" function will be implemented in another issue, I think this one could be closed as long as the UI part is confirmed.

houjoe0829:
Close with another follow-up https://github.com/toeverything/AFFiNE/issues/3867

PR #3675, opened 2023-08-10
feat: add duplicate feature

JimmFly:
Work for #3602 

https://github.com/toeverything/AFFiNE/assets/102217452/da7e3b38-a6dc-48d8-8df7-8ecac86fc29c


hwangdev97:
I checked the UI detail. It's 👌

PR #3683, opened 2023-08-11
ci: use ci env to overwrite preset build preset

thorseraq:


himself65 (changes requested):
What's the use case?

thorseraq:
> What's the use case?

https://github.com/toeverything/AFFiNE/blob/7f542bbae8577b4de4e4e9cc4070c55416191855/.github/workflows/build.yml#L140, we hope to inject some build presets from CI env

Issue #3491, opened 2023-08-01
All Pages: 3 UI bugs


xell:
- [x] the three-button-group does not align center

![Untitled](https://github.com/toeverything/AFFiNE/assets/132558/b8f9aa11-0168-4e3a-a87c-800daa588d7d)

- [x] the hover color should be `hoverColorFilled` in the Add button

![image](https://github.com/toeverything/AFFiNE/assets/132558/89f07ec6-583c-4ef7-85f5-8c4459a5f15e)

- [ ] the overlap problem seen in the above screenshot

Kaneki-Ken-7:
@xell in the first task you want all the 3 buttons to have same width or something else?

JimmFly:
i will handle it.

JimmFly:
Can you check the PR？

JimmFly:
Done.

PR #3685, opened 2023-08-11
feat: refator header

QiShaoXuan:
Components abstracted out  is used to solve the following problems
1. Manage layout issues independently of page or business logic
2. Dynamic centered middle element (relative to the main-container), when the middle element is detected to collide with the two elements, the line wrapping process is performed


JimmFly:
#3491  /cc @xell

himself65 (approved):

PR #3688, opened 2023-08-11
feat: add Intl.Segmenter polyfill

fourdim:


apps/core/src/polyfill/intl-segmenter.ts
```diff
@@ -0,0 +1,9 @@
+if (Intl.Segmenter === undefined) {
+  import('intl-segmenter-polyfill-rs').then(({ Segmenter }) => {
```
himself65: use top-level await here

himself65 (approved):

Fathertime179 (commented):
Good job

Issue #3593, opened 2023-08-06
developer document need update


himself65:
```[tasklist]
### Basic
- [ ] how to build @affine/core
- [ ] how to build @affine/electron
- [ ] the concept of AFFiNE
- [ ] debug with blocksuite
- [ ] how to setup @affine/server
- [ ] run in docker
- [ ] how to contribute
- [ ] how to use @affine/storybook
```

```[tasklist]
### Plugins(WIP)
- [ ] @affine/sdk
- [ ] how to create a plugin
- [ ] plugin restriction
```

imSuMN:
@himself65  Can you guide my what to do? I am new to open source (!coding)😄

teobler:
![affine](https://github.com/toeverything/AFFiNE/assets/26424691/ba2cec1c-7a6a-42b9-9138-85c8dd910749)

To understand affine's high level concepts and infra, I made a draft diagram refer to source code based on my own understanding, if this is helpful, I can complete it and share to community.

HeJiachen-PM:
> ![affine](https://user-images.githubusercontent.com/26424691/259272085-ba2cec1c-7a6a-42b9-9138-85c8dd910749.png)
> 
> To understand affine's high level concepts and infra, I made a draft diagram refer to source code based on my own understanding. There maybe something wrong, but if this is helpful, I can complete it first, then someone from team can help me to review it, after that we can share it with community.

Hello teobler, your diagram looks amazing and in general good to me. Regarding your questions:
1. No indeed. The cloud is optional.
2. Correct but we hope the indexDB can be replaced with SQLite wrapped by octobase in the electron client.
3. This is a major place that needs to be better built by an application data layer and state management tools, to where @himself65 @Brooooooklyn may have many things to add.
4. Currently a plugin system, a complete data driven layer, and cloud support.
5. In general, we hope our plugins can be similar to vscode's plugins where it is hosted on a dedicated plugin store, and its access to data and the network can be totally controlled by the plugin system.

I'd better leave the rest part to other members. Hope this great diagram can be spread for the good of others as well!

teobler:
> > ![affine](https://user-images.githubusercontent.com/26424691/259272085-ba2cec1c-7a6a-42b9-9138-85c8dd910749.png)
> > To understand affine's high level concepts and infra, I made a draft diagram refer to source code based on my own understanding. There maybe something wrong, but if this is helpful, I can complete it first, then someone from team can help me to review it, after that we can share it with community.
> 
[...]

PR #3697, opened 2023-08-11
doc: add doc for concepts introduction(#3593)

teobler:
related to #3593 

all knowledge based on my understanding, if anything wrong, just let me know or feel free to fix it.


himself65 (changes requested):
Thanks for your contribution to the documents.

You have a good understanding of how our project works. But I think it's not very precise on each part. I will handle this PR later

Issue #3660, opened 2023-08-10
Page level references only show link references in paragraphs​


pengx17:
### What happened?

Link references in the fav menu only shows link references in the paragraphs. Need to show link references in database, lists etc.

### Distribution version

macOS x64 (Intel)

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

PR #3702, opened 2023-08-11
fix: page references in list/database

pengx17:
fix https://github.com/toeverything/AFFiNE/issues/3660
image


himself65 (approved):

himself65 (commented):
test is needed

PR #3705, opened 2023-08-11
chore: add virtual list support

qiuqfang:


himself65 (changes requested):
Thanks for your contribution. But I'm thinking to refactor this component using tanstack/table or similar library

packages/component/src/components/page-list/use-virtual-table-height.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useState } from 'react';
+
+export function useVirtualTableHeight() {
+  const [virtualTableHeight, setVirtualTableHeight] = useState(0);
+  useEffect(() => {
+    const mainEl = document.querySelector('.main-container') as HTMLElement;
+    const headEl = document.querySelector('.header-container') as HTMLElement;
```
himself65: prohibit using the constant value

himself65 (changes requested):

packages/component/src/components/page-list/use-virtual-table-height.ts
qiuqfang: Okay, but for what reason?

himself65: not a good engineering approach. I think we pass two `ref<HTMLElemet>` in parameters is better.

And a hook function should be pure, which also means it only depends on the inputs(parameters)

qiuqfang: > not a good engineering approach. I think we pass two `ref<HTMLElemet>` in parameters is better.
> 
> And a hook function should be pure, which also means it only depends on the inputs(parameters)

I've modified it.

apps/core/src/components/blocksuite/workspace-header/header.tsx
```diff
@@ -172,7 +172,7 @@ export const Header = forwardRef<HTMLDivElement, HeaderProps>((props, ref) => {
 
   return (
     <div
-      className={styles.headerContainer}
+      className={(styles.headerContainer, 'header-container')}
```
himself65: use clsx

himself65 (dismissed):

himself65 (approved):
LGTM

packages/component/src/components/page-list/mobile.tsx
```diff
@@ -68,6 +70,11 @@ export const AllPageListMobileView = ({
   createNewEdgeless: () => void;
   importFile: () => void;
 }) => {
+  const virtualTableHeight = useVirtualTableHeight(
+    '.main-container',
```
himself65: Shouldn't use css selector here. We might have multiple page list in a single web

himself65 (changes requested):

packages/component/src/components/page-list/all-page.tsx
```diff
@@ -167,12 +169,15 @@ export const PageList = ({
   getPageInfo,
   propertiesMeta,
 }: PageListProps) => {
+  const virtualTableHeight = useVirtualTableHeight(
+    '.main-container',
+    '.header-container'
```
himself65: We dont have `header-container` classname in the code

himself65 (changes requested):

packages/component/src/components/page-list/virtual-table.tsx
```diff
@@ -0,0 +1,90 @@
...
+        innerElementType={InnerElement}
+        onItemsRendered={({
+          overscanStartIndex,
+          overscanStopIndex,
+          visibleStartIndex,
+          visibleStopIndex,
+        }) => {
+          if (!listRef.current) return;
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
```
himself65: Remove this

packages/component/src/components/page-list/virtual-table.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+const VirtualTableContext = React.createContext<{
+  header: React.ReactNode;
+  footer: React.ReactNode;
+  top: number;
+}>({
+  header: <></>,
+  footer: <></>,
+  top: 0,
+});
```
himself65: use atom instead

packages/component/src/components/page-list/use-virtual-table-height.ts
```diff
@@ -0,0 +1,31 @@
+import { useEffect, useState } from 'react';
+
+export function useVirtualTableHeight(
+  mainSelectors: string,
+  headSelectors: string
```
himself65: use `ObjectRef<HTMLElement>`

himself65 (changes requested):

packages/component/src/components/page-list/use-virtual-table-height.ts
himself65: I will handle this

pengx17:
Hi @qiuqfang , thanks for your contribution! We are replacing the pages using `div` instead of `table` in next version and your current PR will be deprecated.
Will give credits to you in the new implementation. Thanks again!

Issue #3245, opened 2023-07-14
BlobStorage should provide consistent behaviour


zuoxiaodong0815:
The 'get' method of 'createIndexeddbStorage' returns the 'type' type. But, createSQLiteStorage and createStaticStorage do not. Please take time to add 'type' type. 

Please let me know if it is completed. I need to remove the temporary code. https://github.com/toeverything/blocksuite/pull/3555

NextThread:
hey, I'd like to work on this issue

zuoxiaodong0815:
Thank you. It has been assigned to you. @NextThread

NextThread:
@zuoxiaodong0815 i did, but somebody told me that, it's not necessary, this thing can be skipped

zuoxiaodong0815:
If we don't store the file type, we need to calculate the file type when reading the file, refer to the code https://github.com/toeverything/blocksuite/pull/3555/files
<img width="918" alt="image" src="https://github.com/toeverything/AFFiNE/assets/53252747/cbda951a-d2d5-430e-abe6-4f63db96fbe9">
So storing file types is a bit more appropriate.

The following two blob-storage, one of which provides the file type and one of which does not
https://github.com/toeverything/blocksuite/blob/master/packages/store/src/persistence/blob/indexeddb-storage.ts
https://github.com/toeverything/AFFiNE/blob/master/packages/workspace/src/blob/sqlite-blob-storage.ts

NextThread:
ok I'm raising a PR again. Thanks

zhe-he:
@zuoxiaodong0815  
     Hi, I noticed this issue because my editor keeps giving me warnings about the `file-type` module. Then I started trying to fix this problem. Since I'm not familiar with the business, if my approach is incorrect, please consider upgrading the `file-type` module so that my editor no longer gives the following warnings. Thank you in advance.
```
[plugin:vite:resolve] Module "events" has been externalized for browser compatibility, imported by "/Users/hezhe/work/processon/processon-prototype/node_modules/readable-web-to-node-stream/node_modules/readable-stream/lib/_stream_readable.js". See http://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details
```

fourdim:
[...]

PR #3729, opened 2023-08-14
Added 'type' type to CreateStaticStorage & CreateSQLiteStorage

NextThread:


himself65 (changes requested):
image

This is not related to whether you are the first PR. I think you never respect open source because I have told you this is useless change.

NextThread:
> image
> 
> This is not related to whether you are the first PR. I think you never respect open source because I have told you this is useless change.

I respect,
please refer to this- https://github.com/toeverything/AFFiNE/issues/3245#issuecomment-1676588153

I was confused

PR #3741, opened 2023-08-14
fix(plugin): allow multiple loads assets

keepcalmqqf:


himself65 (approved):

himself65:
Nice catch, I didn't notice this bug

PR #3780, opened 2023-08-16
refactor: remove setTimeout inside of delete workspace input ref callback

Garfield550:
https://github.com/toeverything/AFFiNE/assets/3471836/deb60e30-bc7a-47f9-b06d-1a7d73ee8272


Garfield550:
This is not a suitable solution.

Issue #3772, opened 2023-08-16
Unable to edit page title in top bar


PaxtonXuan:
https://github.com/toeverything/blocksuite/assets/134894293/0dcb72b3-ebd7-4af8-98b7-1c1a7221a69e

Issue #3762, opened 2023-08-15
Fix some sidebar alignment issues​


hwangdev97:
### What happened?

![shot-2023-08-15-19 47 42@2x](https://github.com/toeverything/AFFiNE/assets/24713927/0f7e90fa-76ff-45f3-865f-eba4f40ff725)

[the Design file link](https://www.figma.com/file/UTTtfVCcStw1dKUg0gCSPo/%F0%9F%9B%9E-Design-System?type=design&node-id=1185%3A10133&mode=design&t=gQoXvFlLRql8AE5G-1)

- [ ] fix the sidebar‘s padding
- [ ] Modify the style of the non-expandable list item and add an unavailable state for the collapse button
![shot-2023-08-15-19 52 59@2x](https://github.com/toeverything/AFFiNE/assets/24713927/e1d2e568-5eed-4644-a698-aaa118754466)

- [ ] Modify the display of more buttons: display when changing from constant display to hover
![shot-2023-08-15-19 53 53@2x](https://github.com/toeverything/AFFiNE/assets/24713927/36d0ad28-a8b8-4bab-bec5-50b4a73aaf95)

JimmFly:
@pengx17  is working on this.

PR #3783, opened 2023-08-16
fix: app sidebar ui issues

pengx17:
fix https://github.com/toeverything/AFFiNE/issues/3762
fix https://github.com/toeverything/AFFiNE/issues/3772
also fixed an issue that may prevent sidebar from resizing


himself65 (changes requested):
Fix DeepScan error

himself65 (approved):

Issue #3714, opened 2023-08-12
Set `Client Border Style` to false by default on Windows platform


himself65:
It's ugly on Windows

AlfredRufus:
Hey can I solve  this issue. Can you explain  this more clearer

shiivamtaneja:
Hi @himself65  does the issue still persists?

Noothan-am:
is this issue still open?

himself65:
This still open, yeah

hwangdev97:
Windows does not support Mica/Acrylic materials, so I think this option can be turned off by default 

About Mica/Acrylic https://github.com/toeverything/AFFiNE/issues/2399

shiivamtaneja:
Hi @himself65 can you explain about this issue? what exactly u mean by it?

himself65:
I mean just set `client border style` to false by default if `environment.isWindows` is true. This doesn't mean that always set to false when the user refreshes the page(so this is incorrect https://github.com/rohitpawalia/AFFiNE/commit/83de93fe6828578f24703167f0bb5916769e58d3). Just the "default setting"

Asish-AP:
> 



> I mean just set `client border style` to false by default if `environment.isWindows` is true. This doesn't mean that always set to false when the user refreshes the page(so this is incorrect [rohitpawalia@83de93f](https://github.com/rohitpawalia/AFFiNE/commit/83de93fe6828578f24703167f0bb5916769e58d3)). Just the "default setting"

You're suggesting that , This change should only apply to the initial appearance and shouldn't persist if the user refreshes the page. Please do correct me if I am wrong I am new to Open Source

PR #3811, opened 2023-08-17
Fix Default Client Border Style Setting on Windows Platform (#3714)

rohitpawalia:
made changes to:
apps/core/src/components/affine/setting-modal/general-setting/appearance/index.tsx

set clientBorder to false in defaultAppSettings


himself65 (changes requested):
This is not how things work. The correct behavior should let user can save their perferences

himself65:
See https://github.com/toeverything/AFFiNE/pull/3960#issuecomment-1694496240

PR #3848, opened 2023-08-20
fix: recent pages list doesn't update

ideakaran:
I think the issue appeared because "mode" was coming as undefined in utils.tsx file. 
So, set a default value of 'page' if mode is undefined.  With that change I was able to see the newly created page in recent list.

The other case is, When we click on existing page, it did not appear in the recent list. So, I performed setPageMode in detail-page.tsx.


himself65 (commented):
Could you please also add a test case?

himself65 (changes requested):
I tried, seems still not correct

himself65 (approved):
My bad, this PR does fix the bug

Issue #3732, opened 2023-08-14
AFFiNE cloud data migration


thorseraq:
```[tasklist]
### Tasks
```

himself65:
I'm moving the migration logic into the infra layer. Hope I can do it soon

thorseraq:
Currenly I directly used the subdoc part:

[const migratedDoc = migrateToSubdoc(originalDoc);](https://github.com/toeverything/AFFiNE/pull/3869/files#diff-7ba6fdc4e7e512e75f1065acbbec41d5f0a2f918f6a2907548779cec658532faR24)

PR #3869, opened 2023-08-21
feat: YDoc migration script

thorseraq:
upstream issues:  #3732 

## Scope

From Editor Version: 0.0.0-20230420210727-85b7de79-nightly to latest

## TODO
- [x] Migrate from YDoc generated by previous AFFiNE Cloud to AFFiNE Client 0.8.0
- [x] Verify the migation result with BlockSuite playground
- [x] Insert migrated YDoc info into incoming AFFiNE Cloud storage

## Known issue

- Lines drawn by Brush(Pen) will lose
- Sometimes connector will lose, ie, when in old AFFiNE Cloud editor, connector source and target were not anchored with other elements

### Edgeless
image
image

### Blog
image


packages/doc-migration/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "doc-migration",
+  "packageManager": "yarn@3.6.0",
```
himself65: no this

.yarn/patches/@blocksuite-store-npm-0.0.0-20230814155455-ceb5d5d8-nightly-8b4457adbc.patch
```diff
@@ -0,0 +1,19 @@
...
+-        currentVersion,
+-        `previous version for flavour ${flavour} not found`
+-      );
++      if (!currentVersion) {
++        console.warn(`previous version for flavour ${flavour} not found`);
++        return;
++      }
+       this.upgradeBlock(flavour, currentVersion, block);
+     });
+   };
```
himself65: Please contribute to the upstream directly

himself65 (changes requested):

himself65 (commented):
I wonder if we can abstract migration logic to the top level so desktop, web, server can use the same logic

.yarn/patches/@blocksuite-store-npm-0.0.0-20230814155455-ceb5d5d8-nightly-8b4457adbc.patch
thorseraq: Concerning when this migration part is running in the browser, throw might be a good behavior, I removed the patch and add a try catch in migration script

thorseraq:
> I wonder if we can abstract migration logic to the top level so desktop, web, server can use the same logic

Currently the script is not well-abstracted enough, but I think it's doable, a quick idea looks:

- Most of migrations can be treated as normal CRDT operations, which is transparent to different platforms. For this part, we need a stable YDoc Meta structure(which we already have) to confirm the block version, and utils to migrate from version A -> version B, etc

- For the doc -> subdoc part, different storage provider in platforms need implement a common defined interface

Issue #3661, opened 2023-08-10
Single page sharing design updated


Svaney-ssman:
Single page sharing design has been upgraded, affecting the title bar, affine cloud, export, and read-only mode.
 This also relates to [#3287 ](https://github.com/toeverything/AFFiNE/issues/3287)

Here is the [design](https://www.figma.com/file/hVxJQtEqyv5xw1AbCzc50M/%F0%9F%97%92%EF%B8%8F-AFFiNE-2023?type=design&node-id=534-15310&mode=designl)

![Image](https://github.com/toeverything/AFFiNE/assets/110808979/a380a1ac-4769-4637-8d41-f3993c8a2924)

joooye34:
we need to implements this feature new iteration.

Svaney-ssman:
[UI adjusted in figma](https://www.figma.com/file/hVxJQtEqyv5xw1AbCzc50M/%F0%9F%97%92%EF%B8%8F-AFFiNE-2023?type=design&node-id=11259-179690&mode=design&t=X2I8cTOzbj4WEew0-4)
@JimmFly

houjoe0829:
List the outline of the UI upgrade
1. The panel UI of Share page has been changed.
2. The UI of Publicly shared pages has been slightly modified, but the main reason is that the previous development work has not restored the design draft effect, including mobile adaptation.

JimmFly:
Does it sound like the content of this PR#4584  ？

JimmFly:
https://github.com/toeverything/AFFiNE/pull/4856 Close as completed

PR #3871, opened 2023-08-21
feat: add share page menu

JimmFly:
Work for #3661 
Added styling when AFFiNE Cloud is not enabled. The specific logic will be added in the Cloud branch after merging this PR.

https://github.com/toeverything/AFFiNE/assets/102217452/c60cf0f9-4cff-45d6-9484-efb6a1837960


himself65 (changes requested):
Could please rebase the code to v0.9.0-staging? I think we don't have such a feature in 0.8.x

PR #3910, opened 2023-08-23
New features waiting list

Brooooooklyn:


apps/server/src/config/def.ts
```diff
@@ -77,6 +77,10 @@ export interface AFFiNEConfig {
    * System version
    */
   readonly version: string;
+  /**
+   * Deployment environment
+   */
+  readonly affineEnv: 'canary' | 'beta' | 'production';
```
Brooooooklyn, line 83: I'll rename the `dev` namespace in our cluster into `canary` soon

apps/server/src/modules/auth/next-auth.controller.ts
```diff
@@ -127,7 +131,7 @@ export class NextAuthController {
       } else if (action === 'signout') {
         res.send(JSON.stringify({ url: redirect }));
       } else {
-        res.redirect(redirect);
+        res.send(JSON.stringify({ url: redirect }));
```
Brooooooklyn, line 134: match the behavior with: https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/routes/signin.ts#L52

darkskygit (approved):

apps/server/src/modules/auth/next-auth-options.ts
```diff
@@ -261,8 +262,25 @@ export const NextAuthOptionsProvider: FactoryProvider<NextAuthOptions> = {
         }
         return session;
       },
-      redirect(params) {
-        return params.url;
+      signIn: async ({ profile }) => {
+        if (config.affineEnv !== 'beta') {
```
forehalo, line 266: config.beta/canary/prod?

himself65 (approved):

himself65 (approved):

apps/server/src/modules/auth/next-auth-options.ts
```diff
@@ -261,8 +262,25 @@ export const NextAuthOptionsProvider: FactoryProvider<NextAuthOptions> = {
...
+              where: {
+                email: profile.email,
+                type: NewFeaturesKind.EarlyAccess,
+              },
+            })
+            .then(user => !!user)
+            .catch(() => false);
+        }
+        return false;
+      },
```
himself65: I think you shouldn't block user to log in, instead block them from register

himself65 (changes requested):

apps/server/src/modules/auth/next-auth-options.ts
himself65: Or please allow e2e test to skip this whitelist

himself65 (changes requested):

apps/core/src/components/affine/auth/sign-in.tsx
```diff
@@ -37,17 +37,20 @@ export const SignIn: FC<AuthPanelProps> = ({
...
+    if (user) {
+      if (user.hasPassword) {
+        setAuthState('signInWithPassword');
+      } else {
+        signIn('email', {
+          email: email,
+          callbackUrl: buildCallbackUrl('signIn'),
+          redirect: true,
+        }).catch(console.error);
+        setAuthState('afterSignInSendEmail');
```
pengx17, line 52: should this one be put into `signIn.then`?

Brooooooklyn: I'm not sure, it's the original implementation

Brooooooklyn: /cc @QiShaoXuan

PR #3913, opened 2023-08-23
fix(workspace): should load remote state in sync phase

forehalo:
```
client: doc-load ->                  -> send-update(diff)

server:          -> {missing, state}
```

### DIFF
--- a/apps/server/src/modules/doc/manager.ts
+++ b/apps/server/src/modules/doc/manager.ts
@@ -12,16 +12,6 @@ import { Metrics } from '../../metrics/metrics';
 import { PrismaService } from '../../prisma';
 import { mergeUpdatesInApplyWay as jwstMergeUpdates } from '../../storage';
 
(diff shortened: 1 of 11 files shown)


pengx17:
Can we cherry-pick to master first?

forehalo:
> Can we cherry-pick to master first?

I will split the code into commits so they can be easily cherry picked

apps/server/src/modules/doc/manager.ts
```diff
@@ -74,8 +64,24 @@ export class DocManager implements OnModuleInit, OnModuleDestroy {
     this.destroy();
   }
 
+  protected recoverDoc(...updates: Buffer[]): Doc {
+    const doc = new Doc();
+
+    updates.forEach(update => {
+      applyUpdate(doc, update);
```
himself65, line 71: wrap into a single transaction?

```ts
doc.transact(() => {
  // you code
}, 'sourceId')
```

pengx17 (approved):

PR #3921, opened 2023-08-23
ci: try install 3 times

himself65:


himself65:
I think GitHub CI itself has issue on downloading from the internet, closing

PR #3931, opened 2023-08-24
chore: add bump octobase script

darkskygit:


scripts/bump-octobase.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+cargo update -p jwst -p jwst-codec -p jwst-storage
```
himself65: ending backspace

himself65 (changes requested):

Issue #3948, opened 2023-08-25
Optimization of the details of the sidebar​


hwangdev97:
### Description

 
- [ ]  add a circle on right-top of what's new button

People think it's a regular UI, but in fact we want people to open it to see related content：

![shot-2023-08-25-10 06 57@2x](https://github.com/toeverything/AFFiNE/assets/24713927/6ddc0a33-7359-40ec-9262-6d76d6693dbe)

- [ ] when 'Client Border Style' open, we should set the padding to `16px`

![shot-2023-08-25-10 07 33@2x](https://github.com/toeverything/AFFiNE/assets/24713927/5eb23f7e-b406-4ffb-9d76-34b22e826cd5)
 
- [ ]  When the user hovers over the “Update Available” button, it will display “Restart to install update.” After clicking on it, the text should change to “Loading...”

![shot-2023-08-25-10 13 36@2x](https://github.com/toeverything/AFFiNE/assets/24713927/91468a4c-03eb-4e94-a620-5b686ac2eb13)

PR #3950, opened 2023-08-25
fix: left sidebar style fixes

pengx17:
fix https://github.com/toeverything/AFFiNE/issues/3948


apps/core/src/components/root-app-sidebar/index.tsx
```diff
@@ -111,6 +111,7 @@ export const RootAppSidebar = ({
     if (isDesktop) {
       return window.events?.applicationMenu.onNewPageAction(onClickNewPage);
     }
+    return () => {};
```
himself65: noop from foxact/noop

himself65 (changes requested):

apps/core/src/components/root-app-sidebar/index.tsx
pengx17: It's type is 
```ts
interface Noop {
    (...args: any[]): any;
}
```
and ts does not like that

hwangdev97 (approved):

himself65 (approved):

PR #3960, opened 2023-08-25
fix(electron): set client border style to false by default on windows

### LINKED ISSUE #3714: Set `Client Border Style` to false by default on Windows platform

Sarvesh521:
fix #3714

### DIFF
--- a/.github/CLA.md
+++ b/.github/CLA.md
@@ -60,3 +60,4 @@ Example:
 - Moeyua, @moeyua, 2023/04/22
 - Shishu, @shishudesu, 2023/05/19
 - Kushagra Singh, @kush002, 2023/06/28
+- Sarvesh Kumar, @sarvesh521 2023/08/25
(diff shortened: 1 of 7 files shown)


himself65 (approved):

Sarvesh521:
@himself65 can you guide on how to proceed after this point ?

himself65:
> @himself65 can you guide on how to proceed after this point ?

It's weird because I think the tests should pass

himself65:
> > @himself65 can you guide on how to proceed after this point ?
> 
> It's weird because I think the tests should pass

I found that it's our component bug

himself65:
Thanks for your contribution. The initial code, though, doesn't solve the essential problem(actually the PR is the same as https://github.com/toeverything/AFFiNE/pull/3811) but hope you both can get inspired on how to contribute to AFFiNE.

Sarvesh521:
@himself65 I thought adding that extra if condition to check if it was win32 would have solved the problem. Thanks for helping me. Will try to fix other issues with the experienced gained here.

PR #3961, opened 2023-08-26
fix(server): cloud Public Workspace Logic

### LINKED ISSUE #3957: cloud public workspace logic
currently, we have a field `public` in the workspace schema but it's not good if people only share one page.

So the `public` field will have `PUBLIC`, `ALL`, and `PRIVATE` enums now instead of a boolean value

imf-ali:
fix #3957


imf-ali:
Hey @himself65 , can you please have a look at this?

himself65 (changes requested):
There's no database migration

apps/server/schema.prisma
```diff
@@ -8,9 +8,15 @@ datasource db {
...
+enum WorkspaceStatus {
+  PUBLIC
+  ALL
+  PRIVATE
+}
+
 model Workspace {
   id             String                    @id @default(uuid()) @db.VarChar
-  public         Boolean
+  public         WorkspaceStatus
```
himself65: I think it should be a computed value.

himself65 (changes requested):
rebase the PR to `v0.9.x-staging`

imf-ali:
> There's no database migration

Oh, I will add those

himself65:
Thanks for the contribution! I have rewritten the logic so that we don't need to change the schema

See: https://github.com/toeverything/AFFiNE/pull/3997

Issue #3636, opened 2023-08-09
Sidebar： Swap the positions of the forward and backward navigation bar and the receiving side sidebar buttons


hwangdev97:
### Description

by @zanwei :

![shot-2023-08-09-13 42 29@2x](https://github.com/toeverything/AFFiNE/assets/24713927/a92805e9-eb6e-47b3-ba99-22f7c4ec54b9)

![shot-2023-08-09-13 39 50@2x](https://github.com/toeverything/AFFiNE/assets/24713927/ac01f224-0859-4eec-9929-d9b001129b15)

and after resize:

![shot-2023-08-09-13 42 55@2x](https://github.com/toeverything/AFFiNE/assets/24713927/8917aeac-c8ab-44de-b120-da912b96a586)

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

himself65:
> Hey @himself65, I was looking for a good issue as my first contribution and was wondering if I can attempt it? **#3636**

Yes

blacurrant:
Hey @himself65 , I can use a little help here! I did the change but I didn't find the "compare & PR" button and ended up committing it to the forked repo, how should I correct it and submit a PR here?

NamanShah028:
In which file i need to make changes

hwangdev97:
I think we can close this issue, the behavior is now as expected cc @pengx17

PR #3967, opened 2023-08-27
fix: swap navigation bar items

blacurrant:
I swapped the positions of the forward and backward navigation bar and the receiving side sidebar buttons. #3636


himself65 (approved):

himself65 (approved):
Thanks for your contribution!

PR #4008, opened 2023-08-29
fix(server): signup/signin logic

Brooooooklyn:


apps/server/src/modules/sync/redis-adapter.ts
```diff
@@ -30,7 +30,10 @@ export class RedisIoAdapter extends IoAdapter {
   }
 
   override createIOServer(port: number, options?: ServerOptions): any {
-    const server = super.createIOServer(port, options);
+    const server = super.createIOServer(port, {
```
forehalo, line 33: set in `Gateway`, see: https://docs.nestjs.com/websockets/gateways#overview,
and so as the client opts in https://github.com/toeverything/AFFiNE/blob/fffa5b2f5181ce71e2111b2e50dc6712207bcdb5/packages/workspace/src/affine/index.ts#L23

forehalo (approved):

PR #4011, opened 2023-08-29
feat: rate limiter

### LINKED ISSUE #4002: Server rate limit
```[tasklist]
### Rate limits
- [ ] Check user
- [ ] Sign in
- [ ] Sign up
- [x] Invite
```

darkskygit:
- Close #4002


apps/server/src/throttler.ts
```diff
@@ -0,0 +1,34 @@
...
+export class RateLimiterModule {}
+
+@Injectable()
+export class CloudThrottlerGuard extends ThrottlerGuard {
+  override getRequestResponse(context: ExecutionContext) {
+    return getRequestResponseFromContext(context) as any;
+  }
+
+  protected override getTracker(req: Record<string, any>): string {
+    return req.ips.length ? req.ips[0] : req.ip;
```
Brooooooklyn, line 30: using `x-forward-for`? We are behind multi layers proxy in production, the `ips` may be from the proxy

darkskygit: cf also support `X-Forward-For`: https://developers.cloudflare.com/fundamentals/get-started/reference/http-request-headers/#x-forwarded-for
but seem `CF-Connecting-IP` is a better choose, i will switch to this header

PR #4016, opened 2023-08-29
fix: add arch key to setup-node cached files

pengx17:
`runner.arch`
https://github.com/actions/runner/blob/be65955a9d2503060186d1e8d97a87587051161d/src/Runner.Worker/JobRunner.cs#L149

it seems node-setup always gets stuck for mac. Not sure if this is because we save the same key for both arm64 & x86 for mac.

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -76,7 +76,7 @@ runs:
       if: ${{ inputs.playwright-install == 'true' }}
       with:
         path: '~/.cache/ms-playwright'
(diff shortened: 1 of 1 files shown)


pengx17:
Looks like node will store cache regardless the platform's arch. 🤔 
https://github.com/actions/setup-node/blob/main/src/cache-restore.ts#L41C3-L41C64

himself65 (approved):

pengx17:
Sumitted an upstream pr: https://github.com/actions/setup-node/pull/843

PR #4031, opened 2023-08-30
fix(core): migration before connection

### LINKED ISSUE #4029: 0.9.0-canary.3 can not import 0.8.0-canary.27 workspace​
### What happened?

As the title, I overwrite install 0.8.0-canary.27 use 0.9.0-canary.3, but it can not show all the file. I install  0.8.0-canary.27 back, and export workspace. then I install 0.9.0-canary.3, import the workspace, first time it show migration error, second time it finish import, but all the index is wrong and disorder.

### Distribution version

Windows x64

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

himself65:
Fixes: https://github.com/toeverything/AFFiNE/issues/4029

### DIFF
--- a/apps/electron/src/helper/db/base-db-adapter.ts
+++ b/apps/electron/src/helper/db/base-db-adapter.ts
@@ -1,5 +1,10 @@
-import { type InsertRow, SqliteConnection } from '@affine/native';
+import {
+  type InsertRow,
+  SqliteConnection,
+  ValidationResult,
+} from '@affine/native';
 
+import { migrateToLatestDatabase } from '../db/migration';
 import { logger } from '../logger';
 
 /**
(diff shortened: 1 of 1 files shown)


apps/electron/src/helper/db/base-db-adapter.ts
```diff
@@ -13,6 +18,10 @@ export abstract class BaseSQLiteAdapter {
 
   async connectIfNeeded() {
     if (!this.db) {
+      const validation = await SqliteConnection.validate(this.path);
+      if (validation === ValidationResult.MissingVersionColumn) {
+        await migrateToLatestDatabase(this.path);
+      }
```
pengx17, line 24: there is a _rare_ case that db connection quickly connect then disconnect during this time.
Also I would like to try this change with the corrupted data provided by the users first. brb

himself65:
This is not a good way to fix the issue

PR #4046, opened 2023-08-30
fix: some style issues to sidebar and switch

pengx17:
1. sidebar padding issue
2. sidebar switch issues
3. some other issues


hwangdev97:
There is no problem with the behavior of the sidebar on the web page. 
But I think we can provide sidebar components in Storybook to provide windows and mac style display.

pengx17:
> But I think we can provide sidebar components in Storybook to provide windows and mac style display.

I was also thinking about it, but after looked into the codes and saw that the platform checks is coupled too tightly and not easy to move away from that without some refactoring

hwangdev97 (approved):

PR #4047, opened 2023-08-30
fix: sign in issues

QiShaoXuan:
1. Fix countdown is reset each time switch panels
2. Fix can not set password 
3. Unified email template & modify email content
4. Add access panel 
5. Modify style


apps/core/.webpack/config.ts
```diff
@@ -140,6 +140,14 @@ export const createConfiguration: (
         '.mjs': ['.mjs', '.mts'],
       },
       extensions: ['.js', '.ts', '.tsx'],
+      // fallback: {
```
himself65: why

himself65 (changes requested):

apps/core/.webpack/config.ts
QiShaoXuan: Ignore it, just have try in local

himself65 (approved):

PR #4053, opened 2023-08-30
feat: support google login on desktop

pengx17:
Thanks to the help of @darkskygit, it turns out that we don't need any hacks within Electron by creating a hidden window and handling login inside it. Now, we have a much easier solution:

- Instead of directly calling the signin function when signing in is initiated via Electron, we can pop up the login window in the browser and initiate signIn there.
- Once Google signin is finished, it can redirect to the open-app route, fetch the JWT token, and send it to the desktop app.
- The desktop app can then accept the token and set it to the browser window using cookie.

It seems like we don't need to change the electron origin after all. I will go back and fix email login as well later on.

### DIFF
--- a/apps/core/src/pages/desktop-signin.tsx
+++ b/apps/core/src/pages/desktop-signin.tsx
@@ -1,0 +1,35 @@
+import { signIn, type SignInResponse } from 'next-auth/react';
+import { useEffect, useMemo } from 'react';
+import { useSearchParams } from 'react-router-dom';
+import { z } from 'zod';
+
+const supportedProvider = z.enum(['google']);
+
+let wip = false;
+
+export const Component = () => {
+  const [params] = useSearchParams();
+  const provider = useMemo(() => {
+    const maybeProvider = supportedProvider.safeParse(params.get('provider'));
+    return maybeProvider.success ? maybeProvider.data : null;
+  }, [params]);
+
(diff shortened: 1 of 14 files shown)


apps/core/src/pages/desktop-signin.tsx
```diff
@@ -0,0 +1,35 @@
+import { signIn, type SignInResponse } from 'next-auth/react';
+import { useEffect, useMemo } from 'react';
+import { useSearchParams } from 'react-router-dom';
+import { z } from 'zod';
+
+const supportedProvider = z.enum(['google']);
+
+let wip = false;
```
himself65: use a atom

himself65 (changes requested):

apps/core/src/pages/desktop-signin.tsx
himself65: yep

himself65: I believe we will add e2e test on this. Please make sure the UI is pure

pengx17:
related https://github.com/toeverything/AFFiNE/issues/3944

darkskygit (approved):

darkskygit (approved):

himself65 (approved):

PR #4057, opened 2023-08-31
feat(core): support syncing workspaces and blobs in the background

himself65:


packages/workspace/src/affine/sync.ts
```diff
@@ -0,0 +1,87 @@
...
+  rootGuid: string,
+  doc: Doc,
+  signal: AbortSignal
+): Promise<void> => {
+  if (signal.aborted) {
+    return;
+  }
+  const binary = await downloadBinaryFromCloud(rootGuid, doc.guid);
+  if (typeof binary !== 'boolean') {
+    applyUpdate(doc, new Uint8Array(binary), 'affine-cloud-service');
```
pengx17, line 21: I am a little worried that applying update here will cause performance issues for the foreground tasks. is it possible to move to worker instead?

packages/workspace/src/atom.ts
```diff
@@ -175,6 +176,7 @@ const fetchMetadata: FetchMetadata = async (get, { signal }) => {
             version: WorkspaceVersion.DatabaseV3,
           }))
         );
+        Adapter.Events['service:start']?.();
```
pengx17, line 179: shall we do polling every a few minutes?

packages/workspace/src/affine/sync.ts
himself65: Fixed

packages/workspace/src/atom.ts
himself65: No, socket.io will handle everything when the connection established

PR #4059, opened 2023-08-31
feat: exception logger

thorseraq:


apps/server/src/middleware/exception-logger.ts
```diff
@@ -0,0 +1,35 @@
...
+@Catch(HttpException)
+export class ExceptionLogger implements ExceptionFilter {
+  private logger = new Logger('ExceptionLogger');
+
+  catch(exception: Error, host: ArgumentsHost) {
+    const ctx = host.switchToHttp();
+    const request = ctx.getRequest<Request>();
+    const requestId = request?.header(REQUEST_ID);
+    this.logger.error(
+      `${requestId ? `requestId-${requestId}:` : ''}${exception.message}`,
```
Brooooooklyn, line 21:
```suggestion
      new Error(`${requestId ? `requestId-${requestId}:` : ''}${exception.message}`, { cause: exception })
```

apps/server/src/middleware/exception-logger.ts
```diff
@@ -0,0 +1,35 @@
...
+      exception.stack
```
Brooooooklyn, line 22: remove this line

apps/server/src/modules/index.ts
```diff
@@ -37,4 +40,14 @@ switch (SERVER_FLAVOR) {
     break;
 }
 
-export { BusinessModules };
+const Providers: Provider[] = [];
+
+if (NODE_ENV === 'production') {
+  console.log('push app filter');
```
Brooooooklyn, line 46: remove

apps/server/src/modules/index.ts
```diff
@@ -37,4 +40,13 @@ switch (SERVER_FLAVOR) {
     break;
 }
 
-export { BusinessModules };
+const Providers: Provider[] = [];
+
+if (NODE_ENV === 'production') {
```
forehalo, line 45: why production only?

apps/server/src/middleware/exception-logger.ts
```diff
@@ -0,0 +1,38 @@
...
+  Catch,
+  ExceptionFilter,
+  HttpException,
+  Logger,
+} from '@nestjs/common';
+import { Request, Response } from 'express';
+
+import { REQUEST_ID } from '../constants';
+
+@Catch(HttpException)
```
forehalo, line 12: what about other errors

thorseraq: we don't have rpc request, for ws request I suggest don't catch them, make socketIO handle it

apps/server/src/modules/index.ts
thorseraq: It makes test output ugly

thorseraq: how about NODE_ENV !== 'test' https://github.com/toeverything/AFFiNE/blob/cb0d9fa22aaf5eb130ebc8ba92b72e23b9a76bcd/apps/server/scripts/run-test.ts#L25

Brooooooklyn (approved):

PR #4060, opened 2023-08-31
feat: blob size api

darkskygit:


apps/server/src/modules/workspaces/resolver.ts
```diff
@@ -581,6 +581,16 @@ export class WorkspaceResolver {
     return this.storage.listBlobs(workspaceId);
   }
 
+  @Query(() => Number)
```
Brooooooklyn, line 584: Let's create a `ObjectType` for this:

```ts
WorkspaceBlobSize {
  size: number
}
```

Brooooooklyn (approved):

PR #4071, opened 2023-08-31
refactor(workspace): download full binary in electron

himself65:


packages/workspace/src/affine/sync.ts
```diff
@@ -20,12 +18,11 @@ const downloadRecursive = async (
...
-      // only apply the root doc
-      applyUpdate(doc, update, 'affine-cloud-service');
-    } else {
-      await pushBinary(doc.guid, update);
-    }
+    await window.apis.db.applyDocUpdate(
+      rootGuid,
+      update,
+      rootGuid === doc.guid ? undefined : doc.guid
+    );
```
pengx17, line 25: This time the data is not being written to the ydoc in the view. I am not sure if this is a good decision.

PR #4079, opened 2023-08-31
feat!: unified migration logic in server electron, and browser

himself65:
Both server, web, and electron need to consider data migration

- [x] Electron import/export workspace
  - [ ] test suite (add later because we don't release desktop version now)
- [x] Server e2e data migration test
- [x] Preloading page update
- [x] Core data upgrade e2e test from 0.8.3 stable


himself65 (commented):
@Saul-Mirone Please help me!

Saul-Mirone (approved):

apps/electron/src/helper/dialog/dialog.ts
```diff
@@ -223,6 +224,24 @@ export async function loadDBFile(): Promise<LoadDBFileResult> {
...
+    } catch (error) {
+      logger.warn(
+        `loadDBFile, migration version column failed: ${originalPath}`,
+        error
+      );
+      return { error: 'DB_FILE_MIGRATION_FAILED' };
+    } finally {
+      await db.close();
+    }
+
```
pengx17, line 244: Please move all migration related codes to `../db/migration` instead.
Also, always replace the cloned db's doc rows with the migrated ones.

apps/server/src/modules/sync/events/events.gateway.ts
```diff
@@ -105,6 +108,58 @@ export class EventsGateway implements OnGatewayConnection, OnGatewayDisconnect {
       return false;
     }
 
+    // check migration in the rootDoc
+    if (message.workspaceId === message.guid) {
```
pengx17, line 112: @thorseraq please take a look

thorseraq: LGTM

apps/server/src/modules/sync/events/events.gateway.ts
```diff
@@ -105,6 +108,58 @@ export class EventsGateway implements OnGatewayConnection, OnGatewayDisconnect {
...
+              .map(subdoc =>
+                this.docManager.push(
+                  message.workspaceId,
+                  subdoc.guid,
+                  Buffer.from(encodeStateAsUpdate(subdoc))
+                )
+              )
+          );
+
+          await this.docManager.apply();
```
thorseraq, line 158: not needed, there's an interval, it's periodically called

apps/server/src/modules/sync/events/events.gateway.ts
```diff
@@ -105,6 +108,58 @@ export class EventsGateway implements OnGatewayConnection, OnGatewayDisconnect {
...
+          await Promise.all(
+            subdocs
+              .filter((v): v is YDoc => !!v)
+              // always upgrade the page doc to the latest version
+              .map(subdoc => schema.upgradePage(currentVersions, subdoc))
+          ).then(() =>
+            subdocs
+              .filter((v): v is YDoc => !!v)
+              .map(subdoc =>
+                this.docManager.push(
```
thorseraq, line 150: please only push updates, what's pushed will all be merged with snapshots later

thorseraq:
One thing I think worth noting: If server has updated doc to latest version, and there's an old client, it will download the upgraded doc, the old client may not support rendering that

thorseraq (approved):

thorseraq (approved):

darkskygit (approved):

apps/core/src/components/migration-fallback.tsx
```diff
@@ -0,0 +1,111 @@
...
+    );
+    const indexedDbProvider = providers.find(
+      (provider): provider is LocalIndexedDBBackgroundProvider =>
+        provider.flavour === 'local-indexeddb-background'
+    );
+    const provider = sqliteProvider || indexedDbProvider;
+    assertExists(provider, 'no local provider');
+    return provider;
+  }, [providers]);
+  const handleClick = useCallback(async () => {
```
pengx17, line 37: may disable the button when migration is in progress

Issue #4056, opened 2023-08-31
Mode buttons and title disappear while renaming untitled doc


xell:
After adding a new note with empty title and then clicking the title bar to rename it, the mode buttons and title disappear and there's no cursor:

https://github.com/toeverything/AFFiNE/assets/132558/cdebd278-d67d-4263-b1fd-fe9564968c96

houjoe0829:
Similar with https://github.com/toeverything/AFFiNE/issues/3777#event-10125636150 ？

ideakaran:
When we click on the page title (to rename it), the user can still type the title.
However, a border around the title seemed missing which is now added via this pull req https://github.com/toeverything/AFFiNE/pull/4080/files

Change demo: https://www.veed.io/view/f6080bf8-2ac6-43e3-a980-36c7b0d7aebc

PR #4080, opened 2023-08-31
fix: show border around pagetitle when renaming

ideakaran:
fix: #4056 

Showed a border around the page title only when title input is on focus.


himself65 (approved):
what a pretty fast fix. Thank u!

PR #4115, opened 2023-09-01
fix: cookie issues in Electron

pengx17:
need to handling cookies more gracefully
- if getting back cookies from api, we will mark it cross-domain available
- if sending request, add the cookies to headers

also fix some issues related to callbackUrl. If not being explicitly set in electron, it will be something like `file://./xxxx` and the api server does not like it and throw.

### DIFF
--- a/apps/server/src/modules/auth/next-auth-options.ts
+++ b/apps/server/src/modules/auth/next-auth-options.ts
@@ -98,7 +98,7 @@ export const NextAuthOptionsProvider: FactoryProvider<NextAuthOptions> = {
       adapter: prismaAdapter,
       debug: !config.node.prod,
       session: {
-        strategy: config.node.prod ? 'database' : 'jwt',
+        strategy: 'jwt',
       },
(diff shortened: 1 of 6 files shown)


apps/server/src/modules/auth/next-auth-options.ts
```diff
@@ -98,7 +98,7 @@ export const NextAuthOptionsProvider: FactoryProvider<NextAuthOptions> = {
       adapter: prismaAdapter,
       debug: !config.node.prod,
       session: {
-        strategy: config.node.prod ? 'database' : 'jwt',
+        strategy: 'jwt',
```
pengx17, line 101: Also uses `jwt` session token on production.

himself65 (approved):

apps/server/src/modules/auth/next-auth-options.ts
Brooooooklyn: It seems safe; only increase the cookie size, which is acceptable

PR #4117, opened 2023-09-01
feat: add user info edit verify

darkskygit:


Brooooooklyn (approved):

Brooooooklyn (commented):
Why not using existed `ioredis`?

himself65 (approved):

darkskygit:
> Why not using existed `ioredis`?

keyv/redis use ioredis internally:https://github.com/toeverything/AFFiNE/blob/af8d320a6eb8e31460f5d9260b505162b83a083a/apps/server/src/session.ts#L17-L23

keyv can provide a memory kv storage fallback when we dont run with redis(e.g. self hosted image)

Issue #4105, opened 2023-09-01
broken server tests because of race conditions


himself65:
Upstream: https://github.com/nodejs/node/issues/48871#issuecomment-1646680159

PR #4120, opened 2023-09-01
refactor(server): use ava

himself65:
We no longer use node.js test runner since it's too weak engineering for a product targeting the production

Fixes: https://github.com/toeverything/AFFiNE/issues/4105

Upstream: https://github.com/nodejs/node/issues/48871


apps/server/src/index.ts
```diff
@@ -75,6 +76,7 @@ app.use(
   })
 );
 
+app.useGlobalFilters(new ExceptionLogger());
```
himself65, line 79: ExceptionLogger will throw an error in the previous code. I change to another style and it works

@thorseraq

thorseraq (approved):

PR #4131, opened 2023-09-02
ci: enable server coverage

himself65:


apps/server/src/tests/mailer.e2e.ts
```diff
@@ -0,0 +1,105 @@
...
+      GqlModule,
+      AuthModule,
+      MetricsModule,
+      RateLimiterModule,
+    ],
+  }).compile();
+  auth = module.get(AuthService);
+});
+
+test.afterEach(async () => {
```
pengx17, line 67: please use `afterEach.awalys`
https://github.com/avajs/ava/blob/main/docs/01-writing-tests.md#:~:text=Use%20test.afterEach.always()%20to%20register%20an%20after%20hook%20that%20is%20called%20even%20if%20other%20test%20hooks%2C%20or%20the%20test%20itself%2C%20fail

PR #4160, opened 2023-09-04
feat: add shortcut keys for toggling setting modal

TinsFox:
Add shortcut keys for toggling setting modal

`⌘` + `K` or  `Ctrl` + `K`


QiShaoXuan (dismissed):

apps/core/src/components/affine/setting-modal/index.tsx
```diff
@@ -65,6 +65,19 @@ export const SettingModal = ({
     onSettingClick({ activeTab: 'account', workspaceId: null });
   }, [onSettingClick]);
 
+  // Add ‘⌘+,’ shortcut keys as switches
+  useEffect(() => {
+    const keydown = (e: KeyboardEvent) => {
+      if ((e.key === ',' && e.metaKey) || (e.key === ',' && e.ctrlKey)) {
+        e.preventDefault();
+        setOpen(!open);
```
himself65: setOpen(open => !open)

himself65 (changes requested):

apps/core/src/components/affine/setting-modal/index.tsx
TinsFox: There is an error in this way of writing

image

It may need to be modified synchronously here 👇

https://github.com/toeverything/AFFiNE/blob/a2e4ef904bf281d3c2a3969b1b7bf4b78a80a12b/apps/core/src/providers/modal-provider.tsx#L99-L104

Modify here too, if necessary

https://github.com/toeverything/AFFiNE/blob/3b6e145b234614d06c0b8dba3676944753d4ef1b/apps/core/src/components/pure/quick-search-modal/index.tsx#L57

himself65: Yes, the upstream code should be updated same

TinsFox: In fact, useEffect already depends on open. It will trigger an update after setOpen. So is it necessary to use the functional update approach to get the latest value?

himself65: > In fact, useEffect already depends on open. It will trigger an update after setOpen. So is it necessary to use the functional update approach to get the latest value?

This is not supposed to be in react.js. `[open, setOpen]` will trigger this effect every time the value flips, this will cause extra unused re-hooks here

himself65 (approved):

PR #4183, opened 2023-09-05
fix(core): remove list when fetching local

himself65:


pengx17:
ops, I did not know you have opened one

PR #4186, opened 2023-09-05
feat: unexpected request type check for exception middleware

thorseraq:


apps/server/src/middleware/exception-logger.ts
```diff
@@ -16,6 +16,13 @@ export class ExceptionLogger implements ExceptionFilter {
   catch(exception: Error, host: ArgumentsHost) {
     // with useGlobalFilters, the context is always HTTP
     const ctx = host.switchToHttp();
+
+    // TODO: temporary, remove this after stable
+    if (host.getType() !== 'http') {
+      this.logger.error(`Unexpected request type caught: ${host.getType()}`);
```
Brooooooklyn, line 22: provide more information here, maybe request path or something that helpful to debug

darkskygit (approved):

Brooooooklyn (commented):
I have a question, this ExceptionLogger is used to log exceptions, but why does it seem to change the original response, for example I see a lot of response.json calls, why is it handled this way instead of returning the response as is?

thorseraq:
That's the best way I think, but I also tried to return response directly, the http request will not get response (If exception was caught by it)

thorseraq:
throw an exception -> app crash
return an exception -> crash
return response -> no response
reponse.json() -> works

apps/server/src/middleware/exception-logger.ts
```diff
@@ -4,31 +4,60 @@ import {
...
+    const ifVerboseLog = !TrivialExceptions.some(e => exception instanceof e);
     this.logger.error(
       new Error(
         `${requestId ? `requestId-${requestId}:` : ''}${exception.message}`,
         { cause: exception }
-      ),
-      exception.stack
+      )
     );
+    ifVerboseLog && this.logger.error(exception.stack);
```
Brooooooklyn, line 49:
```suggestion
    if (ifVerboseLog) {
      this.logger.error(exception.stack);
    }
```

thorseraq:
closed by https://github.com/toeverything/AFFiNE/pull/4209

PR #4198, opened 2023-09-05
feat: server refactor

darkskygit:
### gold

- [x] relax throttle for low risk apis
- [ ] review safety issues in api
- [x] blob limit in frontend
- [x] reduce blob size apis to support multiple user level quota

### DIFF
--- a/apps/server/src/modules/users/users.ts
+++ b/apps/server/src/modules/users/users.ts
@@ -1,0 +1,50 @@
+import { Injectable } from '@nestjs/common';
+
+import { Config } from '../../config';
+import { PrismaService } from '../../prisma';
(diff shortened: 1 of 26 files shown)


apps/server/src/modules/users/users.ts
```diff
@@ -0,0 +1,68 @@
...
+    return getStorageQuota(features) || this.config.objectStorage.quota;
+  }
+
+  async findUserByEmail(email: string) {
+    return this.prisma.user
+      .findUnique({
+        where: { email },
+      })
+      .catch(() => {
+        return null;
```
Brooooooklyn, line 51: `logger.error` here?

Brooooooklyn (approved):

Issue #4217, opened 2023-09-06
After successfully removing a member, add a toast prompt


houjoe0829:
Toast：Removed successfully

PR #4231, opened 2023-09-06
feat: support pagination for member list

QiShaoXuan:
image

Fixed: https://github.com/toeverything/AFFiNE/issues/4217 (by the way)


packages/component/src/components/member-components/invite-modal.tsx
```diff
@@ -15,49 +15,45 @@ export interface InviteModalProps {
...
-
-  return null;
-  // return (
-  //   <Menu
-  //     trigger="click"
-  //     content={
-  //       <>
-  //         {Object.entries(Permission).map(([permission]) => {
-  //           return (
-  //             <MenuItem
```
Brooooooklyn, line 36: remove it

apps/server/src/modules/workspaces/resolver.ts
```diff
@@ -212,15 +211,25 @@ export class WorkspaceResolver {
...
+    @Parent() workspace: WorkspaceType,
+    @Args('skip', { type: () => Int, nullable: true }) skip?: number,
+    @Args('take', { type: () => Int, nullable: true }) take?: number
+  ) {
     const data = await this.prisma.userWorkspacePermission.findMany({
       where: {
         workspaceId: workspace.id,
       },
+      skip,
+      take: take || 8,
```
Brooooooklyn, line 224: why the default value of take is 10 in frontend and 8 in the backend?

packages/component/src/components/member-components/invite-modal.tsx
```diff
@@ -15,49 +15,45 @@ export interface InviteModalProps {
...
+//         style={{
+//           marginRight: -10,
+//           height: '100%',
+//         }}
+//       >
+//         {currentPermission}
+//       </MenuTrigger>
+//     </Menu>
+//   );
+// };
```
Brooooooklyn, line 57: remove it

himself65 (changes requested):
It would be better if you added a test case

Like, register 10 users and add them into a workspace, open the workspace list, and the pagination should work

QiShaoXuan:
> It would be better if you added a test case
> 
> Like, register 10 users and add them into a workspace, open the workspace list, and the pagination should work

Added in e2e test

himself65 (approved):

Issue #4252, opened 2023-09-07
In the registration and login pop-up window, the back button should be displayed as "Back"


houjoe0829:
Currently displayed is "Back home"

![CleanShot 2023-09-07 at 14 10 00@2x](https://github.com/toeverything/AFFiNE/assets/22443345/75eb3fbd-dc0a-4989-a4cc-def9277d4e87)

PR #4257, opened 2023-09-07
fix: modify back text

QiShaoXuan:
fixed: https://github.com/toeverything/AFFiNE/issues/4252

### DIFF
--- a/packages/i18n/src/resources/en.json
+++ b/packages/i18n/src/resources/en.json
@@ -133,7 +133,7 @@
   "com.affine.auth.toast.message.signed-in": "You have been signed in, start to sync your data with AFFiNE Cloud!",
   "com.affine.auth.toast.title.failed": "Unable to sign in",
   "com.affine.auth.toast.title.signed-in": "Signed in",
-  "com.affine.backButton": "Back Home",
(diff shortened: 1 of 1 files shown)


packages/i18n/src/resources/en.json
```diff
@@ -133,7 +133,7 @@
   "com.affine.auth.toast.message.signed-in": "You have been signed in, start to sync your data with AFFiNE Cloud!",
   "com.affine.auth.toast.title.failed": "Unable to sign in",
   "com.affine.auth.toast.title.signed-in": "Signed in",
-  "com.affine.backButton": "Back Home",
+  "com.affine.backButton": "Back",
```
Garfield550, line 136: `com.affine.backButton` also effect `404 Not Found` page, please to consider use a new key or give `Back Home` button inside `404` page a new key.

Garfield550 (changes requested):

packages/i18n/src/resources/en.json
Garfield550: I will add a new key called `com.affine.notFoundPage.backButton`. You don't need to take any action.

Garfield550 (approved):

Issue #4261, opened 2023-09-07
Clicking "Sign in and Enable" without logging in to the account does not have any response


houjoe0829:
![CleanShot 2023-09-07 at 16 11 41@2x](https://github.com/toeverything/AFFiNE/assets/22443345/9e6938c2-3c0e-4c78-b5fc-bb7e3b41fc1e)

Expected effect：
1. Clicking "Sign in and Enable" will open the login window.
2. After logging in or creating an account, the current Workspace will automatically enable AFFiNE Cloud.

Note that after creating an account, a pop-up window should not appear asking the user to select which Workspace to enable AFFiNE Cloud for. Instead, the current Workspace should automatically enable AFFiNE Cloud.

Issue #4208, opened 2023-09-06
After enabling AFFiNE Cloud, the Toast notification is missing


houjoe0829:
Once AFFiNE Cloud is enabled, a Toast notification should pop up saying "Successfully enabled AFFiNE Cloud".

Issue #4207, opened 2023-09-06
When the account is already logged in, it should not be displayed as "Sign in and Enable"


houjoe0829:
<img width="518" alt="CleanShot 2023-09-06 at 11 12 15@2x" src="https://github.com/toeverything/AFFiNE/assets/22443345/500e90f2-754b-4452-8bce-57f31a9c80be">

It should simply display "Enable".

PR #4265, opened 2023-09-07
fix: can not enable workspace if not sign in

QiShaoXuan:
fixed: https://github.com/toeverything/AFFiNE/issues/4207
fixed: https://github.com/toeverything/AFFiNE/issues/4261
fixed: https://github.com/toeverything/AFFiNE/issues/4208


apps/core/src/components/affine/auth/index.tsx
QiShaoXuan, line 1: @himself65 I try to call onTransformWorkspace to transform to cloud after signed in, but after signed in ,code throw error, I think it is a timing issue
image

himself65: Because you click the button twice

PR #4270, opened 2023-09-07
fix: sigin in different window may not refresh workspace list

pengx17:


apps/core/src/providers/session-provider.tsx
```diff
@@ -0,0 +1,52 @@
...
+      });
+      pushNotification({
+        title: `${email}${t['com.affine.auth.has.signed']()}`,
+        message: '',
+        key: Date.now().toString(),
+        type: 'success',
+      });
+
+      if (isDesktop) {
+        window.affine.ipcRenderer.send('affine:login');
```
himself65, line 35: good to see we are starting use this api

himself65 (approved):

PR #4274, opened 2023-09-08
fix: SHOULD_REPORT_TRACE webpack plugin value

thorseraq:


apps/core/.webpack/config.ts
```diff
@@ -352,9 +352,7 @@ export const createConfiguration: (
         'process.env': JSON.stringify({}),
         'process.env.COVERAGE': JSON.stringify(!!buildFlags.coverage),
         'process.env.NODE_ENV': JSON.stringify(buildFlags.mode),
-        'process.env.SHOULD_REPORT_TRACE': `${Boolean(
-          process.env.SHOULD_REPORT_TRACE
-        )}`,
+        'process.env.SHOULD_REPORT_TRACE': `"${process.env.SHOULD_REPORT_TRACE}"`,
         'process.env.TRACE_REPORT_ENDPOINT': `"${process.env.TRACE_REPORT_ENDPOINT}"`,
```
pengx17, line 356: I think you can use JSON.stringify to make it consistent with the above and it is in fact more robust than using template string:

```
        'process.env.SHOULD_REPORT_TRACE': JSON.stringify(process.env.SHOULD_REPORT_TRACE),
        'process.env.TRACE_REPORT_ENDPOINT': JSON.stringify(process.env.TRACE_REPORT_ENDPOINT),
```

thorseraq: I found it seems wield, `JSON.stringfy('true')`will get `'"true"'`, which is not equal to `'true'; https://github.com/toeverything/AFFiNE/blob/c663e94685bd2b3a0bea3e5cfa4f3042641a36d9/packages/graphql/src/utils.ts#L208

The current impl is also wrong, please help checking this
```
`"${process.env.SHOULD_REPORT_TRACE}"`
```

thorseraq: I compiled at my local machine, very sure it's correct now; [ref](https://webpack.js.org/plugins/define-plugin/#:~:text=Note%20that%20because%20the%20plugin%20does%20a%20direct%20text%20replacement%2C%20the%20value%20given%20to%20it%20must%20include%20actual%20quotes%20inside%20of%20the%20string%20itself.%20Typically%2C%20this%20is%20done%20either%20with%20alternate%20quotes%2C%20such%20as%20%27%22production%22%27%2C%20or%20by%20using%20JSON.stringify(%27production%27).) [ref2](https://webpack.js.org/plugins/define-plugin/#:~:text=After%20passing%20through%20webpack%20with%20no%20minification%20results%20in%3A)

pengx17: I think we should always assume values from `process.env` are strings and never do interpolation. 
That's why I think it may is better to use `JSON.stringify('true')`.

thorseraq: There are some bugs, I can't see my comments...

I compiled affine-core in my local machine, I'm pretty sure it's correct now, here're the refs:
[ref1](https://webpack.js.org/plugins/define-plugin/#:~:text=If%20the%20value%20isn%27t%20a%20string%2C%20it%20will%20be%20stringified%20(including%20functions).), [ref2](https://webpack.js.org/plugins/define-plugin/#:~:text=Note%20that%20because%20the%20plugin%20does%20a%20direct%20text%20replacement%2C%20the%20value%20given%20to%20it%20must%20include%20actual%20quotes%20inside%20of%20the%20string%20itself.%20Typically%2C%20this%20is%20done%20either%20with%20alternate%20quotes%2C%20such%20as%20%27%22production%22%27%2C%20or%20by%20using%20JSON.stringify(%27production%27).), [ref3](https://webpack.js.org/plugins/define-plugin/#:~:text=After%20passing%20through%20webpack%20with%20no%20minification%20results%20in%3A)

pengx17: Could you please also revise the above? Another issue is that we shall make the codes consistent, especially if they are colocated

PR #4283, opened 2023-09-08
fix: edge case when upgrading page

thorseraq:


himself65 (approved):
We need to find our why this happen in the future

PR #4285, opened 2023-09-08
fix(core): temporary remove blockVersions assertion

Brooooooklyn:


himself65 (approved):
We need to move the passive provider into loader in the future

PR #4288, opened 2023-09-08
fix: allow login with credentials on production

pengx17:


Brooooooklyn (approved):

apps/server/src/modules/auth/next-auth.controller.ts
```diff
@@ -140,6 +147,29 @@ export class NextAuthController {
...
     }
 
+    let nextAuthTokenCookie: (CookieOption & { value: string }) | undefined;
+    // next-auth credentials login only support JWT strategy
+    // https://next-auth.js.org/configuration/providers/credentials
+    // let's store the session token in the database
+    if (
+      credentialsSignIn &&
+      (nextAuthTokenCookie = cookies?.find(
+        ({ name }) => name === 'next-auth.session-token'
```
pengx17, line 157: Need to add Secure_ prefix when https is enabled

Brooooooklyn: Fix it for me

PR #4295, opened 2023-09-09
fix: get auth token for development

pengx17:


apps/server/src/modules/auth/resolver.ts
```diff
@@ -49,18 +52,24 @@ export class AuthResolver {
...
     }
 
-    // on production we use session token that is stored in database (strategy = 'database')
-    const sessionToken = this.config.node.prod
-      ? await this.auth.getSessionToken(user.id)
-      : this.auth.sign(user);
+    const cookiePrefix = this.config.node.prod ? '__Secure-' : '';
+    const sessionCookieName = `${cookiePrefix}next-auth.session-token`;
+
+    const sessionToken: string | undefined =
```
darkskygit, line 67: add a check to get the session token only when referer=/open-app and host=config.origin

darkskygit (approved):

Issue #4256, opened 2023-09-07
The process for changing the email address lacks a verification step for the new email


houjoe0829:
In the current case, as long as the new email address is entered, the change is completed directly.

PR #4306, opened 2023-09-11
feat: add verify process in change email progress

QiShaoXuan:
close https://github.com/toeverything/AFFiNE/issues/4256


apps/server/src/modules/auth/service.ts
```diff
@@ -251,4 +251,23 @@ export class AuthService {
...
+        userId: userId,
+      },
+    });
+
+    if (!session) {
+      throw new BadRequestException(`No session found for user id ${userId}`);
+    }
+
+    return session?.sessionToken;
+  }
```
pengx17, line 272: this has been removed in master

apps/core/src/pages/auth.tsx
```diff
@@ -128,7 +173,13 @@ export const Component = () => {
   }
 
   if (loginStatus === 'authenticated') {
-    return <AuthPage />;
+    return (
+      <>
+        <AuthPage />
+        {runtimeConfig.enableNotificationCenter && <NotificationCenter />}
```
pengx17, line 179: it seems to me this should be put at the app root and remove the one in workspace layout, regardless of user's login status.

pengx17 (approved):
LGTM with some minor issues

PR #4309, opened 2023-09-12
fix: incomplete URL substring sanitization

fourdim:


apps/server/src/modules/auth/next-auth.controller.ts
```diff
@@ -179,7 +179,7 @@ export class NextAuthController {
       });
       if (
         !req.headers?.referer ||
-        req.headers.referer.startsWith('https://accounts.google.com')
+        new URL(req.headers.referer).host === 'https://accounts.google.com'
```
pengx17, line 182: `new URL('https://accounts.google.com').host` === 'accounts.google.com'.
```suggestion
        new URL(req.headers.referer).origin === 'https://accounts.google.com'
```

apps/server/src/modules/auth/next-auth.controller.ts
```diff
@@ -179,7 +179,7 @@ export class NextAuthController {
       });
       if (
         !req.headers?.referer ||
-        req.headers.referer.startsWith('https://accounts.google.com')
+        new URL(req.headers.referer).origin === 'https://accounts.google.com'
```
lawvs, line 182: The `new URL()` will throw an exception when the URL is invalid. Consider wrapping it in a try-catch block.

pengx17: Is it really possible that referer is invalid in the realworld?

apps/server/src/modules/auth/next-auth.controller.ts
```diff
@@ -177,10 +177,17 @@ export class NextAuthController {
       this.metrics.authFailCounter(1, {
         reason: 'no_early_access_permission',
       });
-      if (
-        !req.headers?.referer ||
-        req.headers.referer.startsWith('https://accounts.google.com')
-      ) {
+      const checkReferer = () => {
```
Garfield550, line 180: Maybe you can define `checkReferer` top of the file or inside the `NextAuthController` class as a private method?

```
const checkReferer = (referer?: string) => {
    if (!referer) return false
    // other logic
}
```

Garfield550 (changes requested):

apps/server/src/modules/auth/next-auth.controller.ts
lawvs: The "referer" is an HTTP header sent from the client which means it could be any value.

apps/server/src/modules/auth/next-auth.controller.ts
fourdim: I have considered that before.
The thing is this `checkReferer` is not general enough (or robust enough) to be moved to the class level nor module level. I don't want other people to misuse this function.
As we already have
```js
if (!req.headers?.referer || checkReferer()) {
```
it checks the referer whether it's undefined. If undefined, `checkReferer` will be short circuit and not be executed.
Hence, I put
```
new URL(req.headers.referer ?? '')
```
here is just to make compiler happy.

As the function itself depends on the context, I would prefer not moving it outside.

The function you have written is reansonable and intuitive to be put outside.

However, if I would like to extract the original method outside will be

```js
const checkReferer = (referer?: string) => {
 -   if (!referer) return false
 +   if (!referer) return true // to replace !req.headers?.referer
    // other logic
}
```

That will make `checkReferer` function weird. Would you still insist on this?

Garfield550: > As the function itself depends on the context, I would prefer not moving it outside.

Okay, considering that the "early access" may not last long and currently only Google is supported, it is acceptable to place `checkReferer` inside the function.

> That will make checkReferer function weird.

If `checkReferer` will reverse the logic, consider reversing `if {} else {}` as well and add some brief comments? or simply consider a more appropriate name, perhaps starting with `should`?

fourdim: I come up with a better solution.
Since we have
```js
req.headers.referer.startwith('https://accounts.google.com')
```
before.

We can just implement a function called `checkRefererOrigin` to replace this in place.

That is
```js
checkRefererOrigin(req.headers.referer, 'https://accounts.google.com')
```

Unfortunately, I have my computer broken recently and not able to modify it. Would you like to open a new pull request to fix this?
Once you open that, I will close this pull request. Thanks a lot.

Garfield550: > Would you like to open a new pull request to fix this?

Sorry for the late reply, but I can't help you. I have my own things to deal with.

Garfield550 (approved):

Brooooooklyn (approved):

PR #3838, opened 2023-08-19
refactor: change locale key

Garfield550:
Pattern:

`com.affine.<componentName>.<region>.[property]`

packageName Level:

`packages/* > apps/* > plugins/*`

08/31/2023 update:

I have decided not to include the `packageName` in the language locale key.

```diff
- com.affine.<packageName>.<componentName>.<region>.[property]
+ com.affine.<componentName>.<region>.[property]
```

PR #4314, opened 2023-09-12
refactor: change locale key

Garfield550:
part of #3838


Garfield550:
As of the deadline on 20/09/2023, I need to help a friend handle an urgent matter, so the follow-up on this PR will be postponed.

PR #4318, opened 2023-09-12
fix(core): back home botton has no reaction 

### LINKED ISSUE #4201: There is no reaction upon clicking the Back Home button on the 404 Page.
image

@xell

joooye34:
close #4201 

The back to last workspace logic in 404 page just been removed, because index page will do this more safely.

There have some others logic should be improved but not this commit:
1. no 400 exception page, all in 404
2. 404 should have search params to know what resource not found and what page from

### DIFF
--- a/apps/core/src/pages/404.tsx
+++ b/apps/core/src/pages/404.tsx
@@ -1,5 +1,4 @@
 import { displayFlex, styled } from '@affine/component';
-import { WorkspaceSubPath } from '@affine/env/workspace';
 import { useAFFiNEI18N } from '@affine/i18n/hooks';
 import { Button } from '@toeverything/components/button';
 import type { ReactElement } from 'react';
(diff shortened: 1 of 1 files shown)


pengx17 (approved):

apps/core/src/pages/404.tsx
```diff
@@ -26,22 +25,14 @@ export const StyledContainer = styled('div')(() => {
...
-        onClick={() => {
-          const id = localStorage.getItem('last_workspace_id');
-          if (id) {
-            jumpToSubPath(id, WorkspaceSubPath.ALL);
-          } else {
-            jumpToIndex();
-          }
-        }}
-      >
+      <Button onClick={() => jumpToIndex()}>
```
himself65: useCallback

himself65 (changes requested):

himself65 (approved):

apps/core/src/pages/404.tsx
joooye34: I have seen lint error in CI, but why my local lint is successful.

joooye34: oh, u commit a patch to improve the code

PR #4327, opened 2023-09-12
refactor(core): session provider

himself65:


packages/auth/src/react.tsx
```diff
@@ -0,0 +1,533 @@
...
+        }
+      | {
+          data: null;
+          status: 'unauthenticated' | 'loading';
+        };
+
+const sessionLoadingAtom = atom<boolean>(false);
+const sessionValueAtom = atom<CloudSession | null>(null);
+
+export const sessionAtom = atom<
```
pengx17, line 188: I suggest do not use jotai in this package so that it will be easier to be commited to upstream.

Issue #4338, opened 2023-09-13
​subdoc guid should always have root doc guid as its prefix


forehalo:
### What happened?

![image](https://github.com/toeverything/AFFiNE/assets/8281226/eb399cfc-7c9a-4a78-8723-df5930719cbb)

### Distribution version

macOS x64 (Intel)

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

PR #4341, opened 2023-09-13
fix: make server guid consistent

### LINKED ISSUE #4132: inconstant guid in server side
There are different formats in the database 

- `${workspaceId}` (uuid)
- `${workspaceId}:space:${pageId}` (uuid:space:nanoid)
- `space:${pageId}` (space:nanoid)

The reason is server will use `uuidv4` as the id generator, but in the frontend will use `nanoid` with `${workspaceId}:space:` as the prefix.

This might cause the SQL table's primary key inconsistent, and hard to lookup the document because there are three types of id for some reason.

What's more, there already some issues in the existing codespace.

Example:
https://github.com/toeverything/AFFiNE/blob/4f97ea8a5d7cda629e6d17345d74a017b18bfcf5/apps/server/src/modules/workspaces/permission.ts#L52-L55

We need to cralify the guid with same format in both frontend and backend

forehalo:
close #4132
walkaround for #4338

### DIFF
--- a/apps/server/src/modules/doc/manager.ts
+++ b/apps/server/src/modules/doc/manager.ts
@@ -2,6 +2,7 @@ import {
   Inject,
   Injectable,
   Logger,
+  OnApplicationBootstrap,
   OnModuleDestroy,
   OnModuleInit,
 } from '@nestjs/common';
@@ -13,6 +14,7 @@ import { Config } from '../../config';
 import { Metrics } from '../../metrics/metrics';
 import { PrismaService } from '../../prisma';
 import { mergeUpdatesInApplyWay as jwstMergeUpdates } from '../../storage';
+import { DocID } from '../../utils/doc';
 
 function compare(yBinary: Buffer, jwstBinary: Buffer, strict = false): boolean {
   if (yBinary.equals(jwstBinary)) {
@@ -42,7 +44,9 @@ const MAX_SEQ_NUM = 0x3fffffff; // u31
  * along side all the updates that have not been applies to that snapshot(timestamp).
  */
 @Injectable()
-export class DocManager implements OnModuleInit, OnModuleDestroy {
+export class DocManager
+  implements OnModuleInit, OnModuleDestroy, OnApplicationBootstrap
+{
   protected logger = new Logger(DocManager.name);
   private job: NodeJS.Timeout | null = null;
   private seqMap = new Map<string, number>();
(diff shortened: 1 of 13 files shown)


apps/server/src/modules/doc/manager.ts
```diff
@@ -357,4 +367,55 @@ export class DocManager implements OnModuleInit, OnModuleDestroy {
...
+              id_workspaceId: {
+                id: doc.id,
+                workspaceId: doc.workspaceId,
+              },
+            },
+            data: {
+              id: parsed.id,
+            },
+          });
+        }
```
darkskygit, line 415: put in a transaction

himself65 (commented):
I think we should merge this all together

https://github.com/toeverything/blocksuite/pull/4747

apps/server/src/modules/doc/manager.ts
forehalo: The delete statement is meaning to delete the unexpected data with wrong guid, like we fix `${ws}:space:${id}` to `space:${id}`, but the `space:${id}` should never exist before in the database, so in this way we could avoid primary key conflict.

forehalo: I checked my local database with some test workspace, somehow both `${ws}:space:${id}` and `space:${id}` existed, which is impossible in our codebase because guid itself is consistent once the doc is ever created.

apps/core/src/hooks/affine/use-is-shared-page.ts
```diff
@@ -31,18 +31,14 @@ export function useIsSharedPage(
       (enable: boolean) => {
         // todo: push notification
         if (enable) {
-          enableSharePage({
-            workspaceId,
-            pageId,
-          })
+          enableSharePage({ docId: `${workspaceId}:space:${pageId}` })
```
forehalo, line 34: https://github.com/toeverything/blocksuite/pull/4747 carefully bump blocksuite the next time @himself65

himself65 (commented):
blocksuite side already merges the id generator breaking change. I wonder if we should also do some compatible in this PR

forehalo:
> blocksuite side already merges the id generator breaking change. I wonder if we should also do some compatible in this PR

done

Brooooooklyn (approved):

darkskygit (approved):

Brooooooklyn (approved):

PR #4349, opened 2023-09-13
fix: test hang

pengx17:
After https://github.com/toeverything/AFFiNE/pull/4324 is merged, the server test keeps failing on CI.
Yet this is not reproducible on local machine, after several hours of investigation on trials and erros, I noticed in some playwright trace files that it shows me this error:

image

It seems radix-ui‘s react-focus-scope and mui's focus trap are trying to retain the focus state respectively and it falls into an infinite loop. This is pretty strange since we are no longer using mui Modal after #4324.
At last I found thatmui's popover is also a wrapper around modal, which could be the root cause.

You can see in the below image the popover (workspace list) and modal (setting) are opening at the same time. Each of them is trying to take over the focus control, but neither one succeeded at the end.
image

## How to fix

Before we rewrite Popover with radix ui, add `disableEnforceFocus` for mui popover


himself65 (approved):

QiShaoXuan:
Let me replace the mui-popover to radix-popover

QiShaoXuan:
I have try to fix it in there: https://github.com/toeverything/AFFiNE/blob/09510479e27d21b13cc5e7f1916ce334e7607bbe/apps/core/src/components/pure/workspace-list-modal/index.tsx#L213
But it looks like it only works locally, I will replace mui-popover to radix-popover later

PR #4369, opened 2023-09-15
fix: remove useRef in menu & tooltip

### LINKED ISSUE #4362: The language switching menu does not pop up correctly
https://github.com/toeverything/AFFiNE/assets/102217452/672350fb-a79f-464a-b404-ecb261e0a950

### LINKED ISSUE #4381: Can't select the language box when the font is in sans
Image

QiShaoXuan:
fixed: https://github.com/toeverything/AFFiNE/issues/4362
fixed: https://github.com/toeverything/AFFiNE/issues/4381
fixed export, collection menu can not trigger

In the past, ref was used because some components did not have specifications, which caused problems with zIndex. Now new components have been migrated, and ref also brings bugs

### DIFF
--- a/apps/core/package.json
+++ b/apps/core/package.json
@@ -39,7 +39,7 @@
     "@mui/material": "^5.14.7",
     "@radix-ui/react-select": "^1.2.2",
     "@react-hookz/web": "^23.1.0",
-    "@toeverything/components": "^0.0.41",
+    "@toeverything/components": "^0.0.42",
     "async-call-rpc": "^6.3.1",
     "cmdk": "^0.2.0",
     "css-spring": "^4.1.0",
--- a/apps/core/src/components/affine/create-workspace-modal/index.tsx
+++ b/apps/core/src/components/affine/create-workspace-modal/index.tsx
@@ -11,7 +11,7 @@ import type {
 } from '@toeverything/infra/type';
 import { useSetAtom } from 'jotai';
 import type { KeyboardEvent } from 'react';
-import { useEffect, useRef } from 'react';
+import { useEffect } from 'react';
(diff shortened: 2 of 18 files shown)


Brooooooklyn (changes requested):
Can you add some e2e tests to prevent these cases in the future?

QiShaoXuan:
> Can you add some e2e tests to prevent these cases in the future?

It has really added https://github.com/toeverything/AFFiNE/blob/1aec1ce7d0bc21e8443c1fc2837b540bed80dc41/tests/affine-local/e2e/settings.spec.ts#L23
This bug only appears online, but not in local and test environment

QiShaoXuan:
@himself65 I think UI Review error is since merged the error snapshot last time

Brooooooklyn:
> This bug only appears online, but not in local and test environment

Why?

himself65 (changes requested):
add e2e test

himself65:
> > This bug only appears online, but not in local and test environment
> 
> Why?

I think maybe electron capture some event or something. need to also add test on that side

QiShaoXuan:
> > > This bug only appears online, but not in local and test environment
> > 
> > 
> > Why?
> 
> I think maybe electron capture some event or something. need to also add test on that side

Not electron, it appear on the web page

QiShaoXuan:
> > > > This bug only appears online, but not in local and test environment
> > > 
> > > 
> > > Why?
> > 
> > 
> > I think maybe electron capture some event or something. need to also add test on that side
> 
> Not electron, it appear on the web page

I think the cause of this bug is due to the container that set menu through ref, ref did not update the menu component, but it is strange that this issue only occurs online(web and electron)

PR #4375, opened 2023-09-15
fix: orderby in members list

pengx17:


apps/server/src/modules/workspaces/resolver.ts
```diff
@@ -220,6 +220,7 @@ export class WorkspaceResolver {
       take: take || 8,
       orderBy: {
         createdAt: 'desc',
```
Brooooooklyn, line 222: my fault, this should be `'asc'`

Brooooooklyn, line 222:
```suggestion
        createdAt: 'asc',
```

Brooooooklyn (dismissed):

himself65 (changes requested):
I think it's quite easy to add a server unit test

himself65 (approved):

PR #4405, opened 2023-09-18
fix: read permission for subpage

### LINKED ISSUE #4335: When logged into an AFFiNE account on the web version, accessing a public link should not display 404 page
image

The expected result is that public links can be accessed normally regardless of whether an AFFiNE account is logged in or not.

darkskygit:
fix #4335

### DIFF
--- a/apps/server/src/modules/workspaces/permission.ts
+++ b/apps/server/src/modules/workspaces/permission.ts
@@ -47,37 +47,38 @@ export class PermissionService {
 
   async isAccessible(ws: string, id: string, user?: string): Promise<boolean> {
     if (user) {
-      return await this.tryCheck(ws, user);
+      const hasPermission = await this.tryCheck(ws, user);
+      if (hasPermission) return true;
+    }
+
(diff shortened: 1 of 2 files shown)


apps/server/src/modules/workspaces/permission.ts
```diff
@@ -47,37 +47,38 @@ export class PermissionService {
...
-    } else {
-      // check if this is a public workspace
-      const count = await this.prisma.workspace.count({
-        where: { id: ws, public: true },
-      });
-      if (count > 0) {
-        return true;
-      }
+      const hasPermission = await this.tryCheck(ws, user);
+      if (hasPermission) return true;
```
Brooooooklyn, line 51:
```suggestion
      return this.tryCheck(ws, user)
```

darkskygit: if user not have full workspace permission, we need continue check sub page permission

Brooooooklyn (approved):

PR #4430, opened 2023-09-20
feat: only enable playground in dev env

darkskygit:


Brooooooklyn:
why, it’s not harmful that enable playground

apps/server/src/graphql.module.ts
```diff
@@ -32,6 +32,8 @@ import { Metrics } from './metrics/metrics';
             res,
           }),
           plugins: [new GQLLoggerPlugin(metrics)],
+          // only able to use playground in dev mode
+          playground: config.affine.canary,
```
forehalo, line 36: should disable introspection at the same time if you want to disable playground in prod

darkskygit:
after discussion, there is currently no need to disable

PR #4433, opened 2023-09-20
feat: add shortcut keys for toggling setting modal

TinsFox:
Add shortcut keys for toggling setting modal

⌘ + , or Ctrl + ,

Because of the use of a new design, it has resulted in many conflicts with the [previous PR](https://github.com/toeverything/AFFiNE/pull/4160). Now a new one has been opened.


himself65 (changes requested):
I think we should move shortcut into atoms

joooye34 (commented):
@pengx17 cc and check the shortcut key usage.

pengx17 (changes requested):
Thanks for contribution. After we refactored CMDK, you can configure keybindings in https://github.com/toeverything/AFFiNE/blob/master/apps/core/src/commands/affine-navigation.tsx#L88-L102

e.g.,

```tsx
  unsubs.push(
    registerAffineCommand({
      id: 'affine:open-settings',
      category: 'affine:navigation',
      icon: <ArrowRightBigIcon />,
      keyBinding: {
        binding: '$mod+,',
      },
      label: () => t['com.affine.cmdk.affine.navigation.open-settings'](),
      run() {
        store.set(openSettingModalAtom, {
          activeTab: 'appearance',
          workspaceId: null,
          open: true,
        });
      },
    })
  );
```

PR #4440, opened 2023-09-21
feat(server): nestjs stripe integration

forehalo:
add

- four environment variables for using stripe api
```
STRIPE_API_KEY=
STRIPE_WEBHOOK_KEY=
STRIPE_YEARLY_PRICE_ID=
STRIPE_MONTHLY_PRICE_ID=
```

- typed guarded  `@OnEvent` in `payment/service.ts` for listening on stripe webhook events
documentation: https://stripe.com/docs/webhooks


apps/server/src/modules/payment/stripe.ts
```diff
@@ -0,0 +1,18 @@
+import { FactoryProvider } from '@nestjs/common';
+import { omit } from 'lodash-es';
+import Stripe from 'stripe';
```
joooye34, line 3: Do you need to install it? I don't see dep added in package.json.

forehalo: installed in target PR already, need some startup changes 🐸

joooye34 (approved):

PR #4449, opened 2023-09-21
test: fix flaky test

himself65:


himself65:
Fixed in https://github.com/toeverything/AFFiNE/pull/4454

Issue #4464, opened 2023-09-22
storybook preview regression


himself65:
https://github.com/toeverything/AFFiNE/commit/e0063ebc9bf5ddef81cf508b4ddab56b319a30ca

pengx17:
@himself65 do you mean to approve changes in Chromatic right? Sorry I missed these. Will ask the designers to check them.

himself65:
No, didn't you see the preview lost the search component?

himself65:
![image](https://github.com/toeverything/AFFiNE/assets/14026360/a14e9130-1575-4139-95be-3e6ecbf15b11)

pengx17:
Will add it back today

PR #4504, opened 2023-09-26
fix: storybook previews

pengx17:
fix https://github.com/toeverything/AFFiNE/issues/4464

### DIFF
--- a/apps/storybook/src/stories/core.stories.tsx
+++ b/apps/storybook/src/stories/core.stories.tsx
@@ -101,7 +101,7 @@ WorkspaceList.play = async ({ canvasElement }) => {
     }
   );
 
-  // todo: figure out why userEvent cannot click this element
+  // todo: figure out why userEvent cannot click this element?
   // await userEvent.click(currentWorkspace);
   currentWorkspace.click();
 };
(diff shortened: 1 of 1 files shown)


himself65 (approved):

apps/storybook/src/stories/core.stories.tsx
```diff
@@ -89,21 +89,21 @@ export const WorkspaceList: StoryFn = () => {
...
     }
   );
-  const currentWorkspace = canvasElement.querySelector(
-    '[data-testid="current-workspace"]'
-  ) as Element;
-  await userEvent.click(currentWorkspace);
+
+  // todo: figure out why userEvent cannot click this element?
+  // await userEvent.click(currentWorkspace);
+  currentWorkspace.click();
```
joooye34, line 106: Is this cause screen snapshot change ?

pengx17: I am not sure. If you try it locally, `userEvent.click` on currentWorkspace has no effect at all, which is strange.

joooye34 (approved):
No UI been effected, it seems fine.

PR #4506, opened 2023-09-26
fix(server): wrong member count query

forehalo:


apps/server/src/modules/workspaces/resolver.ts
```diff
@@ -221,9 +224,6 @@ export class WorkspaceResolver {
         {
           createdAt: 'asc',
         },
-        {
-          type: 'desc',
-        },
```
pengx17, line 226: This is to make sure the `owner` type is at the top of the list.

darkskygit (approved):

PR #4507, opened 2023-09-26
fix(component): content should subtract height of the header

### LINKED ISSUE #4461: The safe area with padding at the bottom of the page is missing, and there is also a white bar that covers the content
1. The lack of sufficient safe area will affect the experience of content editing and reading.

CleanShot 2023-09-22 at 17 18 34@2x

2. Now there is also a white bar at the bottom.

CleanShot 2023-09-22 at 17 20 10@2x

joooye34:
Close #4461 

Temporarily, content subtract heights of the header to fix safe padding issue. There need to use flex to refactor the whole layout.

Temporarily, use editor classsname to add padding bottom back, which removed on https://github.com/toeverything/blocksuite/pull/4766

### DIFF
--- a/apps/core/src/components/page-detail-editor.css.ts
+++ b/apps/core/src/components/page-detail-editor.css.ts
@@ -15,7 +15,7 @@ export const pluginContainer = style({
 });
 
 export const editor = style({
-  height: 'calc(100% - 52px)',
+  height: '100%',
   selectors: {
     '&.full-screen': {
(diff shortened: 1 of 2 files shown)


pengx17 (approved):

himself65 (approved):

houjoe0829 (approved):
From an implementation perspective, it meets the expected product effect.

PR #4511, opened 2023-09-26
build(electron): allow customizing channel type for internal build

pengx17:
https://www.electronjs.org/docs/latest/api/app
Electron stores application data in `appData`, which contains the data for storage, idb etc, but not cookies.

image


himself65 (dismissed):

joooye34 (commented):
Why the title is a fix ?

pengx17:
> Why the title is a fix ?

Changed to build

joooye34 (approved):

PR #4530, opened 2023-09-27
feat: new collections 

zzj3720:
[Internal doc](https://insider.affine.pro/share/af3478a2-9c9c-4d16-864d-bffa1eb10eb6/IKmsDa5NXK)

This PR introduces the new Collections. Scope

- New interactive for creating Collections
- All Collections are directly displayed in the sidebar
- Support adding pages, including adding pages within a Collection and adding individual Pages to a Collection in Page options
- Support adding rules, with whitelist mechanism for rule effectiveness
- Support removing pages from a Collection
- Support collaboration on Collections, where Collections belong to Workspaces rather than specific accounts
- Support drag and drop, allowing Pages from the Page list to be dragged and added to a Collection

In order to do so, we split the work into two parts

## Collection manager & UI panels

The goal is to have a collection manager that 
1. Migrate old collections data (merge into one).
    - Data in idb: migrate workspace id matched collections data
    - Data in user setting: migrate all collections data
3. Add settings properties to save workspace setting data, make sure it can be synced across all users within the workspace
4. Given a set of filters (collection) and a list page, should be able to return the valid pages.
5. New filter panel UIs.

## New Page List

Goal: given a list of page, rendering the page in a list. It should support a list of "column" options that shows individual row items, like "page name", created date, tags, etc. Should also support drag (& drop ?).

### implementation considerations
**modularization**
Old PageList is a all-in-one table that is intended for all pages and trash etc. Though the design is powerful, but not very flexible.
In the new collections requirement, we will have the page list to be used in a CMDK like modal dialog and we may use page list without filters and table header.

**abandon table**
In the new implementation, we will use flexbox rows with without "table" to make implementation more flexible


joooye34:
Weird twinkling when go to collection path
https://github.com/toeverything/AFFiNE/assets/9527419/0e82876f-0142-48af-a68e-c41378250dbe

joooye34:
Should redirect to new collection location, but not.

https://github.com/toeverything/AFFiNE/assets/9527419/03c77b8b-0d94-4ec9-9a2a-4c23ead0ed4e

joooye34 (approved):

Issue #4498, opened 2023-09-25
remove bookmark from plugin


himself65:
The electron side plugin needs to be considered later

Related: https://github.com/toeverything/AFFiNE/pull/4141

himself65:
link-preview-js should be removed

PR #4544, opened 2023-09-30
refactor: remove bookmark plugin

himself65:
Bookmark will not be considered a plugin as it needs server functionality but our plugin system is not ready on the electron side. This will make it extra cost for now to implement everything just for a bookmark plugin.

The electron side plugin should be considered in the future after the browser-only plugin is ready.

Closes: https://github.com/toeverything/AFFiNE/issues/4498


l2dy (commented):
Did you forget to remove this?

https://github.com/toeverything/AFFiNE/blob/69db99636becef2892143832118f3827563b7634/packages/infra/src/__internal__/plugin.ts#L10

PR #4568, opened 2023-10-09
feat(core): add setting commands

JimmFly:
### Summary

This pull request adds new commands and components to the AFFiNE app to improve the user experience and customization options. It also enhances the performance and functionality of some existing components and hooks.

### DIFF
--- a/apps/core/src/hooks/affine/use-language-helper.ts
+++ b/apps/core/src/hooks/affine/use-language-helper.ts
@@ -1,0 +1,27 @@
+import { LOCALES, useI18N } from '@affine/i18n';
+import { useCallback, useMemo } from 'react';
+
+export function useLanguageHelper() {
+  const i18n = useI18N();
+  const currentLanguage = useMemo(
+    () => LOCALES.find(item => item.tag === i18n.language),
+    [i18n.language]
+  );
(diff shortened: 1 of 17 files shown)


apps/core/src/hooks/affine/use-language-helper.ts
```diff
@@ -0,0 +1,27 @@
...
+    []
+  );
+  const onSelect = useCallback(
+    (event: string) => {
+      i18n.changeLanguage(event);
+    },
+    [i18n]
+  );
+
+  return { languagesList, currentLanguage, onSelect };
```
pengx17, line 26: need to wrap this with `useMemo` otherwise it will cause commands to register/unregister randomly

pengx17 (changes requested):
Generally good. Can you please check the console logs and see if commands are registered/unregistered multiple randomly?

apps/core/src/hooks/affine/use-language-helper.ts
JimmFly: Thank you for the guidance, I'll check and amend it tomorrow.

pengx17:
There is a failing test, please fix it

pengx17 (approved):

Issue #4570, opened 2023-10-10
Disabled form item in the settings can still be operated via keyboard


CatsJuice:
### What happened?

The `Publish` toggle is disabled, but it can still be focused and operated via keyboard.

https://github.com/toeverything/AFFiNE/assets/39363750/f3f56082-4607-4a6f-98e5-3c0da0a00275

### Distribution version

macOS ARM 64 (Apple Silicon)

### What browsers are you seeing the problem on if you're using web version?

Chrome

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

Noothan-am:
hey @CatsJuice, i am not able reproduce this issue please give me more steps.

CatsJuice:
> hey @CatsJuice, i am not able reproduce this issue please give me more steps.

1. open `settings`
2. click `workspace name` input to focus on it
3. toggle with `Tab` until the publish tooltip is displayed
4. press `Space` bar

aditya-sodhiya:
@JimmFly please assign this to me. I can take this up.
/attempt #4570

JimmFly:
> @JimmFly please assign this to me. I can take this up. /attempt #4570

I'm overwhelmed by your proactive contributions, could you all discuss amongst yourselves who will fix this issue?
/cc @Noothan-am @aditya-sodhiya

joooye34:
@aditya-sodhiya Thank you so much for your PR. The PR from @electron97 is ok and we merged. So glad to see your PR next time.

PR #4605, opened 2023-10-14
fix: disabled form item in the settings can still be operated via keyboard

electron97:
## Summary
- Currently, while pressing the tab key the [Tooltip](https://github.com/toeverything/AFFiNE/blob/master/apps/core/src/components/affine/new-workspace-setting-detail/publish.tsx#L100) component gets triggered which puts the focus on it, and after pressing the space key it goes to its child node Switch which activates its event and still gets trigger in disable mode.
- To fix the issue we need to pass a prop onChange which on getting a trigger will not enable the switch.

Fixes: https://github.com/toeverything/AFFiNE/issues/4570

## Videos

**Before**

https://github.com/toeverything/AFFiNE/assets/132813311/b7a79f96-283c-4be0-92a8-4e47baf1d2f4

**After**

https://github.com/toeverything/AFFiNE/assets/132813311/4e3365c3-a966-4fe7-9ca4-0b96e5a08523

### DIFF
--- a/apps/core/src/components/affine/new-workspace-setting-detail/publish.tsx
+++ b/apps/core/src/components/affine/new-workspace-setting-detail/publish.tsx
@@ -100,7 +100,7 @@ const FakePublishPanelAffine = (_props: FakePublishPanelAffineProps) => {
     <Tooltip content={t['com.affine.settings.workspace.publish-tooltip']()}>
       <div className={style.fakeWrapper}>
         <SettingRow name={t['Publish']()} desc={t['Unpublished hint']()}>
-          <Switch checked={false} />
+          <Switch checked={false} onChange={() => {}} />
         </SettingRow>
       </div>
     </Tooltip>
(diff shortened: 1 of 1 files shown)


electron97:
Hey @JimmFly ,
I have raised this issue to fix `disabled form item in the settings can still be operated via keyboard`.
Could you please take a look whenever you get a chance?
Thanks!

apps/core/src/components/affine/new-workspace-setting-detail/publish.tsx
```diff
@@ -100,7 +100,7 @@ const FakePublishPanelAffine = (_props: FakePublishPanelAffineProps) => {
     <Tooltip content={t['com.affine.settings.workspace.publish-tooltip']()}>
       <div className={style.fakeWrapper}>
         <SettingRow name={t['Publish']()} desc={t['Unpublished hint']()}>
-          <Switch checked={false} />
+          <Switch checked={false} onChange={() => {}} />
```
himself65:
```suggestion
          <Switch checked={false} onChange={noop} />
```

himself65 (changes requested):

apps/core/src/components/affine/new-workspace-setting-detail/publish.tsx
electron97: Thanks for the feedback.
Updated to use `onChange={noop}`. @himself65  could you please take another look?

himself65 (approved):

PR #4657, opened 2023-10-19
fix: wrong confirm button position

JimmFly:
before：

image

after:

image

### DIFF
--- a/packages/frontend/core/src/components/affine/enable-affine-cloud-modal/index.tsx
+++ b/packages/frontend/core/src/components/affine/enable-affine-cloud-modal/index.tsx
@@ -50,9 +50,6 @@ export const EnableAffineCloudModal = ({
             ? t['Enable']()
             : t['Sign in and Enable'](),
       }}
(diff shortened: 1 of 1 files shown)


Brooooooklyn (approved):

packages/frontend/core/src/components/affine/enable-affine-cloud-modal/index.tsx
```diff
@@ -50,9 +50,6 @@ export const EnableAffineCloudModal = ({
             ? t['Enable']()
             : t['Sign in and Enable'](),
       }}
-      contentOptions={{
-        ['data-testid' as string]: 'enable-cloud-modal',
-      }}
```
pengx17, line 55: I think we should patch the upstream package so that it won't be an issue applying options here

JimmFly: Yes, but I don’t think it’s for now, it’s rather urgent.

PR #4708, opened 2023-10-24
feat(payment-plans): full width scroll area for plans

CatsJuice:


packages/frontend/core/src/components/affine/setting-modal/index.tsx
```diff
@@ -37,6 +37,39 @@ export const SettingModal = ({
 
   const generalSettingList = useGeneralSettingList();
 
+  const modalContentRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (!modalProps.open) return;
+    let animationFrameId: number;
+    const onResize = () => {
```
joooye34, line 45: add a debounce

joooye34 (approved):

PR #4714, opened 2023-10-24
feat(core): auto scroll to current payment plan

CatsJuice:


packages/frontend/core/src/components/affine/setting-modal/general-setting/plans/index.tsx
```diff
@@ -175,18 +204,19 @@ const Settings = () => {
...
-        <div className={styles.planCardsWrapper}>
+        <div className={styles.planCardsWrapper} ref={scrollWrapper}>
           {Array.from(planDetail.values()).map(detail => {
+            const isCurrent =
+              loggedIn &&
+              detail.plan === currentPlan &&
+              (currentPlan === SubscriptionPlan.Free
+                ? true
+                : currentRecurring === recurring);
             return (
```
joooye34, line 215: Need a PriceCard component to refactor those codes, and style is incorrect now. 
image

PR #4720, opened 2023-10-25
feat: dynamic user usage control

### LINKED ISSUE #4526: Usage controlling
Add flexible usage controlling to AFFiNE.

Currently There is only a simple blob size limit in blob service and the throttle is hard coded in codebase. The blob size calculation is also lazy by summing at query time.

For the seen future, we may extend the controlling items from blob only to more aspects, for instance the amount of cloud workspaces, the maximum collaborators a workspace could have for different plans. This requires the server make usage controlling dynamic and configurable.

We may make the usage controlling a standalone service, with configuration connected with plans(user subscription), easy to use check and quota API, and so on...

```ts
class Workspace {
  constructor(usage: Usage) {}

  invite(workspace: Workspace, user: User) {
    if (!this.usage.canAddMembers(workspace)) {
      // exceed the limit of current plan
    }

    // add user to workspace
  }
}

```

quota query example from GitHub

image

quota query example from Perfsee

image

darkskygit:
this pr impl the new feature gate & quota manage
provide multiple profile for earh user to allow switch user's quota
replace origin early access list & blob size control
fix #4526


packages/backend/server/src/modules/auth/next-auth-options.ts
```diff
@@ -37,11 +35,35 @@ export const NextAuthOptionsProvider: FactoryProvider<NextAuthOptions> = {
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     const createUser = prismaAdapter.createUser!.bind(prismaAdapter);
     prismaAdapter.createUser = async data => {
+      const latestFreePlan = await prisma.userFeatures.aggregate({
+        where: {
+          feature: 'free_plan_v1',
```
forehalo, line 40: make it enumeration?

packages/backend/server/src/modules/auth/next-auth-options.ts
```diff
@@ -37,11 +35,35 @@ export const NextAuthOptionsProvider: FactoryProvider<NextAuthOptions> = {
...
+
       const userData = {
         name: data.name,
         email: data.email,
         avatarUrl: '',
         emailVerified: data.emailVerified,
+        features: {
+          create: {
+            reason: 'created by email sign up',
+            expiresAt: new Date('2021-12-31T23:59:59.000Z'),
```
forehalo, line 55: free will expire?

packages/backend/server/src/modules/users/resolver.ts
```diff
@@ -214,27 +204,55 @@ export class UserResolver {
...
   })
-  @Mutation(() => AddToNewFeaturesWaitingList)
-  async addToNewFeaturesWaitingList(
-    @CurrentUser() user: UserType,
-    @Args('type', {
-      type: () => NewFeaturesKind,
-    })
-    type: NewFeaturesKind,
+  @Mutation(() => String)
+  async addToEarlyAccess(
```
forehalo, line 208: no permission controlling?

packages/backend/server/src/modules/auth/next-auth-options.ts
darkskygit: 2099, typo

packages/backend/server/migrations/20231010081211_user_features_quotas/migration.sql
```diff
@@ -0,0 +1,31 @@
...
+
+*/
+-- AlterTable
+ALTER TABLE "user_feature_gates" DROP COLUMN "feature",
+ADD COLUMN     "activated" BOOLEAN NOT NULL DEFAULT false,
+ADD COLUMN     "expires_at" TIMESTAMPTZ(6) NOT NULL,
+ADD COLUMN     "featureId" VARCHAR NOT NULL;
+
+-- CreateTable
+CREATE TABLE "user_features" (
```
forehalo, line 16: change the name to `features`, with the `user_` prefix indicates it is related with users already

packages/backend/server/migrations/20231010081211_user_features_quotas/migration.sql
```diff
@@ -0,0 +1,31 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `feature` on the `user_feature_gates` table. All the data in the column will be lost.
+  - Added the required column `expires_at` to the `user_feature_gates` table without a default value. This is not possible if the table is not empty.
+  - Added the required column `featureId` to the `user_feature_gates` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE "user_feature_gates" DROP COLUMN "feature",
```
forehalo, line 10: and this is `user_features`

packages/backend/server/schema.prisma
```diff
@@ -48,20 +48,53 @@ model User {
...
   @@map("users")
 }
 
-model UserFeatureGates {
+// feature gates is a way to enable/disable features for a user
+// for example:
+// - early access is a feature that allow some users to access the insider version
+// - pro plan is a quota that allow some users access to more resources after they pay
+model UserFeatures {
+  id        String @id @default(uuid()) @db.VarChar
```
forehalo, line 61: id is not user-facing one, better to be `auto_incremental`

packages/backend/server/schema.prisma
```diff
@@ -48,20 +48,53 @@ model User {
...
+  userId    String @map("user_id") @db.VarChar
```
forehalo, line 62: restrict to `Varchar(36)`

packages/backend/server/schema.prisma
```diff
@@ -48,20 +48,53 @@ model User {
...
+  activated Boolean  @default(false)
+
+  user    User     @relation(fields: [userId], references: [id], onDelete: Cascade)
+  feature Features @relation(fields: [featureId], references: [id], onDelete: Cascade)
+
+  @@map("user_features")
+}
+
+model Features {
   id        String   @id @default(uuid()) @db.VarChar
```
forehalo, line 85: same

packages/backend/server/schema.prisma
```diff
@@ -48,20 +48,53 @@ model User {
...
+  featureId String @db.VarChar
+
+  // we will record the reason why the feature is enabled/disabled
+  // for example:
+  // - pro_plan_v1: "user buy the pro plan"
+  reason    String   @db.VarChar
+  // record the quota enabled time
+  createdAt DateTime @default(now()) @map("created_at") @db.Timestamptz(6)
+  // record the quota expired time, pay plan is a subscription, so it will expired
+  expiresAt DateTime @map("expires_at") @db.Timestamptz(6)
```
forehalo, line 72: does it make more sense to use `expiredAt` so we don't need to hard code a far date for free feature?

packages/backend/server/src/modules/quota/configure/types.ts
```diff
@@ -0,0 +1,58 @@
...
+      storageQuota: 10 * 1024 * 1024 * 1024,
+    },
+  },
+  {
+    feature: FeatureType.Quota_ProPlanV1,
+    type: FeatureKind.Quota,
+    version: 1,
+    configs: {
+      // single blob limit 10MB
+      blobLimit: 10 * 1024 * 1024,
```
forehalo, line 53: 500M?

packages/backend/server/src/modules/workspaces/permission.ts
```diff
@@ -21,6 +21,25 @@ export class PermissionService {
...
+  async getOwnedWorkspaces(userId: string) {
+    return this.prisma.userWorkspacePermission
+      .findMany({
+        where: {
+          userId,
+          accepted: true,
+          type: Permission.Owner,
+        },
+        select: {
+          workspace: {
```
forehalo, line 33: there is workspace id in permission table, no need to join query if only id needed

packages/backend/server/src/modules/workspaces/resolver.ts
```diff
@@ -82,6 +83,9 @@ export class WorkspaceType implements Partial<Workspace> {
     description: 'Members of workspace',
   })
   members!: InviteUserType[];
+
+  @Field({ description: 'Blobs size of workspace' })
+  blobsSize!: number;
```
forehalo, line 88: field resolver at L255 will add filed automatically, no need to manually adding it

packages/backend/server/src/modules/quota/feature.ts
```diff
@@ -0,0 +1,78 @@
+import { Injectable, Logger } from '@nestjs/common';
```
forehalo, line 1: seems `feature` is a standalone service, should it be put under quota folder?

packages/backend/server/src/modules/workspaces/resolver.ts
darkskygit: members also has a resolver, what different about both?

forehalo: oh, that must be a mistake, I will clean graphql resolvers once I have time.

packages/backend/server/schema.prisma
```diff
@@ -48,23 +48,57 @@ model User {
...
+  featureId Int    @map("feature_id") @db.Integer
+
+  // we will record the reason why the feature is enabled/disabled
+  // for example:
+  // - pro_plan_v1: "user buy the pro plan"
+  reason    String    @db.VarChar
+  // record the quota enabled time
+  createdAt DateTime  @default(now()) @map("created_at") @db.Timestamptz(6)
+  // record the quota expired time, pay plan is a subscription, so it will expired
+  expiresAt DateTime? @map("expires_at") @db.Timestamptz(6)
```
forehalo, line 76: `expiredAt`

packages/backend/server/src/modules/auth/next-auth-options.ts
```diff
@@ -42,6 +41,18 @@ export const NextAuthOptionsProvider: FactoryProvider<NextAuthOptions> = {
         email: data.email,
         avatarUrl: '',
         emailVerified: data.emailVerified,
+        features: {
+          create: {
+            reason: 'created by email sign up',
+            expiresAt: new Date('2099-12-31T23:59:59.000Z'),
```
forehalo, line 47: now, it's `expired` flag, no need to set a default value for it.

packages/backend/server/src/modules/auth/service.ts
```diff
@@ -190,6 +191,18 @@ export class AuthService {
         name,
         email,
         password: hashedPassword,
+        features: {
+          create: {
+            reason: 'created by api sign up',
+            expiresAt: new Date('2099-12-31T23:59:59.000Z'),
```
forehalo, line 197: same here

packages/backend/server/src/modules/quota/configure/feature.ts
```diff
@@ -0,0 +1,210 @@
...
+];
+
+@Injectable()
+export class FeatureService implements OnModuleInit {
+  constructor(private readonly prisma: PrismaService) {}
+
+  async onModuleInit() {
+    // upgrade features from lower version to higher version
+    for (const feature of Features) {
+      await this.upsertFeature(feature);
```
forehalo, line 25: can be moved to data-migration script now
`yarn data-migration create`
`yarn data-migration run`

packages/backend/server/src/modules/quota/configure/quota.ts
```diff
@@ -0,0 +1,149 @@
...
+              feature_version: {
+                feature: quota,
+                version: latestFreePlan._max.version || 1,
+              },
+              type: FeatureKind.Quota,
+            },
+          },
+          reason: reason ?? 'switch quota',
+          activated: true,
+          expiresAt: expiresAt ?? '2099-12-31T23:59:59.999Z',
```
forehalo, line 129: nullable

packages/backend/server/src/modules/users/resolver.ts
```diff
@@ -220,27 +210,55 @@ export class UserResolver {
...
+
+  @Throttle({
+    default: {
+      limit: 10,
+      ttl: 60,
+    },
+  })
+  @Query(() => [UserType])
+  async listEarlyAccess(@CurrentUser() user: UserType): Promise<UserType[]> {
+    if (!this.feature.isStaff(user.email)) {
```
forehalo, line 259: seems we need sys-admin accounts now, let's add a todo

darkskygit: Considering whether we need to implement a feature-based implementation to allow specific accounts to have the ability to adjust features, or use a separate user pool to provide backend management capabilities—in our scenario, it seems a bit strange to include the management background in affine.
Maybe we should discuss this problem when removing `next-auth`

darkskygit:
split large pull into #5074 #5075 #5076

PR #4728, opened 2023-10-26
feat(server): auto attach early access coupon

forehalo:


packages/backend/server/src/modules/payment/schedule.ts
```diff
@@ -0,0 +1,214 @@
+import { Injectable } from '@nestjs/common';
+import Stripe from 'stripe';
+
+@Injectable()
+export class ScheduleManager {
+  #schedule: Stripe.SubscriptionSchedule | null = null;
```
Brooooooklyn, line 6: oh shift, I don't like `#`

joooye34: 😞

packages/backend/server/src/modules/payment/schedule.ts
```diff
@@ -0,0 +1,214 @@
...
+  async update(price: string, coupon?: string) {
+    if (!this.#schedule) {
+      throw new Error('No schedule');
+    }
+
+    if (!this.isActive || !this.currentPhase) {
+      throw new Error('Unexpected subscription schedule status');
+    }
+
+    // if current phase's plan matches target, or no coupon change, just release the schedule
```
joooye34, line 182: it is and logic, not or.

joooye34 (approved):

PR #4738, opened 2023-10-27
feat(core): add upgrade success page

JimmFly:


packages/backend/server/src/modules/payment/resolver.ts
```diff
@@ -195,7 +195,7 @@ export class SubscriptionResolver {
       user,
       recurring,
       // TODO: replace with frontend url
```
forehalo, line 197: delete

forehalo: and https://github.com/toeverything/AFFiNE/blob/7ecee01d2078736982e74f186548dc34780f6ee7/packages/backend/server/src/modules/payment/webhook.ts#L35

JimmFly: Based on your advice, I’ve made some adjustments. Do the changes meet your expectations?

joooye34 (approved):

PR #4744, opened 2023-10-27
feat(core): payment plans error boundary

CatsJuice:


packages/frontend/core/src/components/pure/swr-error-bundary.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  const onErrorWithSWR = React.useCallback(
+    (error: Error, info: ErrorInfo) => {
+      errorsRef.current.push(error);
+      onError?.(error, info);
+    },
+    [onError]
+  );
+
+  React.useEffect(() => clearErrorCache, [clearErrorCache]);
```
CatsJuice, line 49: I've added this logic to ensure that resend request when component reloaded(e.g. switch tab and back). @joooye34

joooye34: ok...that Is cool

joooye34 (dismissed):

packages/frontend/core/src/components/pure/swr-error-bundary.tsx
joooye34: Why is this function not executed

joooye34 (approved):

PR #4753, opened 2023-10-29
feat: add idempotent request support for payment apis

darkskygit:


packages/backend/server/src/modules/payment/resolver.ts
```diff
@@ -189,12 +189,14 @@ export class SubscriptionResolver {
   async checkout(
     @CurrentUser() user: User,
     @Args({ name: 'recurring', type: () => SubscriptionRecurring })
-    recurring: SubscriptionRecurring
+    recurring: SubscriptionRecurring,
+    @Args('idempotencyKey') idempotencyKey: string
```
forehalo, line 193: the checkout session is ok to not be non-idempotent

packages/backend/server/src/modules/payment/resolver.ts
```diff
@@ -209,27 +211,41 @@ export class SubscriptionResolver {
   @Mutation(() => String, {
     description: 'Create a stripe customer portal to manage payment methods',
   })
-  async createCustomerPortal(@CurrentUser() user: User) {
-    return this.service.createCustomerPortal(user.id);
+  async createCustomerPortal(
+    @CurrentUser() user: User,
+    @Args('idempotencyKey') idempotencyKey: string
+  ) {
+    return this.service.createCustomerPortal(idempotencyKey, user.id);
```
forehalo, line 218: same

packages/backend/server/src/modules/payment/resolver.ts
darkskygit: stripe.customers.create need to be idempotent

PR #4773, opened 2023-10-30
fix(component): adjust autofill style

### LINKED ISSUE #4769: Quick fill style is wrong
image

JimmFly:
fix #4769


packages/frontend/component/src/ui/input/style.css.ts
```diff
@@ -74,6 +74,11 @@ export const input = style({
     '&::placeholder': {
       color: 'var(--affine-placeholder-color)',
     },
+    '&:autofill, &:-webkit-autofill, &:-internal-autofill-selected, &:-webkit-autofill:hover, &:-webkit-autofill:focus, &:-webkit-autofill:active':
+      {
+        backgroundColor: 'var(--affine-white-10) !important',
```
joooye34, line 79: why use important ? write some comments to record

JimmFly: The user agent style sheets of many browsers use !important in their :-webkit-autofill style declarations

JimmFly: https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill

joooye34 (approved):

PR #4786, opened 2023-10-31
fix: request prevented when re-downgrade subscribe

### LINKED ISSUE #4782: downgrade twice will not execute
downgrade -> resume -> downgrade will not execute success

image

its a prevent-replay problem

darkskygit:
fix #4782

### DIFF
--- a/packages/frontend/core/src/components/affine/setting-modal/general-setting/plans/plan-card.tsx
+++ b/packages/frontend/core/src/components/affine/setting-modal/general-setting/plans/plan-card.tsx
@@ -301,18 +301,19 @@ const Downgrade = ({
 }) => {
   const t = useAFFiNEI18N();
   const [open, setOpen] = useState(false);
+  // allow replay request on network error until component unmount or success
+  const [idempotencyKey, setIdempotencyKey] = useState(nanoid());
   const { isMutating, trigger } = useMutation({
     mutation: cancelSubscriptionMutation,
   });
-
-  // allow replay request on network error until component unmount
-  const idempotencyKey = useMemo(() => nanoid(), []);
 
   const downgrade = useCallback(() => {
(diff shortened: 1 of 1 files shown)


packages/frontend/core/src/components/affine/setting-modal/general-setting/plans/plan-card.tsx
```diff
@@ -532,19 +534,20 @@ const ResumeAction = ({
...
-  // allow replay request on network error until component unmount
-  const idempotencyKey = useMemo(() => nanoid(), []);
-
   const resume = useCallback(() => {
     trigger(
       { idempotencyKey },
       {
         onSuccess: data => {
+          // refresh idempotency key
+          setIdempotencyKey(nanoid());
```
joooye34, line 550: how about error occur?

darkskygit: In our scenario, there are four possible errors:

1. network error
2. server error
3. stripe gateway exception
4. stripe believes that there is something wrong with the information transmitted

These four errors occur in order. When the first three errors occur, the ideapotency key does not take effect, so the client can rest assured to use the same key request.
When the last error occurs, it means that the subscription information of the corresponding user is problematic (for example, he has not started subscribing, or the subscription has ended), but tries to call the resume subscription.
At this time, it is appropriate for the stripe to directly use the cache response corresponding to the ideapotencyKey.

joooye34 (approved):

joooye34 (approved):

joooye34:
Commit need to be signed, seems need recommit -S

darkskygit:
> Commit need to be signed, seems need recommit -S

fixed

PR #4788, opened 2023-10-31
feat(core): support subscribe plan after login

joooye34:
Use `${baseUrl}/signin?subscription_recurring=Yearly&subscription_plan=Pro` to login and subscribe plan

### DIFF
--- a/packages/frontend/core/src/components/affine/auth/after-sign-in-send-email.tsx
+++ b/packages/frontend/core/src/components/affine/auth/after-sign-in-send-email.tsx
@@ -7,13 +7,14 @@ import {
 import { Trans } from '@affine/i18n';
 import { useAFFiNEI18N } from '@affine/i18n/hooks';
 import { Button } from '@toeverything/components/button';
-import { useCallback } from 'react';
+import React, { useCallback } from 'react';
 
 import { useCurrentLoginStatus } from '../../../hooks/affine/use-current-login-status';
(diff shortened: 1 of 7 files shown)


packages/frontend/core/src/components/affine/auth/after-sign-in-send-email.tsx
```diff
@@ -75,26 +85,26 @@ export const AfterSignInSendEmail = ({
       </div>
 
       <div className={style.authMessage} style={{ marginTop: 20 }}>
-        {/*prettier-ignore*/}
```
pengx17, line 78: Please double check if i18n still works correctly after prettier ignore comment is removed

packages/frontend/core/src/components/affine/auth/use-auth.ts
```diff
@@ -75,7 +77,9 @@ export const useAuth = () => {
         'email',
         {
           email: email,
-          callbackUrl: '/auth/signIn',
+          callbackUrl: subscriptionData
+            ? subscriptionData.redirectUrl
+            : '/auth/signIn',
```
pengx17, line 82: I can see this callbackURL could have some other cases that may also change its behavior based on the route params. Maybe it is better to redesign singInCloud somehow (not necessary in this PR).

pengx17 (approved):

PR #4791, opened 2023-10-31
feat(server): sync data with ack

### LINKED ISSUE #4755: Add ACK strategy to WebSocket Client/Server to reduce the pendingStruct situations
For the Client side, if any operation binary isn't ACK by the server, the following operations should not be sent to the server, and the connection status should be `reconnecting`.  The status goes back to the `Synced` only if the operation binary was resent to the Server and ACK by the server.

For the Server side, send the ack back to the server only if the operation binary was persisted in database.

forehalo:
- close #4755


packages/common/y-provider/src/lazy-provider.ts
```diff
@@ -332,6 +332,7 @@ export const createLazyProvider = (
   }
 
   async function disconnect() {
+    console.log('disconnect', origin);
```
Brooooooklyn, line 335: remove

packages/frontend/core/src/index.tsx
```diff
@@ -18,11 +18,9 @@ async function main() {
   assertExists(root);
 
   createRoot(root).render(
-    <StrictMode>
```
Brooooooklyn, line 21: why remove `StrictMode`

forehalo: I I don't don't wanna wanna my my code code to to be be executed executed twice twice

packages/frontend/workspace/src/affine/index.ts
```diff
@@ -156,11 +209,9 @@ function setupAffineAwareness(
...
   };
 
-  const awarenessUpdate = (changes: AwarenessChanges, origin: unknown) => {
-    if (origin === 'server') {
-      return;
-    }
-
+  // debounce awareness update to avoid too many updates
+  // BlockSuite current raise too many same awareness updates
+  const sync = debounce((changes: AwarenessChanges) => {
```
darkskygit, line 214: Maybe we can exclude most of the broadcasts by caching the last sended awareness? In the old keck server, the broadcast number has been reduced by more than half just by comparing the last broadcast msg of the same user.

darkskygit: edgeless mode will sync the cursor position with awareness, and debounce 500ms may make the remote cursor lag

forehalo: > Maybe we can exclude most of the broadcasts by caching the last sended awareness? In the old keck server, the broadcast number has been reduced by more than half just by comparing the last broadcast msg of the same user.

oh, that's cool. let's do it

Brooooooklyn: debounce with `500ms` may lower the cursor fps 👀

packages/frontend/core/src/index.tsx
pengx17: Strict polic incoming 🚔

packages/frontend/workspace/src/affine/batch-sync-sender.ts
```diff
@@ -0,0 +1,104 @@
...
+    this.started = true;
+    this.next();
+  }
+
+  private head(): [string, Uint8Array[]] | undefined {
+    for (const pending of this.buffered.entries()) {
+      if (pending[1].length) {
+        return pending;
+      }
+    }
```
pengx17, line 57: If there are multiple subdoc updates in `.buffered`, it seems that the syncing jobs are serialized. do you think maybe can be parallelized to increase throughput?

forehalo: that makes sense

forehalo: done

packages/backend/server/src/modules/sync/events/events.gateway.ts
```diff
@@ -160,53 +202,138 @@ export class EventsGateway implements OnGatewayConnection, OnGatewayDisconnect {
...
     const state = Buffer.from(encodeStateVector(doc)).toString('base64');
 
-    endTimer();
     return {
       missing,
       state,
     };
   }
 
+  @SubscribeMessage('client-update-v2')
```
pengx17, line 211: maybe using same message name, but distinguish by a `version` field in the payload?

pengx17 (approved):
LGTM

packages/backend/server/src/modules/sync/events/events.gateway.ts
forehalo: that will make the function body too verbose, and by separating them, it will be easier to remove when the time coming(no more old clients)

pengx17 (approved):

packages/frontend/workspace/src/affine/crud.ts
```diff
@@ -85,6 +77,16 @@ export const CRUD: WorkspaceCRUD<WorkspaceFlavour.AFFINE_CLOUD> = {
       })
     );
 
+    const datasource = createAffineDataSource(
+      createWorkspace.id,
+      newBlockSuiteWorkspace.doc,
+      newBlockSuiteWorkspace.awarenessStore.awareness
+    );
+
+    const disconnect = datasource.onDocUpdate(() => {});
```
joooye34, line 86: why there need this and disconnect immediately ?

darkskygit (approved):

packages/frontend/workspace/src/affine/crud.ts
forehalo: this is not the real `connection` but the datasource and listeners behind, we need to drop the resources occupied

Brooooooklyn (approved):

Issue #4785, opened 2023-10-31
Issue list for 0.10


joooye34:
This version include following big features:
- payment system https://github.com/toeverything/AFFiNE/issues/4439
- new collections https://github.com/toeverything/AFFiNE/issues/4601
- new quick search https://github.com/toeverything/AFFiNE/issues/3294
- edgeless improvements https://github.com/toeverything/blocksuite/issues/5009

Testing SOP:
- https://whimsical.com/affine-payment-RNfCmJ1JkgDckKy9zuJx4H
- https://whimsical.com/new-collection-CYZGtDkvS6tsndgAmCYQrM
- https://whimsical.com/quick-search-ucNfiF8ezAtZS4E21o7ZL

```[tasklist]
### Issues
- [ ] https://github.com/toeverything/AFFiNE/issues/4782
- [ ] https://github.com/toeverything/AFFiNE/pull/4783
- [x] P1: paid plan selected state @CatsJuice
- [ ] https://github.com/toeverything/AFFiNE/pull/4789
- [ ] https://github.com/toeverything/AFFiNE/pull/4788
- [ ] https://github.com/toeverything/AFFiNE/pull/4787
- [ ] https://github.com/toeverything/AFFiNE/pull/4780
- [ ] https://github.com/toeverything/AFFiNE/pull/4776
- [x] P0: Confirg Dialog in Billing Setting @CatsJuice
- [x] P1: page list selection issues (esc/dbclick to exit selection, click to enable page selection, fav button not showing issue)
- [ ] https://github.com/toeverything/AFFiNE/pull/4797
- [ ] https://github.com/toeverything/AFFiNE/issues/4777
- [ ] https://github.com/toeverything/AFFiNE/issues/4759
- [x] P1: quick search category order should be (collections > page (title) > page (content) > commands) @JimmFly
- [ ] https://github.com/toeverything/AFFiNE/issues/4800
- [ ] https://github.com/toeverything/AFFiNE/issues/4388
- [x] P0: site pricing login and pay
- [ ] P2: currency follow stripe
- [ ] https://github.com/toeverything/AFFiNE/issues/4804
- [x] P0: subscription canceled status could not be paid member @CatsJuice
- [ ] https://github.com/toeverything/blocksuite/issues/5185
- [ ] https://github.com/toeverything/AFFiNE/issues/4828
- [ ] https://github.com/toeverything/blocksuite/issues/5184
- [ ] https://github.com/toeverything/blocksuite/issues/5089
[...]

PR #4797, opened 2023-11-01
fix(core): adjust payment related text

JimmFly:
work for #4785

image


packages/frontend/i18n/src/resources/en.json
```diff
@@ -684,7 +684,7 @@
   "com.affine.auth.sign-out.confirm-modal.description": "After signing out, the Cloud Workspaces associated with this account will be removed from the current device, and signing in again will add them back.",
   "com.affine.auth.sign-out.confirm-modal.cancel": "Cancel",
   "com.affine.auth.sign-out.confirm-modal.confirm": "Sign Out",
-  "com.affine.payment.recurring-yearly": "Annually",
+  "com.affine.payment.recurring-yearly": "Yearly",
```
joooye34, line 687: I think the lower case should be default, upper case is special.

packages/frontend/core/src/components/affine/setting-modal/general-setting/plans/index.tsx
```diff
@@ -21,13 +21,17 @@ import * as styles from './style.css';
 const getRecurringLabel = ({
   recurring,
   t,
+  shouldLowerCase,
 }: {
   recurring: SubscriptionRecurring;
   t: ReturnType<typeof useAFFiNEI18N>;
+  shouldLowerCase?: boolean;
```
joooye34, line 28: Use upperFirst from lodash to change this logic, let lower be default.

CatsJuice: Will it be better to implement with css only instead?

joooye34: oh, if css can resolve it, that is better, i don't know it.

joooye34 (approved):

PR #4808, opened 2023-11-01
BlobStorage should provide consistent behaviour

zhe-he:
https://github.com/toeverything/AFFiNE/issues/3245

In order to avoid major changes, I defined a variable in `createSQLiteStorage` to store the type.

The default type of the blob returned in the `createStaticStorage` method is response.headers["content-type"], and I believe it does not need to be changed. If it is necessary to make changes, I list the following for reference.
```
get: async (key: string) => {
    let blob: Blob | null = null;
    if (key.startsWith('/static/')) {
        const response = await fetch(key);
        if (response.ok) blob = response.blob();
    } else if (predefinedStaticFiles.includes(key)) {
        const response = await fetch(`static/${key}`);
        if (response.ok) blob = response.blob();
    }
    if (blob && !blob.type) {
        // Determine the type based on the blob's headers.
        const type = ...;
        blob = new Blob([blob], { type });
    }
    return blob;
},
```

### DIFF
--- a/packages/frontend/workspace/src/blob/sqlite-blob-storage.ts
+++ b/packages/frontend/workspace/src/blob/sqlite-blob-storage.ts
@@ -4,13 +4,15 @@ import type { BlobStorage } from '@blocksuite/store';
 export const createSQLiteStorage = (workspaceId: string): BlobStorage => {
   const apis = window.apis;
   assertExists(apis);
+  const typeMap = new Map();
   return {
     crud: {
       get: async (key: string) => {
         const buffer = await apis.db.getBlob(workspaceId, key);
-        return buffer ? new Blob([buffer]) : null;
+        return buffer ? new Blob([buffer], { type: typeMap.get(key) ?? buffer.type }) : null;
       },
       set: async (key: string, value: Blob) => {
+        typeMap.set(key, value.type);
         await apis.db.addBlob(
           workspaceId,
           key,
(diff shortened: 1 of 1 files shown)


lawvs (changes requested):
The `api.db` is a persistent storage, but the `typeMap` only exists in memory. I think the type of blob will be lost still after reloading the client.

zhe-he:
When I consider storing the type information in localStorage, it seems strange to have different storage methods for the same data. I was considering storing the type in api.db as well, but it may conflict with the user's workspaceId, which is also strange to have two storage methods for the same data - one for storing the content and another for storing the type. The most perfect solution I can think of brings us back to the initial discussion, but it would require significant changes.

I believe the perfect solution would be to modify the file `packages/frontend/electron/src/helper/db/index.ts` and change `addBlob: async (workspaceId: string, key: string, data: Uint8Array)` to `addBlob: async (workspaceId: string, key: string, data: Blob)`. (This change would affect all storage and retrieval operations.)

pengx17:
> When I consider storing the type information in localStorage, it seems strange to have different storage methods for the same data.

Indeed, the current solution for storing blobs in different datasource is not very clean, saying when getting a blob may lose the `type` field in the blob since we only store the binary part of it.
IMO I do agree that we shall not have two storages for data and type and they could coexist in a single table.

An alternative solution is to infer the type from the binary header using libraries like [file-type](https://github.com/sindresorhus/file-type) and thus we do not need to store a separate type field for the blob. This should be fine most of the time since blob is binary, but there could be cases that blob are string types like svg/csv etc.

I am closing the PR since as @lawvs has pointed out your current solution only provide a solution that works in memory which means it will break when app reload.

PR #4812, opened 2023-11-01
fix(core): currentUser undefined when all workspaces deleted

### LINKED ISSUE #4809: Unrecoverable error when all workspaces are deleted​
### What happened?

# Reproduction method

Turn all workspaces into affine cloud and delete all workspaces:

图片

### Distribution version

macOS x64 (Intel)

_No response_

### Relevant log output

```shell
react-dom.production.min.js:189 TypeError: Cannot read properties of undefined (reading 'currentUser')
  at l (use-subscription.ts:12:8)
  at o (use-subscription.ts:39:11)
  at x (user-plan-button.tsx:12:26)
  at ay (react-dom.production.min.js:167:137)
  at s (react-dom.production.min.js:290:337)
  at i_ (react-dom.production.min.js:280:389)
  at react-dom.production.min.js:280:320
  at iT (react-dom.production.min.js:280:325)
  at iC (react-dom.production.min.js:271:88)
  at iZ (react-dom.production.min.js:268:429)
a6 @ react-dom.production.min.js:189
index.js:481 React Router caught the following error during render TypeError: Cannot read properties of undefined (reading 'currentUser')
  at l (use-subscription.ts:12:8)
  at o (use-subscription.ts:39:11)
  at x (user-plan-button.tsx:12:26)
  at ay (react-dom.production.min.js:167:137)
  at s (react-dom.production.min.js:290:337)
  at i_ (react-dom.production.min.js:280:389)
  at react-dom.production.min.js:280:320
  at iT (react-dom.production.min.js:280:325)
  at iC (react-dom.production.min.js:271:88)
  at iZ (react-dom.production.min.js:268:429) Object
```

### Anything else?

_No response_

EYHN:
# Issue Fixed #4809 

error log:

```
react-dom.production.min.js:189 TypeError: Cannot read properties of undefined (reading 'currentUser')
  at l (use-subscription.ts:12:8)
  at o (use-subscription.ts:39:11)
  at x (user-plan-button.tsx:12:26)
  at ay (react-dom.production.min.js:167:137)
  at s (react-dom.production.min.js:290:337)
  at i_ (react-dom.production.min.js:280:389)
  at react-dom.production.min.js:280:320
  at iT (react-dom.production.min.js:280:325)
  at iC (react-dom.production.min.js:271:88)
  at iZ (react-dom.production.min.js:268:429)
a6 @ react-dom.production.min.js:189
index.js:481 React Router caught the following error during render TypeError: Cannot read properties of undefined (reading 'currentUser')
  at l (use-subscription.ts:12:8)
  at o (use-subscription.ts:39:11)
  at x (user-plan-button.tsx:12:26)
  at ay (react-dom.production.min.js:167:137)
  at s (react-dom.production.min.js:290:337)
  at i_ (react-dom.production.min.js:280:389)
  at react-dom.production.min.js:280:320
  at iT (react-dom.production.min.js:280:325)
  at iC (react-dom.production.min.js:271:88)
  at iZ (react-dom.production.min.js:268:429) Object
```

# Cause

Because we used `UserAccountItem` within `pages/index`, and made a GraphQL query using `useQuery`, but there is no `SWRConfig` in this route (actually only the workspace route has `SWRConfig`), the call returned `undefined` incorrectly.

# Solution

It seems reasonable to integrate SWRConfig at a global level.

After all, this is a global level change. 
If this is inappropriate, please tell me.

### DIFF
--- a/packages/frontend/workspace/src/affine/gql.ts
+++ b/packages/frontend/workspace/src/affine/gql.ts
@@ -8,6 +8,7 @@ import type {
 } from '@affine/graphql';
 import { gqlFetcherFactory } from '@affine/graphql';
 import type { GraphQLError } from 'graphql';
+import { useMemo } from 'react';
 import type { Key, SWRConfiguration, SWRResponse } from 'swr';
 import useSWR from 'swr';
 import type {
@@ -70,10 +71,18 @@ export function useQuery<Query extends GraphQLQuery>(
   options: QueryOptions<Query>,
   config?: any
 ) {
+  const configWithSuspense: SWRConfiguration = useMemo(
+    () => ({
+      suspense: true,
+      ...config,
+    }),
+    [config]
+  );
+
   return useSWR(
     () => ['cloud', options.query.id, options.variables],
     () => fetcher(options),
-    config
+    configWithSuspense
   );
 }
 


forehalo:
The thing is that the cloud UI mode has already been with SWR suspense, check [provider](https://github.com/toeverything/AFFiNE/blob/bed9310519499bde6def4d41a345234a9897da9d/packages/frontend/core/src/components/cloud/provider.tsx).

I think the main reason causes the uncaught throw is that after user delete all the workspaces, the UI mode should be automatically switched to local, by which all `useQuery` hooks will be unmounted, but that didn't happen.

EYHN:
@forehalo You are right, the page doesn't switch to local mode because there isn't any workspace available, but that's another topic. 

I looked at other UI Mode（local, cloud...) and they all ended up referencing the same provider and SWR configuration, no reason not to extract it globally.

But that's part of a larger refactor, this PR just fixes the issue.

I will create a new PR later, remove the SWR configuration from UI mode.

packages/frontend/core/src/app.tsx
```diff
@@ -46,22 +48,28 @@ async function loadLanguage() {
 
 const languageLoadingPromise = loadLanguage().catch(console.error);
 
+const globalSWRConfig: SWRConfiguration = {
+  suspense: true,
+};
```
pengx17, line 53: it seems a bit dangerous to me to turn this flag on globally. Can you suggest any other side effects it will bring in?

EYHN:
@forehalo @pengx17 

Sorry, I found a less impactful solution and committed new updates.

I add `suspense: true` to `useQuery`.
Considering that `useQuery` already has `suspense: true` in the type definition.
https://github.com/toeverything/AFFiNE/blob/604c3da9fedf1f4d996a774577c1d63724f4c441/packages/frontend/workspace/src/affine/gql.ts#L49

it should be reasonable to add it.

pengx17 (approved):

Issue #4793, opened 2023-10-31
​Mac German keyboard typing "ß" switches between edgeless and paged mode  


antoniozh:
### What happened?

When I am in the editor and try to type the letter "ß", Affine switches between edgeless and paged mode and does not enter the key instead. I can't seem to find the hotkey in the hotkey settings so I can't disable it either.
This does not happen in the demo online. Is this intended hotkey behaviour and if yes, can there be an option to disable/change it?

### Distribution version

macOS ARM 64 (Apple Silicon)

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

EYHN:
Hi, Zhu. This issue looks interesting. I have investigated the cause of this issue, the code doesn't account for the German keyboard.

and i've submitted a PR to fix this issue.
https://github.com/toeverything/AFFiNE/pull/4813.

lawvs:
Fixed in https://github.com/toeverything/AFFiNE/pull/4813, thanks to the efforts of @EYHN

PR #4813, opened 2023-11-01
fix(core): fix block suite edit mode switch shortcut

EYHN:
# fixed issue https://github.com/toeverything/AFFiNE/issues/4793

After my test, using `⌥ + S` in macos, event.key is `ß`, event.code is `KeyS`

Using KeyboardEvent.code seems to be a better solution.

The relevant code was last changed here:

https://github.com/toeverything/AFFiNE/pull/3097/files#diff-7d9ad95b8b3e47ad3d6edd26562e2553cc03171c58ed65c4c14a38b9c41effa7R48


lawvs (approved):
Good catch!

forehalo (approved):

pengx17:
I think this can be refactored by
- adding a new cmdk command for this action
- setup the keybinding in the command config

PR #4817, opened 2023-11-02
feat(core): remove `mode` and `pages` field from Collection

zzj3720:
1. product requirement changes. remove `mode` field, the mode depends on whether there is a filter
2. remove `pages` field, merge into `allowList`
3. refactor `edit-collection.tsx`, split into multiple files


packages/frontend/core/src/components/pure/workspace-slider-bar/collections/collections-list.tsx
```diff
@@ -58,6 +58,11 @@ const CollectionRenderer = ({
     id: `${Collections_DROP_AREA_PREFIX}${collection.id}`,
     data: {
       addToCollection: (id: string) => {
+        if (collection.allowList.includes(id)) {
+          toast('Page already exists');
```
pengx17, line 62: should skip adding to page when it is included

pengx17 (approved):

PR #4821, opened 2023-11-02
fix(core): remove responsive editor width

### LINKED ISSUE #4524: Change the width of the editing area to `944px`
### What happened?

image

### Distribution version

macOS x64 (Intel)

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

joooye34:
close #4524 

Not sure whether the width for print and small screen should keep. cc @hwangdev97


hwangdev97:
Can’t see any problem at the moment 🌚

pengx17:
Before the change, the editor would add a little more padding on both sides when it is less than 640. After the change, this padding has become smaller.
Is this OK @hwangdev97 ?

pengx17 (approved):

PR #4833, opened 2023-11-03
feat(core): temporary expansion files are limited to 100M

JimmFly:
Temporarily, the file upload limit within the editor has been adjusted to 100MB and the image upload limit for workspace and personal account avatars has been set at under 4K, with a size limit of 10MB.


packages/frontend/hooks/src/use-block-suite-workspace-avatar-url.ts
```diff
@@ -3,6 +3,47 @@ import type { Workspace } from '@blocksuite/store';
...
+    const sizeInMB = file.size / (1024 * 1024);
+    if (sizeInMB > 10) {
+      reject('File is too large, please select a file less than 10MB');
+      return;
+    }
+
+    const img = new Image();
+    const url = URL.createObjectURL(file);
+
+    img.onload = () => {
```
darkskygit, line 18: consider use decode().then directly

joooye34 (approved):

darkskygit (approved):

Issue #4825, opened 2023-11-02
Full text search only works on loaded pages


EYHN:
Currently, the page list will load all pages, so this problem is not obvious.

It can be reproduced by open a page content and reload the app.

Before load:
![image](https://github.com/toeverything/AFFiNE/assets/13579374/e9dfb284-851b-4dda-88aa-aa061852bf3e)

After load:
![image](https://github.com/toeverything/AFFiNE/assets/13579374/ce9249fb-f5cd-4f3c-9de0-103f0f3135cf)

pengx17:
Oops, this is indeed an issue that has not been anticipated.

pengx17:
caused by https://github.com/toeverything/AFFiNE/pull/4775

pengx17:
should have been fixed (workaround) in https://github.com/toeverything/AFFiNE/pull/4834
but we may need a better solution for that

PR #4834, opened 2023-11-03
perf(core): load all pages after 10s

pengx17:
1. use `p-queue` to throttle load events
2. load all pages after 10s

workaround for https://github.com/toeverything/AFFiNE/issues/4825

### DIFF
--- a/packages/frontend/core/src/layouts/workspace-layout.tsx
+++ b/packages/frontend/core/src/layouts/workspace-layout.tsx
@@ -133,7 +133,7 @@ function useLoadWorkspacePages() {
           .map(id => currentWorkspace.blockSuiteWorkspace.getPage(id))
           .filter((p): p is Page => !!p);
         pages.forEach(page => {
-          loadPage(page, 10);
+          loadPage(page, -10);
         });
       }, 10 * 1000); // load pages after 10s
       return () => {
--- a/packages/frontend/core/src/layouts/workspace-layout.tsx
+++ b/packages/frontend/core/src/layouts/workspace-layout.tsx
@@ -133,7 +133,7 @@ function useLoadWorkspacePages() {
           .map(id => currentWorkspace.blockSuiteWorkspace.getPage(id))
           .filter((p): p is Page => !!p);
         pages.forEach(page => {
-          loadPage(page);
+          loadPage(page, 10);
         });
       }, 10 * 1000); // load pages after 10s
       return () => {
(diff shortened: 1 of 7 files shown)


packages/frontend/core/src/layouts/workspace-layout.tsx
```diff
@@ -119,6 +121,29 @@ type WorkspaceLayoutProps = {
...
+    if (currentWorkspace) {
+      const timer = setTimeout(() => {
+        const pageIds = pageMetas.map(meta => meta.id);
+        const pages = pageIds
+          .map(id => currentWorkspace.blockSuiteWorkspace.getPage(id))
+          .filter((p): p is Page => !!p);
+        pages.forEach(page => {
+          loadPage(page);
+        });
+      }, 10 * 1000); // load pages after 10s
```
darkskygit, line 138: consider using `requestIdleCallback` to let the browser automatically defer requests?

pengx17: done

darkskygit (approved):

packages/frontend/hooks/src/use-block-suite-workspace-page.ts
```diff
@@ -55,8 +94,10 @@ export function useBlockSuiteWorkspacePage(
   const page = useAtomValue(pageAtom);
 
   useEffect(() => {
-    if (!page?.loaded) {
-      page?.waitForLoaded().catch(console.error);
+    if (page && !page.loaded) {
+      loadPage(page).catch(err => {
```
EYHN, line 98: I think you can use p-queue.add [priority](https://github.com/sindresorhus/p-queue#priority) option, jump into the queue and load the page here as quickly as possible, and reducing interference from automatic page loading.

pengx17: Great suggestion!

packages/frontend/core/src/layouts/workspace-layout.tsx
```diff
@@ -133,7 +133,7 @@ function useLoadWorkspacePages() {
           .map(id => currentWorkspace.blockSuiteWorkspace.getPage(id))
           .filter((p): p is Page => !!p);
         pages.forEach(page => {
-          loadPage(page);
+          loadPage(page, 10);
```
EYHN, line 136: 😊 Hey, i think here should be a small number, like `-10`, because according to the p-queue documentation `Operations with greater priority will be scheduled first.`

Here are lazy load pages so they should have a smaller priority

pengx17: ops, I thought smaller priority number means higher priority

EYHN (approved):

PR #4840, opened 2023-11-05
feat(core): add global loading state

fourdim:


forehalo:
how does it look like?

fourdim:
According to the design in figma, it look like this:
image

fourdim:
I would like to update this commit, it looks like it has a race condition here.

joooye34 (approved):

joooye34 (approved):

PR #4845, opened 2023-11-06
fix(core): adapt blob in sqlite for svg type

pengx17:
SVG `blob`'s type will be missing after it is being saved into data sources because it is essentially a string. This is not an issue for image because image have headers in their binary format.
For idb we have mime table that stores the mime type info, but I think it may not be that necessary since the only counter case for storing blob is svg.

### DIFF
--- a/packages/frontend/workspace/package.json
+++ b/packages/frontend/workspace/package.json
@@ -23,6 +23,7 @@
     "@toeverything/hooks": "workspace:*",
     "@toeverything/y-indexeddb": "workspace:*",
     "async-call-rpc": "^6.3.1",
+    "is-svg": "^5.0.0",
     "jotai": "^2.4.3",
     "js-base64": "^3.7.5",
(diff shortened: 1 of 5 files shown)


doodlewind (approved):

lawvs:
In addition, the missing blob's type also causes https://github.com/toeverything/blocksuite/issues/5097. Furthermore, I believe it will block the attachment preview feature in the future.

pengx17:
> In addition, the missing blob's type also causes https://github.com/toeverything/blocksuite/issues/5097. Furthermore, I believe it will block the attachment preview feature in the future.

I see, it seems that the current fix does not cover the case for embedded attachment because it did use the mime-type from blob.type but it is empty for images.
I tried to use `file-type` to infer mime from image, but it seems it does not work well on browser yet, see
https://github.com/sindresorhus/file-type/issues/578

PR #4849, opened 2023-11-06
docs: update behind-the-code.md

singhjatin297:
typo, it was "if there is only have `affine-cloud`," but it should be "if there is only `affine-cloud`,"


joooye34:
@singhjatin297 thanks for your contribution. The "license/cla" should be signed and I will merge this pr.

joooye34 (approved):

PR #4858, opened 2023-11-07
feat(core): auto select block when jump to block

JimmFly:
https://github.com/toeverything/AFFiNE/assets/102217452/a905b3a5-0fa5-4e35-8bd0-aa81cf122e40


packages/frontend/component/src/components/block-suite-editor/index.tsx
```diff
@@ -77,14 +77,22 @@ const BlockSuiteEditorImpl = (props: EditorProps): ReactElement => {
       if (editor.page && onLoad) {
         disposes.push(onLoad?.(page, editor));
       }
+      if (props.path?.length) {
+        const selectPath = props.path;
+        const selectManager = editor.root.value?.selection;
+        if (selectManager) {
+          const path = selectManager.getInstance('block', { path: selectPath });
```
Flrande, line 84: Use `selection` instead of `path`

packages/frontend/component/src/components/block-suite-editor/index.tsx
```diff
@@ -129,6 +129,21 @@ const BlockSuiteEditorImpl = (props: EditorProps): ReactElement => {
     };
   }, [editor, page.awarenessStore, page.meta.trash, setBlockHub]);
 
+  useEffect(() => {
+    if (props.path?.length) {
+      const selectPath = props.path;
+      const selectManager = editor.root.value?.selection;
+      if (selectManager) {
+        setTimeout(() => {
```
Flrande, line 137: After https://github.com/toeverything/blocksuite/pull/5237 merged we will no need to use `setTimeout`.

```ts
useEffect(() => {
    if (props.path?.length) {
      const selectPath = props.path;
      editor.updateComplete
        .then(() => {
          const selectManager = editor.root.value?.selection;
          assertExists(selectManager);
          const path = selectManager.getInstance('block', { path: selectPath });
          selectManager.set([path]);
        })
        .catch(err => console.error(err));
    }
  }, [editor, props.path]);
```

JimmFly: For unknown reasons this does not work, this will be fixed in the future.

packages/frontend/core/src/components/page-detail-editor.tsx
```diff
@@ -125,6 +103,25 @@ const EditorWrapper = memo(function EditorWrapper({
...
+      setTimeout(() => {
+        blockElement.scrollIntoView({
+          behavior: 'smooth',
+          block: 'center',
+          inline: 'center',
+        });
+        setElementPath(blockElement.path);
+      }, 0);
+    }
+  }, [blockElement]);
```
pengx17, line 123: block scroll to & selection code shall be put into the same `useEffect` scope in a single component. maybe you can move this into `<BlockSuiteEditor />`

packages/frontend/core/src/components/page-detail-editor.tsx
```diff
@@ -125,6 +103,25 @@ const EditorWrapper = memo(function EditorWrapper({
     [switchToEdgelessMode, switchToPageMode, pageId]
   );
 
+  //hack: scroll to block when jumping to block
+  const [elementPath, setElementPath] = useState<string[]>([]);
+  const blockElement = document.querySelector(
+    `[data-block-id="${blockId}"]`
+  ) as BlockElement | null;
```
pengx17, line 110: do not query dom element in render

JimmFly: Sorry, I haven't found a way to make this function run normally without writing it like this.

packages/frontend/core/src/components/page-detail-editor.tsx
JimmFly: I will try to do it, but it may not work successfully

packages/frontend/core/src/components/page-detail-editor.tsx
pengx17: You can query it in `useEffect` when needed

packages/frontend/core/src/components/page-detail-editor.tsx
pengx17: I can take a look if it does not work

packages/frontend/component/src/components/block-suite-editor/index.tsx
```diff
@@ -129,6 +144,42 @@ const BlockSuiteEditorImpl = (props: EditorProps): ReactElement => {
...
+
+  if (!blockElement && blockId) {
+    const element = document.querySelector(
+      `[data-block-id="${blockId}"]`
+    ) as BlockElement | null;
+
+    if (element) {
+      setBlockElement(element);
+    }
+  }
```
pengx17, line 161: This is effectively a while loop in every render, until blockElement is being set.

JimmFly: Could you please tell me how to optimise?

pengx17: I am rewriting this component. The basic idea is to iterate for a few ms until the block element is found ... I guess this is not ideal either, but should be better than relying on render to set the blockElement

packages/frontend/component/src/components/block-suite-editor/index.tsx
```diff
@@ -30,28 +42,62 @@ export type ErrorBoundaryProps = {
...
+// a workaround for returning the webcomponent for the given block id
+// by iterating over the children of the rendered dom tree
+const useBlockElementById = (
+  container: HTMLElement | null,
+  blockId: string | undefined,
+  timeout = 1000
+) => {
+  const [blockElement, setBlockElement] = useState<BlockElement | null>(null);
+  useEffect(() => {
+    if (!blockId || !container) {
```
joooye34, line 54: what is the meanings of container

packages/frontend/component/src/components/block-suite-editor/index.tsx
```diff
@@ -64,36 +110,29 @@ const BlockSuiteEditorImpl = (props: EditorProps): ReactElement => {
...
         onModeChange?.(mode);
       });
-
-      disposes.push(() => dispose?.dispose());
-
-      if (editor.page && onLoad) {
-        disposes.push(onLoad?.(page, editor));
+      disposes.push(() => disposeModeSwitch?.dispose());
+      if (onLoadEditor) {
+        disposes.push(onLoadEditor?.(editor));
```
joooye34, line 121: why there have a optional question in a if condition

packages/frontend/component/src/components/block-suite-editor/index.tsx
JimmFly: /cc @pengx17

pengx17: ops, it should be used on line 63.

joooye34 (approved):

Issue #11, opened 2022-08-01
When switching views, Kanban inner state lose.


mitsuhatu:
After dragging the card exchange order in the Kanban view, switch to other views, and the Kanban status is lost.

PR #4863, opened 2023-11-07
fix(hooks): missing page preview and references

EYHN:
status: This pr fix [#11](https://github.com/toeverything/AFFiNE/issues/4865) , would be able to improve if blocksuite was able to provide page.onSync slat https://github.com/toeverything/blocksuite/issues/5249.

image

### DIFF
--- a/packages/frontend/hooks/src/use-block-suite-page-preview.ts
+++ b/packages/frontend/hooks/src/use-block-suite-page-preview.ts
@@ -25,14 +25,19 @@ export function useBlockSuitePagePreview(page: Page | null): Atom<string> {
   } else if (weakMap.has(page)) {
     return weakMap.get(page) as Atom<string>;
   } else {
-    const baseAtom = atom<string>(getPagePreviewText(page));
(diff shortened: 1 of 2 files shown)


pengx17 (dismissed):

packages/frontend/hooks/src/use-block-suite-page-preview.ts
```diff
@@ -25,14 +25,19 @@ export function useBlockSuitePagePreview(page: Page | null): Atom<string> {
   } else if (weakMap.has(page)) {
     return weakMap.get(page) as Atom<string>;
   } else {
-    const baseAtom = atom<string>(getPagePreviewText(page));
+    const baseAtom = atom<string>('');
```
pengx17, line 28: This is removed because `ready` slot will always be called?

EYHN: It will eventually set in `onMount`

pengx17 (approved):

PR #4867, opened 2023-11-08
ci: add oxlint

EYHN:


.github/workflows/build.yml
```diff
@@ -60,6 +60,19 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Node.js
+        uses: ./.github/actions/setup-node
+        with:
+          electron-install: false
+      - name: Run oxlint
+        run: npx --yes oxlint@latest .
```
Boshen, line 74: I don't use yarn, but is there a `yarnx`?

You should put `oxlint` before everything else because this is the fastest to run.

Brooooooklyn:
Can we disable all rules in eslint that are covered by oxc?

Boshen:
> Can we disable all rules in eslint that are covered by oxc?

I need to create an eslint plugin for this, which we don't have one yet.

The rules oxlint uses are pretty performant in ESLint so performance is not big of a concern. 

The benefit of adding oxlint is mostly for better feedback loop, because it is fast.

.github/workflows/build.yml
```diff
@@ -42,6 +42,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      - name: Run oxlint
+        # oxlint is fast, so wrong code will fail quickly
+        run: npx --yes oxlint@latest .
```
Brooooooklyn, line 47: Use `yarn dlx` because in this way the downloaded `oxc` package can be cached by `actions/setup-node`

EYHN: done

EYHN:
🚀
image

Brooooooklyn (approved):
🚀

PR #4882, opened 2023-11-09
fix(core): visit /signin pay when already logged and subscribed

joooye34:
1. redirect to the payment URL when the user has logged in
2. show the subscription existing page when the user has subscribed


packages/frontend/i18n/src/resources/en.json
```diff
@@ -766,6 +766,7 @@
   "com.affine.payment.upgrade-success-page.title": "Upgrade Successful!",
   "com.affine.payment.upgrade-success-page.text": "Congratulations! Your AFFiNE account has been successfully upgraded to a Pro account.",
   "com.affine.payment.upgrade-success-page.support": "If you have any questions, please contact our <1> customer support</1>.",
+  "com.affine.payment.subscription.exist": "You have subscription already.",
```
pengx17, line 769:
```suggestion
  "com.affine.payment.subscription.exist": "You already have a subscription.",
```

?

pengx17 (approved):

Issue #4861, opened 2023-11-07
SVG cannot be displayed in collaborators


Yuhang1015:
![Image](https://github.com/toeverything/blocksuite/assets/142964904/c203ec9e-2b40-40ec-9148-da412996f9ab)

version：0.10.0 beta internal 11.7
There may be no stable recurrence pattern due to different computer versions.

PR #4886, opened 2023-11-09
fix(core): svg blob syncing issue

pengx17:
fix https://github.com/toeverything/AFFiNE/issues/4861


packages/frontend/workspace/src/blob/cloud-blob-storage.ts
```diff
@@ -17,15 +18,15 @@ export const createCloudBlobStorage = (workspaceId: string): BlobStorage => {
...
           runtimeConfig.serverUrlPrefix + suffix
-        ).then(res => {
+        ).then(async res => {
           if (!res.ok) {
             // status not in the range 200-299
             return null;
           }
-          // todo: shall we add svg type here if it is missing?
-          return res.blob();
+          return bufferToBlob(await res.arrayBuffer());
```
Brooooooklyn, line 29: Do you think it's worth to add a smoke test for this case?

pengx17: Sure, let me add one

pengx17: Done

joooye34 (approved):

PR #4900, opened 2023-11-09
refactor: new provider

EYHN:
## provider

The `Provider` is responsible for sync `Y.Doc` with the local database and the Affine Cloud, serving as the source of Affine's local-first collaborative magic.

When Affine boot, the `Provider` is tasked with reading content from the local database and loading it into the workspace, continuously storing any changes made by the user into the local database.

When using Affine Cloud, the `Provider` also handles sync content with the Cloud.

Additionally, the `Provider` is responsible for implementing a local-first capability, allowing users to edit offline with changes stored in the local database and sync with the Cloud when the network is restored.

## Provider's Issues

Previously, the provider was split into many implementations based on the storage:

- IndexedDBBackgroundProvider
- AffineSocketIOProvider
- BroadcastChannelProvider
- …

Most of these providers' capabilities are provided by the `LazyProvider` in the `y-provider` package. There are several issues with the implementation.

**Synchronization Depends on Management from the React Side**

Currently, there is a lot of code for handling synchronization on both the `React` and `Jotai` sides, making it difficult to manage.

**Disrespect for Update Order**

Yjs relies on updates being applied in the order they were generated, but LazyProvider cannot guarantee this.

**Not Usable Offline**

The current mode does not ensure that the local database contains a usable copy of the data.

## Goals

In this PR, I have refactor the provider to achieve the following goals:

- **Offline viewability and editability.**
- **Local-first data loading to increase loading speed and reduce the failure rate of opening documents.**
- **Correctness of the collaboration synchronization sequence.**
- **Automatic synchronization, relieving the React side from concerns about sync.**

## New provider

The new provider is divided into three parts:

**SyncEngine**

Manages one local storage and multiple remote storages.

Responsible for creating SyncPeers for synchronization, following the local-first strategy.

**SyncPeer**

Responsible for synchronizing a single storage with Y.Doc.

Carries the main synchronization logic.

**Storage**

A unified interface for storage, with three methods: push, pull, and subscribe.

## Done

- [x] Implemented the core synchronization logic (by SyncEngine and SyncPeer)
- [x] Transitioned to new provider. (It works ✅)
- [x] Exception handling and retry mechanism for SyncEngine and SyncPeer.
- [x] Storage backend：Sqlite, IndexedDB, Affine Socket.io
- [x] Sync between multiple browser tabs
- [x] Refactor Awareness Provider
- [x] The sync button cannot be clicked and only be used to display the sync status.
- [x] Cleanup code

## Remaining issues (not for this PR)

- local peer need to prevent page reload when these queues have pending items
- i18n for Sync Status text
- workspace need 'disconnect' when switch to other workspace
- workspace title becomes Untitled in first load, going well in later loads
- refactor AsyncQueue to PriorityQueue, prioritize loading of current page

### <samp>🤖 Generated by Copilot at f97f9af</samp>

This pull request refactors the workspace provider logic and the workspace upgrade logic to use the `SyncEngine` abstraction and the `downloadBinaryFromCloud` function. It also improves the UI and performance of the workspace status and the workspace pages. It removes unused and deprecated code and files related to the `IndexedDBDownloadProvider` and the affine data source. It adds new files and functions for the awareness provider logic that uses the socket.io and the broadcast channel APIs.


packages/frontend/workspace/src/new/index.ts
```diff
@@ -0,0 +1,399 @@
...
+      console.error(err);
+    });
+  }
+
+  stop() {
+    this.abort.abort('manually-stop');
+    this.isStarted = false;
+  }
+
+  async sync(abort: AbortSignal) {
```
forehalo, line 130: this function is too big

packages/frontend/core/src/components/pure/workspace-slider-bar/workspace-card/index.tsx
```diff
@@ -100,49 +125,30 @@ const WorkspaceStatus = ({
     if (!isOnline) {
       return 'Disconnected, please check your network connection';
     }
-    switch (forceSyncStatus.type) {
+    switch (syncEngineStatus) {
```
forehalo, line 128: add a todo: i18n

packages/frontend/core/src/components/pure/workspace-slider-bar/workspace-card/index.tsx
```diff
@@ -86,12 +85,38 @@ const WorkspaceStatus = ({
...
+  )?.engine;
+
+  useEffect(() => {
+    const toStatus = (status?: SyncEngineStatus) => {
+      switch (status) {
+        case SyncEngineStatus.Synced:
+          return 'synced';
+        case SyncEngineStatus.Syncing:
+          return 'syncing';
+        case SyncEngineStatus.Retry:
```
forehalo, line 104: `Retring`, aligned with `Syncing`

packages/frontend/core/src/components/pure/workspace-slider-bar/workspace-card/index.tsx
```diff
@@ -86,12 +85,38 @@ const WorkspaceStatus = ({
...
-  const [forceSyncStatus, startForceSync] = useDatasourceSync(
-    currentWorkspace.blockSuiteWorkspace
-  );
-
-  const setIsHovered = useSetAtom(hoverAtom);
+  const [syncEngineStatus, setSyncEngineStatus] = useState<
+    'syncing' | 'synced' | 'error'
+  >('syncing');
+
+  // FIXME: This is a hack to get the sync engine, we need refactor this in the future.
```
forehalo, line 92: does it make sense that we make sync engine a workspace space level state, along side with `useCurrentWorkspace()`?

packages/frontend/core/src/components/pure/workspace-slider-bar/workspace-card/index.tsx
```diff
@@ -100,49 +125,30 @@ const WorkspaceStatus = ({
...
     <div style={{ display: 'flex' }}>
       <Tooltip content={content}>
-        <StyledWorkspaceStatus
-          onMouseEnter={() => {
-            setIsHovered(true);
-          }}
-          onMouseLeave={() => setIsHovered(false)}
-          onClick={handleClick}
-        >
+        <StyledWorkspaceStatus>
```
forehalo, line 151: now we don't allow user to manually start sync progress, right?

packages/frontend/core/src/components/workspace-upgrade/upgrade-hooks.ts
```diff
@@ -17,69 +10,24 @@ export function useUpgradeWorkspace() {
...
-    assertExists(provider, 'no local provider');
-    return provider;
-  }, [providers]);
 
   const upgradeWorkspace = useCallback(() => {
     setState('upgrading');
     setError(null);
 
     (async () => {
-      await syncDocFromDataSource(
```
forehalo, line 45: ur surperman![beautiful](https://app.graphite.dev/macros/toeverything)

packages/frontend/workspace/src/affine/download.ts
```diff
@@ -0,0 +1,35 @@
...
+  pageGuid: string
+): Promise<CloudDoc | boolean> {
+  if (hashMap.has(`${rootGuid}/${pageGuid}`)) {
+    return true;
+  }
+  const response = await fetchWithTraceReport(
+    runtimeConfig.serverUrlPrefix +
+      `/api/workspaces/${rootGuid}/docs/${pageGuid}`,
+    {
+      priority: 'high',
```
forehalo, line 22: [bow](https://app.graphite.dev/macros/toeverything)

packages/frontend/workspace/src/affine/download.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (response.ok) {
+    const publishMode = (response.headers.get('publish-mode') ||
+      'page') as DocPublishMode;
+    const arrayBuffer = await response.arrayBuffer();
+    hashMap.set(`${rootGuid}/${pageGuid}`, arrayBuffer);
+
+    // return both arrayBuffer and publish mode
+    return { arrayBuffer, publishMode };
+  }
+  return false;
```
forehalo, line 34: no false, null please

packages/frontend/workspace/src/providers/awareness/broadcast-channel/index.ts
```diff
@@ -0,0 +1,63 @@
...
+} from 'y-protocols/awareness.js';
+
+import type { AwarenessProvider } from '..';
+import type { AwarenessChanges } from '../affine';
+
+type ChannelMessage =
+  | { type: 'connect' }
+  | { type: 'update'; update: Uint8Array };
+
+export function createBroadcastChannelAwarenessProvider(
```
forehalo, line 14: will this make doc receives duplicated awareness update from both cloud and other tabs in same channel?

packages/frontend/workspace/src/providers/index.ts
```diff
@@ -1,160 +1,114 @@
...
-  }
-
-  if (environment.isDesktop && runtimeConfig.enableSQLiteProvider) {
-    providers.push(createSQLiteProvider, createSQLiteDBDownloadProvider);
-  }
-
-  return providers;
+  return [
+    (_, doc, { awareness }) => {
+      const engine = new SyncEngine(
```
forehalo, line 30: I prefer moving `Doc` as first parameter of constructor of `SyncEngine` because the doc is the first thing that makes sync happen

packages/frontend/workspace/src/providers/index.ts
```diff
@@ -1,160 +1,114 @@
...
+      );
+      engine.start();
+
+      const awarenessProviders = [
+        createBroadcastChannelAwarenessProvider(doc.guid, awareness),
+      ];
+
+      for (const provider of awarenessProviders) {
+        provider.connect();
+      }
```
forehalo, line 45: should we delay the `starting` after `connect`?

packages/frontend/workspace/src/providers/index.ts
```diff
@@ -1,160 +1,114 @@
...
+
+      let connected = false;
+
+      return {
+        flavour: '_',
+        passive: true,
+        connect() {
+          connected = true;
+          // TODO: actually connect
```
forehalo, line 54: delay the start to here?

packages/frontend/workspace/src/providers/storage/affine/index.ts
```diff
@@ -0,0 +1,162 @@
...
+        workspaceId,
+        guid: docId,
+        update,
+      });
+
+      await syncSender.send(docId, update);
+    },
+    async subscribe(cb, disconnect) {
+      const response: { error?: any } = await socket
+        .timeout(10000)
```
forehalo, line 127: dude

packages/frontend/workspace/src/providers/storage/index.ts
```diff
@@ -0,0 +1,29 @@
...
+   * Subscribe to updates from peer
+   *
+   * @param cb callback to handle updates
+   * @param disconnect callback to handle disconnect, reason can be something like 'network-error'
+   *
+   * @returns unsubscribe function
+   */
+  subscribe(
+    cb: (docId: string, data: Uint8Array) => void,
+    disconnect: (reason: string) => void
```
forehalo, line 23: `onDisconnected`?

packages/frontend/workspace/src/providers/storage/indexeddb/index.ts
```diff
@@ -0,0 +1,131 @@
...
+} from 'yjs';
+
+import type { Storage } from '..';
+
+export const dbVersion = 1;
+export const DEFAULT_DB_NAME = 'affine-local';
+
+export function mergeUpdates(updates: Uint8Array[]) {
+  const doc = new Doc();
+  updates.forEach(update => {
```
forehalo, line 17:
```js
transact(doc, () => {
  // apply updates
})
```

This will make the progress more efficient

packages/frontend/core/src/components/pure/workspace-slider-bar/workspace-card/index.tsx
EYHN: right

packages/frontend/workspace/src/providers/awareness/broadcast-channel/index.ts
EYHN: no, because both broadcast and cloud mark received updates as `origin:'remote'`

packages/frontend/workspace/src/providers/storage/index.ts
EYHN: disconnect is only something you need to care after subscribe, that is the storage tell upper "we may have lost some remote updates".

packages/frontend/workspace/src/providers/storage/affine/index.ts
EYHN: I just copy and paste it here..

EYHN: shorter ?

packages/frontend/workspace/src/providers/index.ts
EYHN: Maybe in the future? currently we control `connect` and `disconnect` in jotai, but some use workspace outside jotai. I want to refactor this part in the future.

packages/frontend/workspace/src/providers/storage/index.ts
forehalo: that's not what onDisconnected does?

packages/frontend/workspace/src/providers/index.ts
forehalo: if so, would it be better we expose the `SyncEngine` as app level state as well, and let the owner to control the connect/disconnect/start?

packages/frontend/workspace/src/providers/index.ts
```diff
@@ -1,160 +1,114 @@
...
+      const engine = new SyncEngine(
+        doc,
+        environment.isDesktop
+          ? createSQLiteStorage(doc.guid)
+          : createIndexedDBStorage(doc.guid),
+        []
+      );
+      engine.start();
+
+      const awarenessProviders = [
```
forehalo, line 39: How about a awareness manager?
Make code clean

packages/frontend/workspace/src/providers/sync/engine.ts
```diff
@@ -0,0 +1,161 @@
...
+
+export class SyncEngine {
+  get rootDocId() {
+    return this.rootDoc.guid;
+  }
+
+  logger = new DebugLogger('affine:sync-engine:' + this.rootDocId);
+  private _status = SyncEngineStatus.Stopped;
+  onStatusChange = new Slot<SyncEngineStatus>();
+  set status(s: SyncEngineStatus) {
```
forehalo, line 50: Should we make it private?

packages/frontend/workspace/src/providers/sync/engine.ts
```diff
@@ -0,0 +1,161 @@
...
+  start() {
+    if (this.status !== SyncEngineStatus.Stopped) {
+      this.stop();
+    }
+    this.abort = new AbortController();
+
+    this.status = SyncEngineStatus.Syncing;
+    this.sync(this.abort.signal).catch(err => {
+      // should never reach here
+      console.error(err);
```
forehalo, line 79: this.logger.error

packages/frontend/workspace/src/providers/sync/engine.ts
```diff
@@ -0,0 +1,161 @@
...
+      for (const remotePeer of remotePeers) {
+        remotePeer.stop();
+      }
+      for (const clean of cleanUp) {
+        clean();
+      }
+    }
+  }
+
+  updateSyncingState(...peers: SyncPeer[]) {
```
forehalo, line 145: No need to use rest parameters

packages/frontend/workspace/src/providers/sync/peer.ts
```diff
@@ -0,0 +1,379 @@
...
+ *       │               │ onReady()        │
+ *    ┌──▼──┐      ┌─────▼───────┐       ┌──▼──┐
+ *    │queue├──────►apply updates◄───────┤queue│
+ *    └─────┘      └─────────────┘       └─────┘
+ *
+ * listen: listen for updates from ydoc, typically from user modifications.
+ * subscribe: listen for updates from storage, typically from other users.
+ *
+ */
+export enum SyncPeerStatus {
```
forehalo, line 34: Can we make `SyncPeerStatus` a strings enumeration for better debugging experience? So as status of `SyncEngine`

packages/frontend/workspace/src/providers/sync/peer.ts
```diff
@@ -0,0 +1,379 @@
...
+    this.status = SyncPeerStatus.LoadingRootDoc;
+    this.abort = new AbortController();
+
+    this.syncRetryLoop(this.abort.signal).catch(err => {
+      // should not reach here
+      console.error(err);
+    });
+  }
+
+  set status(s: SyncPeerStatus) {
```
forehalo, line 67: Private?

packages/frontend/workspace/src/providers/sync/peer.ts
```diff
@@ -0,0 +1,379 @@
...
+      await this.connectDoc(this.rootDoc, abortInner.signal);
+
+      this.status = SyncPeerStatus.LoadingSubDoc;
+
+      // Step 2: load subdocs
+      this.state.subdocsLoadQueue.push(
+        ...Array.from(this.rootDoc.getSubdocs())
+      );
+
+      this.rootDoc.on('subdocs', this.handleSubdocsUpdate);
```
forehalo, line 183: I think it should be put before sync root doc? Will the subdocs event be fired after that?

packages/frontend/workspace/src/providers/index.ts
EYHN: maybe next time

packages/frontend/workspace/src/providers/index.ts
EYHN: I just put the connection part into the sync() function and it magically worked. and only connect to the workspace when needed

packages/frontend/workspace/src/providers/sync/peer.ts
EYHN: After sync root doc, user can create new pages, and also possible create new subdocs by remotes.

We get all subdocs here and listen for new subdocs after, We can guarantee that all subdocs are in the load queue.

packages/frontend/workspace/src/providers/sync/peer.ts
EYHN: But number can be compared

like
```js
status > SyncPeerStatus.Loaded
```

packages/frontend/workspace/src/providers/storage/index.ts
EYHN: There does seem to be some confusion here.

For the `socketio` storage `onDisconnect` and `disconnect` callback are the same.

I consider a **stateless** storage like
`pull` and `push` are using a single http request
`subscribe` will initial a SSE(Server-Sent Events) stream transport updates
this SSE stream only exists in the current `subscribe` context and feeds back disconnect events through `disconnect` callback.

to avoid confusion,it might be better to do this?
```js
subscribe(
    cb: (event: {docId: string, data: Uint8Array} | { disconnect: true }) => void,
)
```

packages/frontend/core/src/components/pure/workspace-slider-bar/workspace-card/index.tsx
```diff
@@ -86,63 +86,68 @@ const WorkspaceStatus = ({
...
     if (!isOnline) {
       return 'Disconnected, please check your network connection';
     }
-    switch (forceSyncStatus.type) {
+    switch (syncEngineStatus) {
       case 'syncing':
         return 'Syncing with AFFiNE Cloud';
       case 'error':
-        return 'Sync failed due to server issues, please try again later.';
+        return 'Sync disconnected due to server issues, reconnecting.';
```
forehalo, line 131: the peer may not always be `server`, need to update the words

packages/frontend/core/src/components/pure/workspace-slider-bar/workspace-card/index.tsx
```diff
@@ -86,63 +86,68 @@ const WorkspaceStatus = ({
...
-
-  const setIsHovered = useSetAtom(hoverAtom);
+  const [syncEngineStatus, setSyncEngineStatus] = useState<
+    'syncing' | 'synced' | 'error'
+  >('syncing');
+
+  const syncEngine = useCurrentSyncEngine();
+
+  useEffect(() => {
+    const toStatus = (status?: SyncEngineStatus) => {
```
forehalo, line 96: btw, do we really need to convert `SyncEngineStatus` to strings here? seems the only difference is change `Retrying` to `error` but it will never shown to user. could we just keep it as it is?

packages/frontend/core/src/hooks/current/use-current-sync-engine.ts
```diff
@@ -0,0 +1,13 @@
+import type { SyncEngine } from '@affine/workspace/providers';
+
+import { useCurrentWorkspace } from './use-current-workspace';
+
+export function useCurrentSyncEngine(): SyncEngine | undefined {
+  const [workspace] = useCurrentWorkspace();
+  // FIXME: This is a hack to get the sync engine, we need refactor this in the future.
+  const syncEngine = (
+    workspace.blockSuiteWorkspace.providers[0] as { engine?: SyncEngine }
```
forehalo, line 9: the provider must exists right?

packages/frontend/workspace/src/providers/storage/index.ts
forehalo: never mind, I think the original signature is cool after seeing such
```ts
subscribe(
    cb: (event: {docId: string, data: Uint8Array} | { disconnect: true }) => void,
)
```

packages/frontend/workspace/src/providers/sync/peer.ts
```diff
@@ -0,0 +1,379 @@
...
+            const subdoc = this.state.connectedDocs.get(docId);
+            if (subdoc) {
+              applyUpdate(subdoc, data, this.name);
+            }
+          }
+        })(),
+        // push updates
+        (async () => {
+          while (throwIfAborted(abortInner.signal)) {
+            const { docId, data } = await this.state.pushUpdatesQueue.next(
```
forehalo, line 225: we gonna need to filter out the update with empty uint8Array or `uint8Array([0, 0])`, which is empty doc serialization result

forehalo: image
batch of useless updates sended

packages/frontend/workspace/src/providers/awareness/affine/index.ts
```diff
@@ -0,0 +1,101 @@
...
+      socket.on('new-client-awareness-init', newClientAwarenessInitHandler);
+      awareness.on('update', awarenessUpdate);
+
+      window.addEventListener('beforeunload', windowBeforeUnloadHandler);
+
+      socket.emit('awareness-init', workspaceId);
+      socket.connect();
+    },
+    disconnect: () => {
+      awareness.off('update', awarenessUpdate);
```
forehalo, line 94: seems it will never be called if I switch workspace. I can still get remote awareness broadcast from old workspace after switching

EYHN: Yes, the workspace currently never disconnect

forehalo: that sounds not good

packages/frontend/core/src/hooks/current/use-current-sync-engine.ts
EYHN: Affine public workspace(shared page) currently does not have a SyncEngine

packages/frontend/workspace/src/providers/awareness/affine/index.ts
EYHN: This is also the case before this pr.

packages/frontend/core/src/hooks/current/use-current-sync-engine.ts
forehalo: we could make a standalone "sync" peer for it, with pull `download` and push do nothing

packages/frontend/workspace/src/providers/awareness/affine/index.ts
forehalo: the connection definitely will be closed after switch workspace, if there is only one socketio instance, at least the 'leave-workspace' event will be sent to server that indicate the session do not want to receive message from old workspace room any more

pengx17: I think we can fix this after this PR get merged.

packages/frontend/workspace/src/providers/index.ts
```diff
@@ -1,160 +1,129 @@
...
-      runtimeConfig.enableCloud && createMergeCloudSnapshotProvider,
-    ] as DocProviderCreator[]
-  ).filter(v => Boolean(v));
+  return [
+    (_, doc, { awareness }) => {
+      const engine = new SyncEngine(
+        doc,
+        environment.isDesktop
+          ? createSQLiteStorage(doc.guid)
+          : createIndexedDBStorage(doc.guid),
```
pengx17, line 83: Shall we treat idb as a remote storage if it is desktop?

packages/frontend/workspace/src/providers/sync/peer.ts
```diff
@@ -0,0 +1,386 @@
...
+ *    ┌──▼──┐      ┌─────▼───────┐       ┌──▼──┐
+ *    │queue├──────►apply updates◄───────┤queue│
+ *    └─────┘      └─────────────┘       └─────┘
+ *
+ * listen: listen for updates from ydoc, typically from user modifications.
+ * subscribe: listen for updates from storage, typically from other users.
+ *
+ */
+export enum SyncPeerStatus {
+  Stopped = 0,
```
pengx17, line 35: Is it possible that we make status more fine-grained? For example, give more observability on how many docs are being syncing (to calculate the syncing progress etc)

packages/frontend/workspace/src/providers/index.ts
```diff
@@ -1,160 +1,129 @@
...
-  return (
-    [
-      ...createLocalProviders(),
-      runtimeConfig.enableCloud && createAffineSocketIOProvider,
-      runtimeConfig.enableCloud && createMergeCloudSnapshotProvider,
-    ] as DocProviderCreator[]
-  ).filter(v => Boolean(v));
+  return [
+    (_, doc, { awareness }) => {
+      const engine = new SyncEngine(
```
forehalo, line 79: I meet issues of opening a cloud workspace without data persisted

- workspace title becomes `Untitled` in first load, going well in later loads
- `Upgrade workspace data` shown in first load, going well in later loads

These are exactly the issues I mentioned by `switching local` feature in design

packages/frontend/workspace/src/providers/index.ts
```diff
@@ -1,160 +1,129 @@
...
+              provider.connect();
+            }
+            connected = true;
+          }
+        },
+        get whenReady() {
+          console.log('!!!!!!!!!!!!!!!');
+          return engine.waitForLoadedRootDoc();
+        },
+        connect() {
```
forehalo, line 112: looks like we don't need the provider to be PassiveProvider

packages/frontend/workspace/src/providers/sync/peer.ts
forehalo: year, sure. we gonna need a status map. I have the consideration in design

packages/frontend/workspace/src/providers/sync/peer.ts
```diff
@@ -0,0 +1,386 @@
...
+            this.status = SyncPeerStatus.LoadingSubDoc;
+            await this.connectDoc(subdoc, abortInner.signal);
+            this.status = SyncPeerStatus.Syncing;
+            this.updateSyncStatus();
+          }
+        })(),
+        // pull updates
+        (async () => {
+          while (throwIfAborted(abortInner.signal)) {
+            const { docId, data } = await this.state.pullUpdatesQueue.next(
```
forehalo, line 212: need empty data filter here as well

packages/frontend/workspace/src/providers/index.ts
EYHN: The second has been fixed.

I plan to solve the first issue when I refactor the workspace manager

packages/frontend/workspace/src/providers/index.ts
EYHN: Looks like it's no longer needed

packages/frontend/workspace/src/providers/index.ts
EYHN: Just leave this here and tell us this needs to be refactored.

packages/frontend/workspace/src/providers/sync/peer.ts
EYHN: This is not difficult, but requires some UI design to display this information, which will be implemented in the future

packages/frontend/core/src/hooks/current/use-current-sync-engine.ts
EYHN: The logic here is very different. The shared page is implemented by manually download the doc from the `react-router` loader.

There is currently no good way to tell SyncEngine not to load the entire workspace.

packages/frontend/workspace/src/providers/sync/peer.ts
```diff
@@ -0,0 +1,397 @@
...
+
+      this.status = SyncPeerStatus.Syncing;
+      this.updateSyncStatus();
+
+      // Finally: start sync
+      await Promise.all([
+        // listen subdocs
+        (async () => {
+          while (throwIfAborted(abortInner.signal)) {
+            const subdoc = await this.state.subdocsLoadQueue.next(
```
pengx17, line 200: May also need to prevent page reload when these queues have pending items

pengx17 (dismissed):

forehalo (dismissed):

packages/frontend/workspace/src/providers/sync/peer.ts
forehalo: only local peer need to be cared

pengx17: yes, should be only the one that writes to persistent data sources

EYHN: Yes, i will do that after this PR

packages/frontend/workspace/src/providers/sync/engine.ts
```diff
@@ -0,0 +1,223 @@
...
+            if (!signal.aborted) this.updateSyncingState(peers);
+          }).dispose
+        );
+      }
+
+      // Step 4: continuously sync local and remote
+
+      // wait for abort
+      await new Promise((_, reject) => {
+        if (signal?.aborted) {
```
joooye34, line 123: why here have a null check ?

packages/frontend/workspace/src/providers/sync/peer.ts
```diff
@@ -0,0 +1,397 @@
...
+  }
+  logger = new DebugLogger('affine:sync-peer:' + this.name);
+
+  constructor(
+    private rootDoc: Doc,
+    private storage: Storage
+  ) {
+    this.logger.debug('peer start');
+    this.status = SyncPeerStatus.LoadingRootDoc;
+    this.abort = new AbortController();
```
joooye34, line 59: The `abort` inited twice

packages/frontend/workspace/src/providers/sync/peer.ts
```diff
@@ -0,0 +1,397 @@
...
+
+  /**
+   * auto retry after 5 seconds if sync failed
+   */
+  async syncRetryLoop(abort: AbortSignal) {
+    while (abort.aborted === false) {
+      try {
+        await this.sync(abort);
+      } catch (err) {
+        if (err === 'manually-stop') {
```
joooye34, line 97: declare a constant variable to replace it

pengx17 (approved):

joooye34 (approved):

joooye34:
It is so clear when we have the storage and sync engine abstract!

PR #4909, opened 2023-11-10
fix(core): handle the getSession network error properly

Brooooooklyn:
If network offline or API error happens, the `session` returned by the `useSession` hook will be null, so we can't assume it is not null.

There should be following changes:
1. create a page in ErrorBoundary to let the user refetch the session.
2. The `SessionProvider` stop to pull the new session once the session is null, we need to figure out a way to pull the new session when the network is back or the user click the refetch button.


packages/frontend/core/src/components/affine/affine-error-eoundary.tsx
```diff
@@ -91,6 +94,18 @@ export class AffineErrorBoundary extends Component<
             </>
           </>
         );
+      } else if (error instanceof SessionFetchErrorRightAfterLoginOrSignUp) {
+        errorDetail = (
+          <>
+            <h1>Sorry.. there was an error</h1>
+            <>
+              <span> Fetching session failed </span>
+              <span onClick={error.refetchSession}>
```
joooye34, line 103: image

This click is too hard to use.

joooye34: Use `<a>` to replace `<span>`, and we need to implement the design after this PR. https://www.figma.com/file/UTTtfVCcStw1dKUg0gCSPo/%F0%9F%93%9A-Design-System?type=design&node-id=11684-4096&mode=design&t=EnURvqBaeKkWsQAo-0

joooye34: I am thinking. We should suggest user to refresh the page after too much retrying.

Brooooooklyn:
### Merge activity

* **Nov 17, 3:50 AM**: @Brooooooklyn merged this pull request with [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/4909).

PR #4912, opened 2023-11-11
fix(infra): compatibility fix for space prefix

pengx17:
It seems there are some cases that [this upstream PR](https://github.com/toeverything/blocksuite/pull/4747) will cause data loss.

In version 0.10, page id in spaces no longer has prefix "space:"
The data flow for fetching a doc's updates is:
- page id in `meta.pages` -> find `${page-id}` in `doc.spaces` -> `doc` -> `doc.guid`
if `doc` is not found in `doc.spaces`, a new doc will be created and its `doc.guid` is the same with its pageId
- because of guid logic change, the doc that previously prefixed with `space:` will not be found in `doc.spaces`
- when fetching the rows of this doc using the doc id === page id,
  it will return EMPTY since there is no updates associated with the page id

The provided fix in the PR will path the `spaces` field of the root doc so that after 0.10 the doc can still be found in the `spaces` map. It shall both of the idb & sqlite datasources.


packages/frontend/electron/src/helper/db/base-db-adapter.ts
```diff
@@ -29,6 +29,7 @@ export abstract class BaseSQLiteAdapter {
       if (maxVersion !== WorkspaceVersion.Surface) {
         await migrateToLatest(this.path, WorkspaceVersion.Surface);
       }
+      await applyGuidCompatibilityFix(this.db);
```
darkskygit, line 32: can we move this process as a migrate process?

packages/common/infra/src/blocksuite/index.ts
```diff
@@ -654,6 +656,32 @@ async function upgradeV2ToV3(options: UpgradeOptions): Promise<boolean> {
...
+// in version 0.10, page id in spaces no longer has prefix "space:"
+// The data flow for fetching a doc's updates is:
+// - page id in `meta.pages` -> find `${page-id}` in `doc.spaces` -> `doc` -> `doc.guid`
+// if `doc` is not found in `doc.spaces`, a new doc will be created and its `doc.guid` is the same with its pageId
+// - because of guid logic change, the doc that previously prefixed with "space:" will not be found in `doc.spaces`
+// - when fetching the rows of this doc using the doc id === page id,
+//   it will return empty since there is no updates associated with the page id
+export function guidCompatibilityFix(rootDoc: YDoc) {
+  let changed = false;
+  const prefix = 'space:';
```
forehalo, line 670: prefix could be `page:`

packages/frontend/electron/src/helper/db/base-db-adapter.ts
forehalo: once migrated，never pain

pengx17: DB operations in electron is already in a help process. I guess what we shall do is to migrate in the browser

Brooooooklyn (dismissed):

Brooooooklyn:
* stable
  * **PR #4912** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/4912?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/4912?utm_source=stack-comment).

Brooooooklyn (approved):

joooye34 (approved):

Brooooooklyn:
### Merge activity

* **Nov 13, 4:57 AM**: @Brooooooklyn started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/4912).

PR #4924, opened 2023-11-13
fix(storybook): page tags display

pengx17:
before
image

after
image


CatsJuice (commented):
It's better to test with Chinese characters mixed.

pengx17:
> It's better to test with Chinese characters mixed.

This is not an accurate implementation on the tags real length so I guess it is ok to do an estimation. If we use the real tag width here, we could do it better without using position sticky anyway. 🤔

CatsJuice (approved):

PR #4921, opened 2023-11-13
test(e2e): add subdoc migration test

Brooooooklyn:
test(e2e): add subdoc migration test

fix: remove .only

PR #4929, opened 2023-11-14
fix: cherry pick hotfix back to master

Brooooooklyn:
- fix(server): wrap updates applying in a transaction (#4922)

- fix(infra): compatibility fix for space prefix (#4912)

It seems there are some cases that [this upstream PR](https://github.com/toeverything/blocksuite/pull/4747) will cause data loss.

Because of some historical reasons, the page id could be different with its doc id.
It might be caused by subdoc migration in the following (not 100% sure if all white screen issue is caused by it) https://github.com/toeverything/AFFiNE/blob/0714c12703ff8840acc180633e5489c755ddfba3/packages/common/infra/src/blocksuite/index.ts#L538-L540

In version 0.10, page id in spaces no longer has prefix "space:"
The data flow for fetching a doc's updates is:
- page id in `meta.pages` -> find `${page-id}` in `doc.spaces` -> `doc` -> `doc.guid`
if `doc` is not found in `doc.spaces`, a new doc will be created and its `doc.guid` is the same with its pageId
- because of guid logic change, the doc that previously prefixed with `space:` will not be found in `doc.spaces`
- when fetching the rows of this doc using the doc id === page id,
  it will return EMPTY since there is no updates associated with the page id

The provided fix in the PR will patch the `spaces` field of the root doc so that after 0.10 the page doc can still be found in the `spaces` map. It shall apply to both of the idb & sqlite datasources.

Special thanks to @lawvs 's db file for investigation!

test(e2e): add subdoc migration test (#4921)

test(e2e): add subdoc migration test

fix: remove .only

fix(core): change server url of stable to insider (#4902) (#4926)


Brooooooklyn:
* master
  * **PR #4929** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/4929?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/4929?utm_source=stack-comment).

joooye34:
hmmmm, if use pr to pick commits, the author will be wrong.

PR #4933, opened 2023-11-14
perf(component): use png instead of svg for rendering noise svg

pengx17:
It looks like rendering svg with filter will cause some performance issue.
Replaced it with a pre-rendered png to mitigate it.

Note: there is some issue when embedding resource url directly into vanilla-extract. Not sure why yet.

related https://github.com/vanilla-extract-css/vanilla-extract/discussions/1209


Brooooooklyn:
@pengx17 let's target this pull request to stable

Issue #4727, opened 2023-10-26
Perhaps it would be better to disable the 'calt' feature for Inter font


fundon:
### What happened?

See issue: https://github.com/rsms/inter/issues/342

https://learn.microsoft.com/zh-cn/typography/opentype/spec/features_ae#tag-calt

Add a css rule to disable it:

```css
font-feature-settings: 'calt' 0;
```

https://github.com/toeverything/AFFiNE/assets/27926/52823e29-a1f4-4513-98d6-2370871cc32b

### Distribution version

macOS x64 (Intel)

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

PR #4969, opened 2023-11-17
fix: close font calt to avoid render twinkle

umar23faiz:
Fixed issue #4727 . I have added the property on globalcss so that it would be applicable accross the application.


umar23faiz:
https://github.com/toeverything/AFFiNE/assets/83939820/1fe1de8e-7786-4d79-a681-0b7c4ddb7df2

After Fixing

joooye34 (approved):

PR #4984, opened 2023-11-18
feat(workspace): more status for SyncEngine

EYHN:

PR #4983, opened 2023-11-18
feat(workspace): more status for SyncPeer

EYHN:

PR #4985, opened 2023-11-18
feat(core): notice unsaved changes when close browser

EYHN:


EYHN:
* canary
  * **PR #4983** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/4983?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #4984** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/4984?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #4985** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/4985?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/4985?utm_source=stack-comment).

packages/frontend/core/src/layouts/workspace-layout.tsx
```diff
@@ -145,6 +146,28 @@ export const WorkspaceLayoutInner = ({
...
   const t = useAFFiNEI18N();
+  const currentSyncEngineStatus = useCurrentSyncEngineStatus();
+
+  const haveUnsavedChanges = currentSyncEngineStatus?.local?.pendingPushUpdates;
+  useEffect(() => {
+    if (haveUnsavedChanges) {
+      const saveAlert = (event: BeforeUnloadEvent) => {
+        event.preventDefault();
+        return (event.returnValue =
+          'Data is not saved. Are you sure you want to leave?');
```
joooye34, line 157: i18n ?

joooye34 (dismissed):

packages/frontend/core/src/layouts/workspace-layout.tsx
```diff
@@ -145,6 +146,27 @@ export const WorkspaceLayoutInner = ({
   const { openPage } = useNavigateHelper();
   const pageHelper = usePageHelper(currentWorkspace.blockSuiteWorkspace);
   const t = useAFFiNEI18N();
+  const currentSyncEngineStatus = useCurrentSyncEngineStatus();
+
+  const haveUnsavedChanges = currentSyncEngineStatus?.local?.pendingPushUpdates;
```
pengx17: Based on `useCurrentSyncEngineStatus`, it seems the status is not a reactive value that will trigger React re-render on change.

packages/frontend/core/src/layouts/workspace-layout.tsx
```diff
@@ -145,6 +146,27 @@ export const WorkspaceLayoutInner = ({
...
+  useEffect(() => {
+    if (haveUnsavedChanges) {
```
pengx17, line 152: I think maybe it is better to register the events in module directly and prevent reloading for all sync engines instead of only the current one.

joooye34:
### Merge activity

* **Nov 20, 10:27 AM**: [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/4985) rebased this pull request after merging its parent, because this pull request is set to merge when ready.

packages/frontend/core/src/layouts/workspace-layout.tsx
EYHN: I change to useHaveUnsavedChanges, now only when haveUnsavedChanges changes between true or false will trigger rerender

packages/frontend/core/src/hooks/current/use-current-sync-engine.ts
```diff
@@ -31,3 +31,31 @@ export function useCurrentSyncEngineStatus(): SyncEngineStatus | undefined {
...
+          syncEngine.status.local.pendingPushUpdates > 0
+      );
+      return syncEngine.onStatusChange.on(status => {
+        setHaveUnsavedChanges(
+          !!status.local && status.local.pendingPushUpdates > 0
+        );
+      }).dispose;
+    } else {
+      setHaveUnsavedChanges(false);
+    }
```
pengx17, line 56: not tested:
```suggestion
      const updateStatus = (status: SyncEngineStatus) => {
        return !!status.local && status.local.pendingPushUpdates > 0
      };
      setHaveUnsavedChanges(updateStatus(syncEngine.status));
      return syncEngine.onStatusChange.on(updateStatus).dispose;
    } else {
      setHaveUnsavedChanges(false);
    }
```

packages/frontend/core/src/layouts/workspace-layout.tsx
EYHN: I think it is ok to put this in useEffect.

We may let SyncEngine run in worker in the future, I want to keep SyncEngine as clean as possible.

pengx17: I am a bit concerned about the case that the user switchs active workspaces when the push queue has a lot of pending items, and the user refresh the page in the meantime. It looks rare, but hmm ...

packages/frontend/core/src/hooks/current/use-current-sync-engine.ts
joooye34: @EYHN cc this?

packages/frontend/core/src/layouts/workspace-layout.tsx
joooye34: I agree with @pengx17

PR #4975, opened 2023-11-17
feat(server): reduce duplidated merge with cache

forehalo:

PR #4973, opened 2023-11-17
feat(server): add cache module

forehalo:

PR #5004, opened 2023-11-20
feat(server): impl doc history

forehalo:


forehalo:
* canary
  * **PR #4970** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/4970?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #4971** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/4971?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #5002** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5002?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #5004** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5004?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
      * **PR #4973** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/4973?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #4975** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/4975?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5004?utm_source=stack-comment).

packages/backend/server/src/modules/doc/history.ts
```diff
@@ -0,0 +1,237 @@
...
+      select: {
+        timestamp: true,
+      },
+      orderBy: {
+        timestamp: 'desc',
+      },
+    });
+  }
+
+  async recover(workspaceId: string, id: string, timestamp: Date) {
```
pengx17, line 139: I am thinking about the recover api in the history recovery workflow. I assume when calling `recover` there will not broadcast update events to peers right? So the workflow is:

1. ui: user chooses a snapshot in history view and click recover
2. ui: send "recover" api to the server
3. server: create a new snapshot and push to history
4. server: replace the latest snapshot with the given snapshot
5. ui: replace the given page's doc & apply the snapshot
6. ui: send updates to server
7. server: updates the snapshot
8. server: broadcast updates to other peers

It seems to me that there are some redundant in no.3/4 & 7. The issue is that when the user updates the local doc, it will send updates to the server to update the doc, but it might be already be updated.

**Updates:**
I searched yjs repo and seems this is [a discussed topic](https://github.com/yjs/yjs/issues/159#issuecomment-1525961593). It seems not possible force reset a doc to a certain state, unless create a new doc and replace the existing subodc in the doc tree with the new doc.

> Once a document that was created using createDocFromSnapshot syncs with other docs the discarded changes will be restored - unless you find a way to force everyone to discard the changes simultaneously.

So perhaps what we should do is not reseting the doc's update, but find a way to push a new update via blocksuite to the history that:
- erase all contents i
- import the doc into the page

🤔?

packages/backend/server/src/modules/workspaces/controller.ts
```diff
@@ -104,4 +105,51 @@ export class WorkspacesController {
...
+      }
+
+      ts = new Date(timeNum);
+    } catch (e) {
+      throw new Error('Invalid timestamp');
+    }
+
+    if (
+      // if a user has the permission
+      !(await this.permission.isAccessible(
```
darkskygit: isAccessible only check if has Read permission, we need > Write when a user want to read or restore the history?

packages/backend/server/src/modules/doc/history.ts
darkskygit: This feature has been implemented in the internal y-indexeddb before: https://github.com/toeverything/AFFiNE/blob/00c11d40cff9fa203f11864a01a927e52cbeaf82/packages/common/y-indexeddb/src/index.ts#L34-L72
We only need to implement it once in the server and then broadcast the changes

forehalo: oh I forget that, and the bad news is, the recover is doing in main server and the broadcast happens in sync server. we might need some hack.

Seems we could do such undo in client and just let the sync system to update the server data?

packages/backend/server/src/modules/workspaces/controller.ts
forehalo: we're doing the permission controlling badly. actually all users with `Read` permission can edit the docs, if we limit the history functionality to `Write` or `Admin`, users may complain about it

darkskygit (approved):

packages/backend/server/src/modules/doc/history.ts
darkskygit: Consider that we have stricter permission management requirements in the future (such as the team version), I don't think it is safe to roll back the version in frontend.
Malicious users can compile the malicious version to roll back and broadcast the update to the server
It is difficult for the server to prevent the broadcast of this kind of update information before we implement a update analysis service

PR #5005, opened 2023-11-20
feat: allow third party mail server

saranshisatgit:
Not sure this is the right approach since gmail services  is  default so this creates if the SendGrid is provided as the smtp settings !


packages/backend/server/src/modules/auth/mailer/mailer.ts
```diff
@@ -26,6 +26,17 @@ export const MAILER: FactoryProvider<
         },
       });
     }
+  // Create a transport for sendgrid 
+    if(config.auth.email.server == 'smtp.sendgrid.net'){
+      return createTransport({
+        host: config.auth.email.server,
+        port: 587,
```
darkskygit: can use `config.auth.email.port`

packages/backend/server/src/modules/auth/mailer/mailer.ts
```diff
@@ -26,6 +26,17 @@ export const MAILER: FactoryProvider<
         },
       });
     }
+  // Create a transport for sendgrid 
+    if(config.auth.email.server == 'smtp.sendgrid.net'){
```
darkskygit: can check if server not equal `smtp.gmail.com`

darkskygit (changes requested):

saranshisatgit:
@darkskygit This one could be merged thanks !

saranshisatgit:
hey @darkskygit any updates?

darkskygit:
this problem was fix by #5835 , thanks for your PR

Now you can use the following envs to define the email sender:
``` env
MAILER_HOST=
MAILER_PORT=
MAILER_USER=
MAILER_PASSWORD=
MAILER_SENDER=
```

PR #5006, opened 2023-11-20
fix(core): hmr issue on dev

pengx17:
I suspect HMR does not working properly on dev because we have multiple entries.
One relative issue: https://github.com/webpack/webpack-dev-server/issues/2792/

I think we do not need multiple entries for polyfills & plugins after all. They could be in the same chunk, and could be later optimized through splitChunks option.

`ses.ts` is changed to `ses-lockdown.ts` because `ses.ts` does not pass circular dependency check by madge. I haven't looked through the real root cause though. See https://github.com/pahen/madge/issues/355


pengx17:
* canary
  * **PR #5006** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5006?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5006?utm_source=stack-comment).

Brooooooklyn (dismissed):

Brooooooklyn (approved):

Issue #4873, opened 2023-11-08
Page crash when typing text in the quick search ​input


lawvs:
### What happened?

https://github.com/toeverything/AFFiNE/assets/18554747/5e60ab67-5561-4a4c-ac56-f15503441a5c

<img width="1161" alt="Screenshot 2023-11-09 at 02 03 06" src="https://github.com/toeverything/AFFiNE/assets/18554747/03399f18-bbf9-4816-a613-e5d715267d53">

### Distribution version

macOS x64 (Intel)

### What browsers are you seeing the problem on if you're using web version?

Chrome

### Relevant log output

```shell
Unexpected Application Error!
Failed to execute 'querySelector' on 'Element': '[cmdk-item=""][data-value="welcome to affine alpha "downhills"__>>>44c6dc3e950b462ca8b519aaae6f5992.affine:pages<<<____$$content_matched_without_subtitle$$__"]' is not a valid selector.
Error: Failed to execute 'querySelector' on 'Element': '[cmdk-item=""][data-value="welcome to affine alpha "downhills"__>>>44c6dc3e950b462ca8b519aaae6f5992.affine:pages<<<____$$content_matched_without_subtitle$$__"]' is not a valid selector.
  at https://prod.affineassets.com/744cd4748/js/chunk.1778-dc6374d8.js:1:7589
  at Object.aW [as useMemo] (https://prod.affineassets.com/744cd4748/js/3460-f4cafa9e.js:75:65280)
  at t.useMemo (https://prod.affineassets.com/744cd4748/js/npm-react-1ef8c35d.js:9:5984)
  at https://prod.affineassets.com/744cd4748/js/chunk.1778-dc6374d8.js:1:7548
  at ay (https://prod.affineassets.com/744cd4748/js/3460-f4cafa9e.js:75:60909)
  at oi (https://prod.affineassets.com/744cd4748/js/3460-f4cafa9e.js:75:71212)
  at s (https://prod.affineassets.com/744cd4748/js/3460-f4cafa9e.js:75:121508)
  at i_ (https://prod.affineassets.com/744cd4748/js/3460-f4cafa9e.js:75:99079)
  at https://prod.affineassets.com/744cd4748/js/3460-f4cafa9e.js:75:98946
  at iT (https://prod.affineassets.com/744cd4748/js/3460-f4cafa9e.js:75:98953)
```

### Anything else?

App Version
0.10.0-canary.14
Editor Version
[...]

PR #5008, opened 2023-11-20
fix(core): cmdk crash when entering double quotes

JimmFly:
Due to a bug in the upstream repository, a temporary fix was implemented until the issue in the upstream repository is resolved.
https://github.com/pacocoursey/cmdk/issues/189

### DIFF
--- a/packages/frontend/core/src/components/pure/cmdk/data.tsx
+++ b/packages/frontend/core/src/components/pure/cmdk/data.tsx
@@ -24,7 +24,7 @@ import {
   PreconditionStrategy,
 } from '@toeverything/infra/command';
 import { atom, useAtomValue } from 'jotai';
-import { escape, groupBy } from 'lodash-es';
+import { groupBy } from 'lodash-es';
 import { useCallback, useMemo } from 'react';
 
 import {
@@ -42,6 +42,10 @@ import type { CMDKCommand, CommandContext } from './types';
 interface SearchResultsValue {
   space: string;
   content: string;
+}
+
+export function removeDoubleQuotes(str?: string): string | undefined {
+  return str?.replace(/"/g, '');
 }
 
 export const cmdkQueryAtom = atom('');
@@ -168,7 +172,6 @@ export const pageToCommand = (
   const commandLabel = label || {
     title: title,
   };
-  const escapedTitle = escape(title);
 
   return {
     id: page.id,
@@ -176,13 +179,8 @@ export const pageToCommand = (
     // hack: when comparing, the part between >>> and <<< will be ignored
     // adding this patch so that CMDK will not complain about duplicated commands
     value:
-      escapedTitle +
-      valueWrapperStart +
-      page.id +
-      '.' +
-      category +
-      valueWrapperEnd,
-    originalValue: escapedTitle,
+      title + valueWrapperStart + page.id + '.' + category + valueWrapperEnd,
+    originalValue: title,
     category: category,
     run: () => {
       if (!currentWorkspaceId) {
@@ -429,7 +427,10 @@ export const customCommandFilter = (value: string, search: string) => {
     label = label.replace(contentMatchedWithoutSubtitle, '');
   }
 
-  const originalScore = commandScore(label, search);
+  // use to remove double quotes from a string until this issue is fixed
+  // https://github.com/pacocoursey/cmdk/issues/189
+  const escapedSearch = removeDoubleQuotes(search) || '';
+  const originalScore = commandScore(label, escapedSearch);
 
   // hack to make the page title result always before the content result
   // if the command has matched the title but not the subtitle,
--- a/packages/frontend/core/src/components/pure/cmdk/main.tsx
+++ b/packages/frontend/core/src/components/pure/cmdk/main.tsx
@@ -12,6 +12,7 @@ import {
   cmdkQueryAtom,
   cmdkValueAtom,
   customCommandFilter,
+  removeDoubleQuotes,
   useCMDKCommandGroups,
 } from './data';
 import { HighlightLabel } from './highlight';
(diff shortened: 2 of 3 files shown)


JimmFly:
* canary
  * **PR #5008** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5008?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5008?utm_source=stack-comment).

packages/frontend/core/src/components/pure/cmdk/data.tsx
```diff
@@ -168,21 +169,15 @@ export const pageToCommand = (
...
     // adding this patch so that CMDK will not complain about duplicated commands
     value:
-      escapedTitle +
-      valueWrapperStart +
-      page.id +
-      '.' +
-      category +
-      valueWrapperEnd,
-    originalValue: escapedTitle,
+      title + valueWrapperStart + page.id + '.' + category + valueWrapperEnd,
```
pengx17, line 182: I am curious why the existing solution does not work. Can you please explain?

packages/frontend/core/src/components/pure/cmdk/main.tsx
```diff
@@ -44,6 +44,10 @@ const categoryToI18nKey: Record<CommandCategory, i18nKey> = {
   'editor:page': 'com.affine.cmdk.affine.category.editor.page',
 };
 
+export function removeDoubleQuotes(str?: string): string | undefined {
+  return str?.replace(/"/g, '');
```
EYHN: Do we need to care about other special characters besides double quotes?

packages/frontend/core/src/components/pure/cmdk/data.tsx
JimmFly: It appears to work well for searching existing articles, however, in reality, when you solely search for a double quote, the results turn up entirely empty. Also, when I attempted to alter the value of the ensuing command using the same method, the problem arose. I am incapable of creating any pages using the command.

Nonetheless, this issue only arises when inputting double quotations. Other special symbols seem to be managed efficiently. I attempted to escape the value of the input and command item, but it necessitated a change in the upstream repository, hence, I abandoned this idea.

```
...
      if (results.every(command => command.originalValue !== query)) {
        const escapedQuery = escape(query);
        results.push({
          id: 'affine:pages:create-page',
          label: t['com.affine.cmdk.affine.create-new-page-as']({
            keyWord: query,
          }),
          value: 'affine::create-page' + escapedQuery, 
          category: 'affine:creation',
          run: async () => {
            const page = pageHelper.createPage();
            await page.waitForLoaded();
            pageMetaHelper.setPageTitle(page.id, query);
          },
          icon: <PageIcon />,
        });
...
```
image
image

packages/frontend/core/src/components/pure/cmdk/main.tsx
JimmFly: I gave it a few tries and couldn’t pinpoint any additional special characters that might trigger the crash. Could you possibly have another go, though? I’m uncertain whether my limited testing of special symbols has encompassed all scenarios.

packages/frontend/core/src/components/pure/cmdk/data.tsx
pengx17: Sounds fair enough.
We may not need to worry about too much on how CMDK library for searching & sorting later, since implementing an in-house solution for that is already planned :/

pengx17 (dismissed):

packages/frontend/core/src/components/pure/cmdk/data.tsx
```diff
@@ -37,6 +37,7 @@ import { useCurrentWorkspace } from '../../../hooks/current/use-current-workspac
 import { useNavigateHelper } from '../../../hooks/use-navigate-helper';
 import { WorkspaceSubPath } from '../../../shared';
 import { usePageHelper } from '../../blocksuite/block-suite-page-list/utils';
+import { removeDoubleQuotes } from '.';
```
pengx17: Please do not import `.`.

packages/frontend/core/src/components/pure/cmdk/main.tsx
```diff
@@ -44,6 +44,10 @@ const categoryToI18nKey: Record<CommandCategory, i18nKey> = {
   'editor:page': 'com.affine.cmdk.affine.category.editor.page',
 };
 
+export function removeDoubleQuotes(str?: string): string | undefined {
```
pengx17: Maybe move this function to `data` and escape the `command.value` directly to `pageToCommand`.

EYHN (dismissed):

lawvs:
Related to https://github.com/toeverything/AFFiNE/issues/4873

packages/frontend/core/src/components/pure/cmdk/main.tsx
```diff
@@ -78,11 +79,15 @@ const QuickSearchGroup = ({
                 title: command.label,
               }
             : command.label;
+
+        // use to remove double quotes from a string until this issue is fixed
+        // https://github.com/pacocoursey/cmdk/issues/189
+        const escapeValue = removeDoubleQuotes(command.value);
```
pengx17, line 85: As I said, you should escape the value in `pageToCommand`

JimmFly: The implementation of create page as xxx is not based on pageToCommand, and it seems that there is no guarantee that other commands will not have similar problems in the future.

pengx17: hmm, sounds good.

pengx17 (dismissed):

EYHN (approved):

PR #5022, opened 2023-11-22
refactor(core): use manual upgrade to replace auto migration when web setup

joooye34:
1. Split logic in `packages/common/infra/src/blocksuite/index.ts` to multiple single files
2. Move migration logic from setup to upgrade module, to prevent auto migration problems and loading problem


joooye34:
* canary
  * **PR #5022** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5022?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5022?utm_source=stack-comment).

packages/common/infra/src/blocksuite/initialzation/index.ts
```diff
@@ -0,0 +1,289 @@
+import { assertExists } from '@blocksuite/global/utils';
+import type { Page, PageMeta, Workspace } from '@blocksuite/store';
+import type { Schema } from '@blocksuite/store';
+import type { createStore, WritableAtom } from 'jotai/vanilla';
+import { nanoid } from 'nanoid';
+
+export async function initEmptyPage(page: Page, title?: string) {
+  await page.waitForLoaded();
+  const pageBlockId = page.addBlock('affine:page', {
```
EYHN: blocksuite has a new way to initial block structure

https://block-suite.com/data-persistence.html#block-tree-initialization-basics

```ts
await page.load(() => {
  const pageBlockId = page.addBlock('affine:page', {
    title: new page.Text(title ?? ''),
  });
  page.addBlock('affine:surface', {}, pageBlockId);
  const noteBlockId = page.addBlock('affine:note', {}, pageBlockId);
  page.addBlock('affine:paragraph', {}, noteBlockId);
});
```

joooye34: Those codes I just moved from `index.ts`, maybe refactor next PR, I can add some comments.

pengx17 (commented):
LGTM. So in the new impl there will not be auto migration in the background and always brings the migration route for user to do it manually right?

joooye34:
> LGTM. So in the new impl there will not be auto migration in the background and always brings the migration route for user to do it manually right?

@pengx17 No background logic, let the user determine when to migrate.

pengx17:
> @pengx17 No background logic, let the user determine when to migrate.

With the new impl, I think we can safely move some setup logic from setup to workspace route. For example, we do not need to await for rootWorkspacesMetadataAtom for non `/workspaces` route (like 404 pages).

packages/frontend/core/src/components/workspace-upgrade/upgrade-hooks.ts
```diff
@@ -1,42 +1,126 @@
...
   const [workspace] = useCurrentWorkspace();
   const syncEngine = useCurrentSyncEngine();
+  const rootStore = getCurrentStore();
 
-  const upgradeWorkspace = useCallback(() => {
+  const upgradeWorkspace = useAsyncCallback(async () => {
     setState('upgrading');
     setError(null);
+    try {
+      if (environment.isDesktop) {
```
EYHN: I'm not so sure,
It seems that Decktop starting stage upgrade only handles local workspace? cloud workspace may still need this function.

tests/affine-migration/e2e/basic.spec.ts
```diff
@@ -80,81 +79,58 @@ test('v2 to v4, database migration', async ({ page }) => {
...
 
-  //#region fixme(himself65): blocksuite issue, data cannot be loaded to store
-  const allPagePath = `${coreUrl}/workspace/${localStorageData.last_workspace_id}/all`;
-  await page.goto(allPagePath);
-  await page.waitForTimeout(5000);
-  //#endregion
-
   const detailPagePath = `${coreUrl}/workspace/${localStorageData.last_workspace_id}/${localStorageData.last_page_id}`;
   await page.goto(detailPagePath);
+  await page.waitForTimeout(3000);
```
EYHN: playwright doesn't recommend using waitForTimeout, which can cause erratic results.

https://playwright.dev/docs/api/class-frame#frame-wait-for-timeout

packages/common/infra/src/blocksuite/initialzation/index.ts
```diff
@@ -0,0 +1,289 @@
...
+    title: new page.Text(title ?? ''),
+  });
+  page.addBlock('affine:surface', {}, pageBlockId);
+  const noteBlockId = page.addBlock('affine:note', {}, pageBlockId);
+  page.addBlock('affine:paragraph', {}, noteBlockId);
+}
+
+export async function buildEmptyBlockSuite(workspace: Workspace) {
```
EYHN: I think this is dead code.  I searched that this function is not referenced anywhere.

packages/frontend/core/src/components/workspace-upgrade/upgrade-hooks.ts
```diff
@@ -1,42 +1,126 @@
...
+          'jotai-workspaces',
+          JSON.stringify(localMetadataList)
+        );
+        localStorage.setItem('last_workspace_id', newId);
+        localStorage.removeItem('last_page_id');
+      } else {
+        applyDoc(workspace.blockSuiteWorkspace.doc, resultDoc);
+      }
+
       await syncEngine?.waitForSynced();
```
pengx17, line 111: this `waitForSynced` is called after migration is to wait for changes to be written to data sources?

packages/frontend/core/src/components/workspace-upgrade/upgrade-hooks.ts
pengx17: Agreed. Assume on the desktop the user enabled cloud and he opens a never opened outdated cloud workspace, the data will be pulled from remote and will still go into this function.

packages/frontend/core/src/components/workspace-upgrade/upgrade-hooks.ts
EYHN: We will display a "refresh page" button. After this sync is completed, the user can safely click refresh.

packages/frontend/core/src/components/workspace-upgrade/upgrade-hooks.ts
joooye34: Good point! This condition can be removed!

EYHN (approved):

pengx17 (approved):

PR #5023, opened 2023-11-22
feat(core): add download app button to web

JimmFly:
image


JimmFly:
* canary
  * **PR #5023** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5023?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5023?utm_source=stack-comment).

hwangdev97:
The current non-login status is ok.
When using storybook, we can not check the status after logging in. I think if you have time, you may need to find a solution.

packages/frontend/component/src/components/app-sidebar/app-download-button/index.css.ts
CatsJuice, line 1: Is this button's animation same as [AppUpdaterButton](https://github.com/toeverything/AFFiNE/blob/canary/packages/frontend/component/src/components/app-sidebar/app-updater-button/index.css.ts)? @hwangdev97

If so, it's better to reuse the animation.

packages/frontend/component/src/components/affine-banner/download-client.tsx
```diff
@@ -1,44 +1,54 @@
...
+import * as styles from './index.css';
+
+type DownloadTipsProps = {
+  isLoggedIn: boolean;
+  onLogin: () => void;
+  onEnableCloud: () => void;
+  onClose: () => void;
+};
+
+export const DownloadTips = ({
```
joooye34: This component not download now, it is local demo tips.

packages/frontend/component/src/components/app-sidebar/app-download-button/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  style?: React.CSSProperties;
+}) {
+  const [show, setShow] = useState(true);
+
+  const handleClose = useCallback(() => {
+    setShow(false);
+  }, []);
+
+  const handleClick = useCallback(() => {
+    const url = `https://affine.pro/download?channel=stable`;
```
joooye34, line 23: Leave a `TODO` comment to unify this type of literal value.

packages/frontend/core/src/components/top-tip.tsx
```diff
@@ -0,0 +1,130 @@
...
+export const TopTip = ({
+  workspace,
+}: {
+  workspace: AffineOfficialWorkspace;
+}) => {
+  const loginStatus = useCurrentLoginStatus();
+  const isLoggedIn = loginStatus === 'authenticated';
+
+  const [showWarning, setShowWarning] = useState(false);
+  const [showDownloadTip, setShowDownloadTip] = useState(true);
```
joooye34: It will show this tip every time, do we need to use localstorage to record close state?

packages/frontend/core/src/components/top-tip.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { useCallback, useEffect, useState } from 'react';
+
+import { authAtom } from '../atoms';
+import { useCurrentLoginStatus } from '../hooks/affine/use-current-login-status';
+import { useOnTransformWorkspace } from '../hooks/root/use-on-transform-workspace';
+import { EnableAffineCloudModal } from './affine/enable-affine-cloud-modal';
+
+const minimumChromeVersion = 102;
+
+const shouldShowWarning = () => {
```
joooye34:
```suggestion
const shouldShowWarning = (() => {
if (environment.isDesktop) {
    // even though desktop has compatibility issues,
    //  we don't want to show the warning
    return false;
  }
  if (!environment.isBrowser) {
    // disable in SSR
    return false;
  }
  if (environment.isChrome) {
    return environment.chromeVersion < minimumChromeVersion;
  } else {
    return !environment.isMobile;
  }
})();
```

Calculate the result here, not need a funtion.

packages/frontend/core/src/components/top-tip.tsx
```diff
@@ -0,0 +1,130 @@
...
+    return !environment.isMobile;
+  }
+};
+
+const OSWarningMessage = () => {
+  const t = useAFFiNEI18N();
+  const [notChrome, setNotChrome] = useState(false);
+  const [notGoodVersion, setNotGoodVersion] = useState(false);
+  useEffect(() => {
+    setNotChrome(environment.isBrowser && !environment.isChrome);
```
joooye34: Those value could be initialize value, no need to use effect to calculate?

packages/frontend/core/src/components/top-tip.tsx
JimmFly: No, that's the design, we need it to keep appearing

packages/frontend/component/src/components/app-sidebar/app-download-button/index.css.ts
JimmFly: Does the latest change align with your intent? If you have any suggestions, please let me know directly.

CatsJuice (approved):

PR #5028, opened 2023-11-22
fix(electron): clone db file when enable cloud for desktop

pengx17:
At the moment on desktop the user's local blob data will be lost after enable cloud.
This is because blob data is only synced from old idb to new idb, but not sync into sqlitedb.

This pr will simply clone the db file for desktop app. It should also speed up the time when enabling cloud for a large local workspace.

### DIFF
--- a/packages/common/infra/src/type.ts
+++ b/packages/common/infra/src/type.ts
@@ -200,6 +200,7 @@ export type WorkspaceHandlers = {
   list: () => Promise<[workspaceId: string, meta: WorkspaceMeta][]>;
   delete: (id: string) => Promise<void>;
   getMeta: (id: string) => Promise<WorkspaceMeta>;
+  clone: (id: string, newId: string) => Promise<void>;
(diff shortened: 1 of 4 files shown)


pengx17:
* canary
  * **PR #5028** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5028?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5028?utm_source=stack-comment).

Brooooooklyn (commented):
there may some issues on Windows I think...

pengx17:
> there may some issues on Windows I think...

I am not sure but this only involves copying files. Will it fail randomly on windows as well?

EYHN (approved):

PR #5037, opened 2023-11-23
refactor(workspace): blob sync

EYHN:
This pr implements a blob engine.
It exposes a single `BlobStorage` to the `blocksuite`, and in it we sync blobs between multiple storages.

The implement still have few issues, but we can merge this pr first and fix them in future.

* BlobEngine currently **do nothing when delete**, because synchronization logic conflicts with deletion logic.
* BlobEngine sync between storages by querying the blob list at regular intervals. This will **cause many queries**, we can avoid this in the future by subscribing to remote changes.


EYHN:
* canary
  * **PR #5037** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5037?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5037?utm_source=stack-comment).

packages/frontend/workspace/src/blob/engine.ts
```diff
@@ -0,0 +1,139 @@
...
+            if (data) {
+              await remote.set(key, data);
+            }
+          } catch (err) {
+            logger.error(
+              `error when sync ${key} from [${this.local.name}] to [${remote.name}]`,
+              err
+            );
+          }
+        }
```
pengx17, line 46: Shall we make it parallel here?

EYHN: parallel may cause lag and affect other tasks.

packages/frontend/core/src/layouts/workspace-layout.tsx
```diff
@@ -116,10 +117,48 @@ type WorkspaceLayoutProps = {
...
+      }
+
+      blobEngine
+        ?.sync()
+        .catch(error => {
+          console.error('sync blob error', error);
+        })
+        .finally(() => {
+          // sync every 1 minute
+          setTimeout(() => {
```
Brooooooklyn: interval?

packages/frontend/core/src/layouts/workspace-layout.tsx
```diff
@@ -116,10 +117,48 @@ type WorkspaceLayoutProps = {
...
+        });
+    }
+
+    // after currentWorkspace changed, wait 1 second to start sync
+    setTimeout(() => {
+      sync();
+    }, 1000);
+
+    return () => {
+      stopped = true;
```
Brooooooklyn, line 148: why don't clear timeout here

packages/frontend/workspace/src/blob/storage/indexeddb.ts
```diff
@@ -0,0 +1,32 @@
+import { createStore, del, get, keys, set } from 'idb-keyval';
+
+import type { BlobStorage } from '../engine';
+
+export const createIndexeddbBlobStorage = (
```
pengx17, line 5: may need blocksuite team's awareness that we moved idb's blob storage into affine.
@doodlewind

packages/frontend/core/src/layouts/workspace-layout.tsx
EYHN: If there are many large blobs, the sync may take more than a minute, and we only time it after the sync is finished.

packages/frontend/workspace/src/blob/storage/indexeddb.ts
```diff
@@ -0,0 +1,32 @@
...
+  workspaceId: string
+): BlobStorage => {
+  const db = createStore(`${workspaceId}_blob`, 'blob');
+  const mimeTypeDb = createStore(`${workspaceId}_blob_mime`, 'blob_mime');
```
pengx17, line 9: I think we do not need to create another store for mine type. The mine type can be saved within the same value object in `xxx_blob`.
However this involves some migration efforts though.

packages/frontend/core/src/layouts/workspace-layout.tsx
EYHN: Because clear timeout is useless when running the promise

```
 -> promise
   .finish(setTimeout)
```

We still need a variable

packages/frontend/workspace/src/blob/storage/indexeddb.ts
EYHN: Good suggestion 
Here is compatible with the original implementation.

pengx17 (approved):

packages/frontend/core/src/layouts/workspace-layout.tsx
Brooooooklyn: If not cleared, it seems that an increasing number of recursive calls will accumulate.

EYHN: Recursion will be interrupted when `stopped == true`

PR #5045, opened 2023-11-23
feat: optional payment for self hosted server

darkskygit:
some hardcode logic will be fix after #4720 merged


JimmFly (approved):
Got it

darkskygit:
moved to #5055 #5056

PR #5051, opened 2023-11-24
chore: bump blocksuite

Flrande:
https://github.com/toeverything/blocksuite/pull/5337


packages/frontend/hooks/src/use-block-suite-page-preview.ts
```diff
@@ -60,7 +60,7 @@ export function useBlockSuitePagePreview(page: Page | null): Atom<string> {
         page.slots.ready.on(() => {
           set(getPagePreviewText(page));
         }),
-        page.slots.yUpdated.on(() => {
```
EYHN: Will block content changes trigger the blockUpdated event?

Flrande: I will try to use page indexer to get all references.

EYHN (approved):

pengx17 (approved):

PR #5055, opened 2023-11-24
feat: optional payment for server

### LINKED ISSUE #5003: Make Payment Module/Frontend optional
This is a prerequisite task for the self-host feature.

darkskygit:


darkskygit:
* canary
  * **PR #5055** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5055?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #5056** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5056?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5055?utm_source=stack-comment).

packages/backend/server/src/modules/index.ts
```diff
@@ -22,30 +22,40 @@ switch (SERVER_FLAVOR) {
   case 'sync':
     BusinessModules.push(SyncModule, DocModule.forSync());
     break;
+  case 'selfhosted':
```
forehalo, line 25: Actually I have practiced before about module plugin/out system.

Just like how Gitlab did, we could make it an Application Level configs stored in DB, and before client init modules, it should fetching the features list from server so it can know which modules should be enabled and whichs are not.

darkskygit: make backend module load optional and configable is not complicated, but currently, many requests on the frontend are executed in the background and cannot be skipped or fault-tolerant, if backend disable some module, frontend will fault when request the apis of the corresponding module. i think we need refactor frontend data stream before make server module optional
in currently stage, only payment module need to disable in come scene(mainly in selfhosted), so i think provide a new flavor of server is acceptable

forehalo: make a plan

forehalo (approved):

packages/backend/server/src/modules/index.ts
darkskygit: https://github.com/toeverything/AFFiNE/issues/5057 already make a issue

Brooooooklyn (approved):

PR #5060, opened 2023-11-24
refactor: workspace manager

EYHN:


EYHN:
* `canary`
  * **PR #5285** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5285?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5060** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5060?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5060?utm_source=stack-comment).

packages/common/infra/src/blocksuite/migration/fixing.ts
```diff
@@ -12,36 +12,41 @@ import { applyUpdate, encodeStateAsUpdate } from 'yjs';
...
     });
-  });
-  return changed;
+  }
+
+  doFix();
+
+  // do fix every time when meta or spaces changed
+  rootDoc.getMap('meta').observe(() => doFix());
+  rootDoc.getMap('spaces').observe(() => doFix());
```
pengx17: need to unobserve if you change the function (and fixWorkspaceVersion).
also rename the function to make sure it is not being misused?

packages/frontend/hooks/src/use-workspace-blob.ts
```diff
@@ -0,0 +1,37 @@
...
+    setBlob(undefined);
+    if (!blobKey || !meta) {
+      return;
+    }
+    let canceled = false;
+    workspaceManager
+      .getWorkspaceBlob(meta, blobKey)
+      .then(blob => {
+        if (blob && !canceled) {
+          setBlob(URL.createObjectURL(blob));
```
pengx17: 1. better revoke the object url if it is not used any longer
2. or, we can cache the url if it is already in cache

packages/common/infra/src/blocksuite/migration/fixing.ts
EYHN: Because we might be sync incorrect data from other nodes, we run the fix repeatedly with each update.

We don't need `unobserve` because we run the fix repeatedly.

packages/frontend/workspace/src/list/cache.ts
```diff
@@ -0,0 +1,23 @@
...
+      console.error('cannot parse worksapce', e);
+    }
+    return [];
+  }
+  return [];
+}
+
+export function writeWorkspaceListCache(metadata: WorkspaceMetadata[]) {
+  const metadataMap = new Map(metadata.map(x => [x.id, x]));
+  metadata = Array.from(metadataMap.values());
```
pengx17: why trasform the data back and forth using a map? to dedupe the values by id?

packages/common/infra/src/blocksuite/migration/fixing.ts
pengx17: I thought we may need to unobserve these on switching or deleting a workspace, but after reading the related code it seems the assumption is that a workspace will be opened once only.
I think it is still a good practice that return disposable for any event binding functions though to make sure this function is not being misused.

packages/frontend/hooks/src/use-workspace-info.ts
```diff
@@ -0,0 +1,30 @@
...
+export function useWorkspaceInfo(
+  meta: WorkspaceMetadata,
+  workspace?: Workspace
+) {
+  const workspaceManager = useAtomValue(workspaceManagerAtom);
+
+  const [information, setInformation] = useState<{
+    avatar?: string;
+    name?: string;
+  } | null>(null);
```
pengx17: it looks to me this hook may always return non-null values since `getInformation` will create one instance if not exist.
maybe init initial value using `workspaceManager.list.getInformation(meta).info`?

packages/frontend/workspace/src/pool.ts
```diff
@@ -0,0 +1,86 @@
...
+  private gc() {
+    for (const [id, { workspace, rc }] of new Map(
+      this.openedWorkspaces /* clone the map, because the origin will be modified during iteration */
+    )) {
+      if (rc === 0 && workspace.canGracefulStop()) {
+        // we can safely close the workspace
+        logger.info(`close workspace [${workspace.flavour}] ${workspace.id}`);
+        workspace.forceStop();
+
+        this.openedWorkspaces.delete(id);
```
pengx17, line 71: It seems that opening a workspace is not sideeffect-free and it is not closed after being deleted from pool. When a workspace is gc'ed and opened again, the side effects will be registered again.

EYHN: workspace.forceStop should be able to remove all sideeffect, including stop all setInterval, websocket, and broadcast channel.
The entire workspace including yjs doc should be gc'ed by the js engine.

And we not open a workspace twice, it will be a new workspace instance the next time it is opened.

pengx17:
> 
> 
> workspace.forceStop should be able to remove all sideeffect, including stop all setInterval, websocket, and broadcast channel.
> The entire workspace including yjs doc should be gc'ed by the js engine.
> 
> And we not open a workspace twice, it will be a new workspace instance the next time it is opened.

👍 makes sense

packages/frontend/workspace/src/pool.ts
pengx17: 👍 makes sense

packages/common/env/src/workspace.ts
```diff
@@ -14,73 +8,6 @@ export enum WorkspaceSubPath {
   SHARED = 'shared',
 }
 
-export interface AffineDownloadProvider extends PassiveDocProvider {
```
forehalo, line 17: [celebrate](https://app.graphite.dev/macros/toeverything)

packages/common/env/src/workspace.ts
```diff
@@ -109,71 +35,33 @@ export const settingPanel = {
   Export: 'export',
   Sync: 'sync',
 } as const;
-export const settingPanelValues = [...Object.values(settingPanel)] as const;
```
forehalo: [](https://app.graphite.dev/macros/toeverything)

packages/frontend/core/src/components/affine/new-workspace-setting-detail/delete-leave-workspace/index.tsx
```diff
@@ -33,18 +48,41 @@ export const DeleteLeaveWorkspace = ({
...
+          WorkspaceSubPath.ALL,
+          RouteLogic.REPLACE
+        );
+      } else {
+        jumpToIndex(RouteLogic.REPLACE);
+      }
     }
-  }, [onDeleteCloudWorkspace, onDeleteLocalWorkspace, workspace.flavour]);
+
+    await workspaceManager.deleteWorkspace(workspaceMetadata);
```
forehalo, line 70: should be put before navigation?

packages/frontend/core/src/components/affine/new-workspace-setting-detail/export.tsx
```diff
@@ -1,69 +1,35 @@
...
+    if (saving || !workspace) {
       return;
     }
-    setSyncing(true);
+    setSaving(true);
     try {
-      await syncBlobsToSqliteDb(workspace);
-      await syncDocsToSqliteDb(workspace);
+      await workspace.engine.sync.waitForSynced();
+      await workspace.engine.blob.sync();
```
forehalo, line 32: a progress indicator is kind of necessary

packages/frontend/core/src/components/affine/new-workspace-setting-detail/index.tsx
```diff
@@ -3,47 +3,38 @@ import {
...
-      );
-    } else {
-      return null;
-    }
-  }, [t, workspace]);
+  // useWorkspace hook is a vary heavy operation here, but we need syncing name and avatar changes here,
+  // we don't have a better way to do this now
+  const workspace = useWorkspace(workspaceMetadata);
+
+  const workspaceInfo = useWorkspaceInfo(props.workspaceMetadata);
```
forehalo: `workspaceMetadata` instead of `props.workspaceMetadata`

packages/frontend/core/src/components/affine/page-history-modal/data.ts
```diff
@@ -107,7 +106,7 @@ const workspaceMap = new Map<string, Workspace>();
 const getOrCreateWorkspace = (workspaceId: string) => {
   let workspace = workspaceMap.get(workspaceId);
   if (!workspace) {
-    const blobEngine = createAffineCloudBlobEngine(workspaceId);
+    const blobEngine = null as any; // createAffineCloudBlobEngine(workspaceId);
```
forehalo: is this safe?

packages/frontend/hooks/src/use-workspace-info.ts
forehalo: yeah, the init value of useState could be a getter function

packages/frontend/hooks/src/use-workspace-status.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  // avoid re-render when selector is changed
+  const [cachedSelector] = useState(() => selector);
+
+  useEffect(() => {
+    if (!workspace) {
+      setStatus(null);
+      return;
+    }
+    setStatus(
```
forehalo, line 25: same as above, all initial `setState` in effect could be moved to `useState` init function, to reduce one page rendering

packages/common/env/src/workspace.ts
EYHN: just fix lint error

packages/frontend/core/src/components/affine/new-workspace-setting-detail/delete-leave-workspace/index.tsx
EYHN: It is best to navigate first. delete current workspace may cause render a flashing blank page.

forehalo (commented):
[ci](https://app.graphite.dev/macros/toeverything)

packages/frontend/hooks/src/use-workspace-blob.ts
EYHN: i revoke the object url on this effect cleanup function

packages/frontend/workspace/src/list/cache.ts
EYHN: This is copy from old code
now it's removed

packages/frontend/core/src/components/affine/new-workspace-setting-detail/export.tsx
EYHN: true

tests/affine-migration/e2e/basic.spec.ts
```diff
@@ -38,9 +38,6 @@ test('v1 to v4', async ({ page }) => {
   await expect(page.getByTestId('upgrade-workspace-button')).toBeVisible();
   await page.getByTestId('upgrade-workspace-button').click();
 
-  await expect(page.getByText('Refresh Current Page')).toBeVisible();
```
joooye34, line 41: why delete those codes, no upgrade happened?

EYHN: I removed the Refresh button after the update, and now it automatically resumes the normal page after update.

packages/frontend/hooks/src/use-workspace-blob.ts
```diff
@@ -0,0 +1,40 @@
+import { workspaceManagerAtom } from '@affine/workspace/atom';
+import type { WorkspaceMetadata } from '@affine/workspace/metadata';
+import { useAtomValue } from 'jotai';
+import { useEffect, useState } from 'react';
+
+export function useWorkspaceBlobObjectUrl(
```
pengx17, line 6: The way this hook works will re-run the effect whenever it re-mounts (for example, re-opening the workspace list dropdown) and causing the workspace avatar blinks.
It seems the blob's life-cycle could also be bound to workspace manager & using a lru cache for creating the object url

packages/frontend/core/src/providers/session-provider.tsx
```diff
@@ -22,14 +21,14 @@ const SessionDefence = (props: PropsWithChildren) => {
...
-  const refreshMetadata = useSetAtom(refreshRootMetadataAtom);
   const onceSignedInEvents = useOnceSignedInEvents();
   const t = useAFFiNEI18N();
 
   const refreshAfterSignedInEvents = useAsyncCallback(async () => {
     await onceSignedInEvents();
-    refreshMetadata();
-  }, [onceSignedInEvents, refreshMetadata]);
+    // TODO: refresh metadata
+    // refreshMetadata();
```
pengx17: It seems the workspace list does not refresh after logged in, which is critical.
Do you have any idea on fixing this issue?

packages/frontend/workspace/src/list/index.ts
```diff
@@ -0,0 +1,292 @@
...
+        });
+        if (this.abortController.signal.aborted) {
+          unsubscribe();
+          return;
+        }
+        this.abortController.signal.addEventListener('abort', () => {
+          unsubscribe();
+        });
+        try {
+          const list = await provider.getList();
```
pengx17, line 255: I guess we may need to refresh periodically the list if the user has logged in.

packages/frontend/workspace/src/manager.ts
```diff
@@ -0,0 +1,189 @@
...
+ * Once opened, workspaces are stored in the `WorkspacePool`.
+ *
+ * # Pool
+ *
+ * The `WorkspacePool` use reference counting to manage active workspaces.
+ * Calling `use()` to create a reference to the workspace. Calling `release()` to release the reference.
+ * When the reference count is 0, it will close the workspace.
+ *
+ */
+export class WorkspaceManager {
```
pengx17, line 54: The design looks good to me. One little concern is about data reactivity, that how does the React side bind view to the properties?

packages/frontend/hooks/src/use-workspace-blob.ts
EYHN: I tried adding a lru cache for the blob, but the avatar will still flicker briefly because we create a new object url every time, but object url is hard to cache...

I think I'll leave this issue and resolve it in the future.

packages/frontend/core/src/providers/session-provider.tsx
EYHN: Oh, I missed this todo.
I'm fix this by sending a `affine-cloud-workspace-changed` event via `BroadcastChannel`.

packages/frontend/workspace/src/manager.ts
EYHN: Good question, Right now, we update React using some onChanged events.

Jotai isn't really a good fit in this case. To handle multiple instances, the atom needs to be part of the class.

Also, the atom doesn't hold data by itself.

For it to work, the class needs a store. But it's a bit odd for a class to have a Jotai store or get a Jotai store from global.

I think we need an atom that comes with its own store.

packages/frontend/workspace/src/list/index.ts
EYHN: https://github.com/toeverything/AFFiNE/blob/4470eb39ccf53e0b8735fad7f2c783af4ada74ea/packages/frontend/workspace/src/list/index.ts#L227

In fact, WorkspaceList subscribes to the workspace list update from the provider. How to implement subscription updates is implemented by the provider.

Currently, both `local` and `cloud` providers listen to events from other tabs to rescan the list, so if the user manually created a workspace, it will be automatically sync between tabs.

But there is currently no way to know if the workspace list changes come from the server.

If you want to set up a scheduled refresh, I have no idea about this interval time.

packages/frontend/workspace/src/manager.ts
pengx17: One issue is how to make a top-down data flow pattern to let the biz modules read/write the properties, e.g., a page's meta is derived all the way from the root manager.

Regarding the implementation, I am fine to have Jotai but it seems the class itself will have little control over the data held by Jotai; in our case a simple `useSyncExternalStore` could bridge global state stored in manager with hooks that can be consumed by the react side.
Another solution may be that reversing the relation between Jotai and the manager. Instead of using manager to control atoms, but define atoms outside of the manager that consumes the manager's properties and listen to change events.

What do you think?

packages/frontend/workspace/src/list/index.ts
pengx17: How about revalidate the workspace list when opening the list like SWR?

packages/frontend/hooks/src/use-workspace-blob.ts
pengx17: Yeah, maybe not relevant to the workspace manger PR. Can revisit it later.

packages/frontend/workspace/src/manager.ts
EYHN: Defining jotai outside of the manager may result in
* With one more jotai layer, it is difficult to define how abstract this layer is and what logic it should contain.
* A class may have multiple instances, and the jotai layer also needs to be changed accordingly.

I think any observable data type will work, we will decide in future refactoring.

packages/frontend/workspace/src/list/index.ts
EYHN: I can add a revalidate function

pengx17 (approved):

joooye34 (approved):

PR #5105, opened 2023-11-28
fix(core): implement editor timeout and report error from boundary

joooye34:
1. Implement editor root timeout check
2. Init sentry to report data
3. Use sentry error boundary to report unexpected error from react boundary

image
image


Brooooooklyn (approved):

packages/frontend/component/src/components/block-suite-editor/index.tsx
```diff
@@ -77,6 +74,52 @@ const useBlockElementById = (
...
+        expectPageId: page.id,
+        expectGuid: page.spaceDoc.guid,
+        spaceVectors,
+      })
+    );
+  }
+}
+
+const PAGE_ROOT_KEY = Symbol('PAGE_ROOT');
+function loadPageRoot(page: Page) {
```
EYHN: this is a react hook function, should be named as `useLoadPageRoot`

packages/frontend/component/src/components/block-suite-editor/index.tsx
```diff
@@ -77,6 +74,52 @@ const useBlockElementById = (
...
+          resolve();
+        });
+        window.setTimeout(() => {
+          disposable.dispose();
+          reject(new NoPageRootError(page));
+        }, 10 * 1000);
+      });
+      Reflect.set(page, PAGE_ROOT_KEY, root$);
+    }
+    use(root$);
```
EYHN, line 119: react hook cannot be used in if

packages/frontend/component/src/components/block-suite-editor/index.tsx
joooye34: it is not a hook, it is a suspense load function.

packages/frontend/component/src/components/block-suite-editor/index.tsx
joooye34: nope, suspense load function is different with hook function

Brooooooklyn (approved):

PR #5132, opened 2023-11-29
feat(core): support sidebar page item dnd

JimmFly:
Added the ability to drag page items from the `all pages` view to the sidebar, including `favourites,` `collection` and `trash`. Page items in `favourites` and `collection` can also be dragged between each other. However, linked subpages cannot be dragged.

Additionally, an operation menu and ‘add’ button have been provided for the sidebar’s page items, enabling the addition of a subpage, renaming, deletion or removal from the sidebar.

On the code front, the `useSidebarDrag` hooks have been implemented for consolidating drag events. The functions `getDragItemId` and `getDropItemId` have been created, and they accept type and ID to obtain itemId.

https://github.com/toeverything/AFFiNE/assets/102217452/d06bac18-3c28-41c9-a7d4-72de955d7b11


JimmFly:
* canary
  * **PR #5132** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5132?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5132?utm_source=stack-comment).

packages/frontend/component/src/components/rename-modal/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+        onOpenChange: onOpenChange,
+      }}
+      contentOptions={{
+        side: 'left',
+        onPointerDownOutside: handleRename,
+        sideOffset: -12,
+      }}
+      items={
+        <Input
+          ref={ref => {
```
pengx17: Please use callback (or perhaps add autofocus prop for Input)

packages/frontend/core/src/components/blocksuite/block-suite-page-list/utils.tsx
```diff
@@ -61,15 +63,43 @@ export const usePageHelper = (blockSuiteWorkspace: BlockSuiteWorkspace) => {
...
+            },
+          },
+        ]);
+        const [frame] = parentPage.getBlockByFlavour('affine:note');
+        frame && parentPage.addBlock('affine:paragraph', { text }, frame.id);
+        setPageMeta(page.id, {});
+      }
+    },
+    [blockSuiteWorkspace, createPageAndOpen, setPageMeta]
+  );
```
pengx17, line 90: Please move this function to a helper method. Maybe into `usePageHelper`?

packages/frontend/core/src/components/pure/workspace-slider-bar/collections/collections-list.tsx
```diff
@@ -81,60 +84,66 @@ const CollectionRenderer = ({
...
+          <div
+            onClick={stopPropagation}
+            style={{ display: 'flex', alignItems: 'center' }}
+          >
             <CollectionOperations
               info={info}
               collection={collection}
               setting={setting}
               config={config}
+              openRenameModal={() => setOpen(true)}
```
pengx17: useCallback please

packages/frontend/core/src/hooks/affine/use-sidebar-drag.ts
```diff
@@ -0,0 +1,128 @@
...
+  SidebarTrash = 'sidebar-trash',
+  SidebarFavorites = 'sidebar-favorites',
+}
+
+export enum DragPrefix {
+  PageListItem = 'page-list-item-title-',
+  FavouriteListItem = 'favourite-list-item-',
+  CollectionListItem = 'collection-list-item-',
+  CollectionListPageItem = 'collection-list-page-item-',
+}
```
pengx17: May not need enum this way. You are using enum as a key -> string mapping, but enum's value should not be used plain strings.
In this case it might be even better to use plain const objects.

Also, better to expose function like `function getDragItemId(type: 'collection' | 'page', id: string): string;` instead.

packages/frontend/core/src/layouts/workspace-layout.tsx
```diff
@@ -223,27 +216,30 @@ export const WorkspaceLayoutInner = ({
...
+      processRemoveDrag(e);
     },
-    [moveToTrash, t]
+    [
+      processCollectionsDrag,
+      processFavouritesDrag,
+      processMoveToTrashDrag,
+      processRemoveDrag,
+    ]
   );
```
pengx17: I think this is the only thing that `useSidebarDrag` should return.

tests/affine-local/e2e/drag-page.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+  const favourites = page.getByTestId('favourites');
+  await dragTo(
+    page,
+    page.locator(`[role="button"]:has-text("${title}")`),
+    favourites
+  );
+  const favouritePage = page.getByTestId(`favourite-page-${pageId}`);
+  expect(favouritePage).not.toBeUndefined();
+
+  // drag to collections
```
pengx17: Split the test into different tests

tests/affine-local/e2e/local-first-collections-items.spec.ts
```diff
@@ -100,13 +98,10 @@ test('edit collection', async ({ page }) => {
...
   await first.getByTestId('collection-options').click();
-  const editCollection = page
-    .getByTestId('collection-option')
-    .getByText('Rename');
+  const editCollection = page.getByText('Rename');
   await editCollection.click();
-  const title = page.getByTestId('input-collection-title');
-  await title.fill('123');
-  await page.getByTestId('save-collection').click();
+  await page.keyboard.type('123', { delay: 50 });
```
pengx17: why replace fill with type + delay?

pengx17: seems [playwright docs marks `type` deprecated and recommends using `pressSequentially`](https://playwright.dev/docs/api/class-locator#locator-press-sequentially:~:text=Added%20in%3A%20v1.14-,DEPRECATED,handling%20on%20the%20page%20%2D%20in%20this%20case%20use%20locator.pressSequentially().,-Focuses%20the%20element)

pengx17 (commented):
Overall looks great but I think you may need to list all features you have done in the summary, and also list what features that are not included.

packages/frontend/component/src/components/rename-modal/index.tsx
JimmFly: Autofocus doesn't work and I'm not sure how to fix it.

pengx17: input's autofocus maybe the web apis native autofocus attribute, but I guess it will not work most of the time. Maybe you could move the autofocus fix into input directly.

packages/frontend/core/src/components/blocksuite/block-suite-page-list/utils.tsx
JimmFly: Hmm, this function seems to be in `usePageHelper` now?

packages/frontend/core/src/layouts/workspace-layout.tsx
```diff
@@ -223,27 +216,13 @@ export const WorkspaceLayoutInner = ({
     })
   );
 
-  const { removeToTrash: moveToTrash } = useBlockSuiteMetaHelper(
-    currentWorkspace.blockSuiteWorkspace
-  );
+  const processDrag = useSidebarDrag();
```
pengx17: just rename this to `handleDragEnd` and remove line 211~225

pengx17:
In the current design the main hook `useSidebarDrag ` in this PR is not only for sidebar, but for all drag end event in the workspace layout. IMO this hook will need some refactor later to support more kinds of DND from/to pairs. some ideas

- using a unique abstraction for handling all types of drag end, and return a boolean to mark if it is being handled in the current handler. then have an array of handlers to store the handlers. when handling one drag event, iterate the array and pass the event to each of the handler. If any of the handler return true, then the rest of the handlers will get skipped
- do not bind addToCollection in collection-list, but put it in this universal hook instead
- `processDrag` is good but it seems it lacks the ability to handle all possible dnd cases, but only for dragging page target for now

pengx17 (approved):
We can revisit the implementation later

PR #5138, opened 2023-11-30
chore: bump blocksuite

Flrande:


packages/frontend/hooks/src/use-block-suite-page-references.ts
```diff
@@ -5,12 +5,9 @@ import { useBlockSuiteWorkspacePage } from './use-block-suite-workspace-page';
...
 const weakMap = new WeakMap<Page, Atom<string[]>>();
 function getPageReferences(page: Page): string[] {
-  // todo: is there a way to use page indexer to get all references?
-  return ['affine:paragraph', 'affine:list', 'affine:database']
-    .flatMap(f => page.getBlockByFlavour(f))
-    .flatMap(b => b.text?.toDelta())
-    .map(v => v?.attributes?.reference?.pageId)
-    .filter(Boolean);
+  return Object.values(
+    page?.workspace.indexer.backlink.linkIndexMap[page.id] ?? {}
```
EYHN:
```suggestion
    page.workspace.indexer.backlink.linkIndexMap[page.id] ?? {}
```

packages/frontend/hooks/src/use-block-suite-page-preview.ts
```diff
@@ -60,7 +60,7 @@ export function useBlockSuitePagePreview(page: Page | null): Atom<string> {
         page.slots.ready.on(() => {
           set(getPagePreviewText(page));
         }),
-        page.slots.yUpdated.on(() => {
+        page.slots.blockUpdated.on(() => {
```
EYHN, line 63: so block content update will trigger `blockUpdated` event?

Flrande: I think the preview doesn't need to be updated so promptly. If any changes in content will update the preview, I think it's easy to have performance issues.

joooye34 (approved):

joooye34:
Do we have a changelog between those two versions?

PR #5143, opened 2023-11-30
fix(infra): compatibility logic follow blocksuite

joooye34:


packages/frontend/core/src/pages/workspace/index.tsx
```diff
@@ -75,7 +75,7 @@ export const Component = (): ReactElement => {
 
   const migration = useLoaderData() as MigrationPoint | undefined;
   return (
-    <AffineErrorBoundary height="100vh">
+    <AffineErrorBoundary key={params.workspaceId} height="100vh">
```
pengx17, line 78: Why add this prop?

pengx17 (approved):

packages/frontend/core/src/pages/workspace/index.tsx
joooye34: I found out the error state can't rerender correctly sometimes, then I added this key to force rerender.

PR #5147, opened 2023-12-01
fix(core): rerender error boundary when route change and improve sentry report

joooye34:


packages/frontend/core/src/bootstrap/setup.ts
```diff
@@ -59,7 +59,7 @@ export async function setup(store: ReturnType<typeof createStore>) {
   performanceSetupLogger.info('setup global');
   setupGlobal();
 
-  if (window.SENTRY_RELEASE) {
+  if (window.SENTRY_RELEASE || environment.isDebug) {
```
Brooooooklyn, line 62:
```suggestion
  if (window.SENTRY_RELEASE || !environment.isDebug) {
```

Brooooooklyn (approved):

PR #5169, opened 2023-12-03
feat(component): new right sidebar

pengx17:
Refactor AFFiNE layout to support new right sidebar.

The new layout:
image

**Highlights:**
- new sidebar UI/UX
- favoring top-down UI components that are composed by basic building blocks in each route, instead of creating universal component like `WorkspaceHeader` that renders every possible cases (which I think is really hard to maintain)
- remove plugin based solution

**Pros/cons for current plugin-based solution:**

The current solution is somewhat a Dependency Injection (DI) approach, where the layout is defined at the top and UI items can be injected using Jotai atom slots.
This approach works well if we want a fully configurable system with everything being handled by plugins. It provides flexibility for custom extensions.
However, this solution is more suitable for single-page applications where the UI is completely controlled by configuration. It becomes challenging to achieve an optimized and visually appealing UI that remains under our control. An example of such a scenario would be a customizable dashboard like Grafana.
Another drawback of the existing solution is that we need to use Jotai and hooks to access context values, resulting in an unclear data flow within the component hierarchy.

**Alternatively, our approach in this PR** provides layout building blocks such as headers and sidebars, which can then be composed in individual route components. The good is that we have cleaner biz component instead of vague all-in-one layout component (like `<WorkspaceHeader />`).

**Issues of the implementation in this PR:**
Some UI layouts that that seems to be defined at the root layout are now defined in individual route component instead.
New 3-col layout component like the right sidebar still needs some abstraction and they are right now just for the detail editor only.


pengx17:
* canary
  * **PR #5169** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5169?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5169?utm_source=stack-comment).

packages/frontend/core/src/components/affine/awareness/index.tsx
```diff
@@ -0,0 +1,45 @@
+import { Suspense, useEffect } from 'react';
+
+import { useCurrentLoginStatus } from '../../../hooks/affine/use-current-login-status';
+import { useCurrentUser } from '../../../hooks/affine/use-current-user';
+import { useCurrentWorkspace } from '../../../hooks/current/use-current-workspace';
+
+const SyncAwarenessInnerLoggedIn = () => {
```
joooye34, line 7: It is so weird to use component to run effect, what is the benefit？

packages/frontend/core/src/components/pure/trash-page-footer/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+  assertExists(pageMeta);
+  const t = useAFFiNEI18N();
+  const { appSettings } = useAppSettingHelper();
+  const { jumpToSubPath } = useNavigateHelper();
+  const { restoreFromTrash } = useBlockSuiteMetaHelper(blockSuiteWorkspace);
+  const restoreRef = useRef(null);
+  const deleteRef = useRef(null);
+  const hintTextRef = useRef(null);
+  const [open, setOpen] = useState(false);
+  const hintText =
```
joooye34: i18n @JimmFly

packages/frontend/core/src/components/pure/trash-page-footer/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+          content={t['com.affine.trashOperation.restoreIt']()}
+          portalOptions={{
+            container: restoreRef.current,
+          }}
+        >
+          <Button
+            ref={restoreRef}
+            data-testid="page-restore-button"
+            type="primary"
+            onClick={() => {
```
joooye34: useCallback？

packages/frontend/core/src/components/pure/trash-page-footer/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+      <ConfirmModal
+        title={t['com.affine.trashOperation.delete.title']()}
+        cancelText={t['com.affine.confirmModal.button.cancel']()}
+        description={t['com.affine.trashOperation.delete.description']()}
+        confirmButtonOptions={{
+          type: 'error',
+          children: t['com.affine.trashOperation.delete'](),
+        }}
+        open={open}
+        onConfirm={useCallback(() => {
```
joooye34: move callback declare ahead of vdom tree

packages/frontend/core/src/components/pure/trash-page-footer/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+      >
+        <div ref={hintTextRef} className={styles.deleteHintText}>
+          {hintText}
+        </div>
+      </Tooltip>
+
+      <div className={styles.group}>
+        <Tooltip
+          content={t['com.affine.trashOperation.restoreIt']()}
+          portalOptions={{
```
joooye34: when first render, the ref is null. I am curious why this ref is needed.

packages/frontend/core/src/hooks/affine/use-current-user.ts
```diff
@@ -106,5 +106,5 @@ export function useCurrentUser(): CheckedUser {
       hasPassword: user?.hasPassword ?? false,
       update,
     };
-  }, [user, update]);
+  }, [user.id, user.name, user.email, user.image, user?.hasPassword, update]);
```
joooye34: add comments to explain why should dep on every properties？

joooye34: btw, question optional check before hasPassword is unnecessary

packages/frontend/core/src/components/pure/header/index.tsx
```diff
@@ -5,56 +5,62 @@ import {
...
   right?: ReactNode;
   center?: ReactNode;
-  mainContainerAtom: Atom<HTMLDivElement | null>;
   bottomBorder?: boolean;
 }
 
 // The Header component is used to solve the following problems
 // 1. Manage layout issues independently of page or business logic
 // 2. Dynamic centered middle element (relative to the main-container), when the middle element is detected to collide with the two elements, the line wrapping process is performed
 export const Header = forwardRef<HTMLDivElement, HeaderPros>(function Header(
```
joooye34: Seems forwardRef is unnecessary, I didn't find anywhere using it.

packages/frontend/core/src/pages/workspace/all-page/all-page-filter.tsx
```diff
@@ -0,0 +1,62 @@
...
+          propertiesMeta={currentWorkspace.blockSuiteWorkspace.meta.properties}
+          value={setting.currentCollection.filterList}
+          onChange={onFilterChange}
+        />
+      </div>
+      <div>
+        {setting.currentCollection.filterList.length > 0 ? (
+          <SaveAsCollectionButton
+            onConfirm={saveToCollection}
+          ></SaveAsCollectionButton>
```
joooye34: close self？

packages/frontend/core/src/pages/workspace/detail-page/detail-page.tsx
```diff
@@ -0,0 +1,362 @@
...
+      panelHandle.expand();
+    } else {
+      panelHandle.collapse();
+    }
+  }, [sidebarState.isOpen]);
+
+  const enableAnimation = useEnableAnimation();
+
+  return (
+    <PanelGroup
```
joooye34, line 128: do we still need this for current？

packages/frontend/core/src/pages/workspace/detail-page/detail-page.tsx
```diff
@@ -0,0 +1,362 @@
...
+
+  const { setTemporaryFilter } = useCollectionManager(collectionsCRUDAtom);
+  const mode = useAtomValue(currentModeAtom);
+  const setPageMode = useSetAtom(setPageModeAtom);
+  useRegisterBlocksuiteEditorCommands(currentPageId, mode);
+
+  const onLoad = useCallback(
+    (page: Page, editor: EditorContainer) => {
+      try {
+        const surfaceBlock = page.getBlockByFlavour('affine:surface')[0];
```
joooye34, line 197: Comment a todo to me, let me improve this code later

packages/frontend/core/src/pages/workspace/detail-page/editor-sidebar/editor-sidebar.tsx
```diff
@@ -0,0 +1,16 @@
+import { useAtomValue } from 'jotai';
+
+import { editorSidebarStateAtom } from './atoms';
+import * as styles from './editor-sidebar.css';
+
+export const EditorSidebar = () => {
+  const sidebarState = useAtomValue(editorSidebarStateAtom);
+  const Component = sidebarState.activeExtension?.Component;
+
+  // do we need this?
```
joooye34, line 10: for the plugin？

packages/frontend/core/src/pages/workspace/detail-page/editor-sidebar/extensions/frame.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { assertExists } from '@blocksuite/global/utils';
+import { FrameIcon } from '@blocksuite/icons';
+import { useCallback, useRef } from 'react';
+
+import { useCurrentPage } from '../../../../../hooks/current/use-current-page';
+import type { EditorExtension } from '../types';
+import * as styles from './frame.css';
+
+// todo: should be side effect free
+registerTOCComponents(components => {
```
joooye34: could this move to setup？

packages/frontend/core/src/pages/workspace/detail-page/editor-sidebar/extensions/outline.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { assertExists } from '@blocksuite/global/utils';
+import { TocIcon } from '@blocksuite/icons';
+import { useCallback, useRef } from 'react';
+
+import { useCurrentPage } from '../../../../../hooks/current/use-current-page';
+import type { EditorExtension } from '../types';
+import * as styles from './outline.css';
+
+// todo: should be side effect free
+registerTOCComponents(components => {
```
joooye34: move to setup ?

packages/frontend/core/src/components/pure/trash-page-footer/index.tsx
pengx17: Oh, this file is moved and I haven't checked the implementation. Let me take a look

packages/frontend/core/src/components/affine/awareness/index.tsx
pengx17: This component is mostly moved from https://github.com/toeverything/AFFiNE/blob/7ec8e49b3ba7115c7e38d6c2cb97c9e1fdc76cb6/packages/frontend/core/src/adapters/cloud/ui.tsx#L35-L47 but I think the original implementation is flawed. The awareness should be initialized when logged in.

It is used as a component for conditional rendering, which will cause lint issues if using a hook

packages/frontend/core/src/components/pure/header/index.tsx
pengx17: I haven't changed this layout component yet, but I can remove the forward ref

packages/frontend/core/src/pages/workspace/detail-page/detail-page.tsx
pengx17: Yeah, I am still using react-resizable-panels for panel resizing

packages/frontend/core/src/pages/workspace/detail-page/editor-sidebar/editor-sidebar.tsx
pengx17: It is to make sure the component will be re-rendered on open.

packages/frontend/core/src/pages/workspace/detail-page/editor-sidebar/extensions/outline.tsx
pengx17: done

packages/frontend/core/src/components/pure/trash-page-footer/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  return (
+    <div
+      className={styles.deleteHintContainer}
+      data-has-background={!appSettings.clientBorder}
+    >
+      <Tooltip
+        content={hintText}
+        portalOptions={{
+          container: hintTextRef.current,
```
CatsJuice: Better to mount portal  under `body`?

pengx17: yeah, or perhaps we could make tooltip to attach to body by default?

packages/frontend/core/src/components/pure/trash-page-footer/index.tsx
pengx17: the ref is not uncessary. Removed.

packages/frontend/core/src/components/affine/awareness/index.tsx
joooye34: ok... no need to change, I just think maybe this is bad practice.

joooye34 (approved):

PR #5142, opened 2023-11-30
feat: struct type feature config

darkskygit:

PR #5087, opened 2023-11-28
feat: integrate new modules

darkskygit:

PR #5076, opened 2023-11-27
feat: user usage gql & test case improve

darkskygit:

PR #5075, opened 2023-11-27
feat: integrate user usage into apis

darkskygit:

PR #5178, opened 2023-12-04
feat: add cache for blob query

darkskygit:


darkskygit:
* canary
  * **PR #5074** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5074?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5075** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5075?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #5076** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5076?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #5087** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5087?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #5142** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5142?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
            * **PR #5178** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5178?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5178?utm_source=stack-comment).

packages/backend/server/src/modules/workspaces/resolver.ts
```diff
@@ -137,6 +139,7 @@ export class UpdateWorkspaceInput extends PickType(
 @UseGuards(CloudThrottlerGuard)
 @Auth()
 @Resolver(() => WorkspaceType)
+@UseInterceptors(CacheInterceptor)
```
forehalo: can we make it globally? and skip cache by default, only controllers/resolvers with `Cache` decorator go through that logic

Brooooooklyn (changes requested):
Really don't like the idea about in-memory cache, need a complex strategy to make it valid/invalid without logic bug. Besides that, we also need a complex strategy to prevent OOM with in memory cache

Brooooooklyn:
I think cache it in redis like kv outside our app is a better solution, since we can flush the cache independently if states are unexpected or cache is full.

darkskygit:
can 

> Really don't like the idea about in-memory cache, need a complex strategy to make it valid/invalid without logic bug. Besides that, we also need a complex strategy to prevent OOM with in memory cache

i consider reuse global cache module that can switch between local/redis cache, local caching is useful for our self-hosted version

darkskygit:
> Really don't like the idea about in-memory cache, need a complex strategy to make it valid/invalid without logic bug. Besides that, we also need a complex strategy to prevent OOM with in memory cache


The current implementation can automatically switch to RedisCache according to configuration:

https://github.com/toeverything/AFFiNE/blob/94a1280768a03410b61f240f212b727cfb1da94b/packages/backend/server/src/cache/index.ts#L8-L16

Brooooooklyn (approved):

PR #5180, opened 2023-12-04
fix(core): error page 500 status dark mode

### LINKED ISSUE #5155: Error page dose not fit dark mode
### What happened?

image

Follow design system https://www.figma.com/file/UTTtfVCcStw1dKUg0gCSPo/%F0%9F%93%9A-Design-System?type=design&node-id=11815-11510&mode=design&t=x0CJ7aVuA2jYd4IX-0

### Distribution version

macOS x64 (Intel)

Chrome

### Relevant log output

_No response_

### Anything else?

_No response_

joooye34:
close #5155


CatsJuice (commented):
I recommend use single SVG for both dark and light mode, and use css variable for `fill` attribute.

joooye34:
> I recommend use single SVG for both dark and light mode, and use css variable for `fill` attribute.
> 
> 

Sometimes, the image couldn't use one single color to change style between dark and light, and image usage has better render performance, we should support image usage, not all use svg component.

pengx17 (approved):

pengx17 (approved):

pengx17:
I think if in the current case the SVG can be changed via variables, it should be fine, unless the SVG contain some performance issues like `filter` and `animation`.
Also for better perf we may want always `memo` or `innerHTML` to get rid of v-dom diff.

joooye34:
> I think if in the current case the SVG can be changed via variables, it should be fine, unless the SVG contain some performance issues like `filter` and `animation`. Also for better perf we may want always `memo` or `innerHTML` to get rid of v-dom diff.

@pengx17 memo is not good, maybe use innerHTML, we should design a usage as image, not all svg used as component.

CatsJuice:
BTW, use svg as `img` or css `background` may have compatibility issues in Safari where, when SVG uses special syntax(e.g. `mask`), the edges of the shape will get blurry.

Check [this site](https://safari-svg-mask-compatibility.vercel.app/) to compare  Safari with Chrome.

CleanShot 2023-12-05 at 18 24 25@2x

PR #5194, opened 2023-12-05
feat: bump blocksuite

doodlewind:
Upstream: https://github.com/toeverything/blocksuite/pull/5570, also temporally disabled block hub entry due to the feature conflict after migrating it as widget.

Change history: https://github.com/toeverything/blocksuite/compare/0f858b95...03964196


doodlewind:
Convert to draft since there are BlockHub breaking change in toeverything/blocksuite#5549 that should be adapted. The plan is to make it a `Fragment` in the `@blocksuite/preset` package.

Flrande (approved):

PR #5207, opened 2023-12-05
feat(core): allow core modules to import itself

pengx17:


pengx17:
* canary
  * **PR #5207** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5207?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5207?utm_source=stack-comment).

packages/frontend/core/package.json
```diff
@@ -13,7 +13,7 @@
     "./router": "./src/router.ts",
     "./bootstrap/setup": "./src/bootstrap/setup.ts",
     "./bootstrap/register-plugins": "./src/bootstrap/register-plugins.ts",
-    "./components/pure/*": "./src/components/pure/*/index.tsx"
+    "./*": "./src/*"
```
Brooooooklyn, line 16: Use the [`imports field`](https://nodejs.org/api/packages.html#imports) to implement this feature. If you use the exports field, the reference is implemented through the symlink in node_modules. We should not rely on this behavior.

Brooooooklyn (changes requested):
need to move self import map into `imports` field

packages/frontend/core/package.json
pengx17: I also noticed this field in `package.json`. 

I tried it but it seems there are some issues using `imports`:
- the field defined in `imports` must start with `#`. e.g., `#core`
- the files in `imports` should also be listed in `exports` for self-reference. Maybe it is because `exports` filed is used?

So we need to define some thing like this in package.json:

```json
  "imports": {
    "#core/*": "./src/*"
  },
  "exports": {
    "./*": "./src/*"
  }
```

and in ts
```ts
import { usePageHelper } from '#core/components/blocksuite/block-suite-page-list/utils';
```

---

Also it seems [self-referencing is a valid use case](https://nodejs.org/api/packages.html#self-referencing-a-package-using-its-name), if we do not care about how link works in package manager.

Brooooooklyn (approved):

PR #5238, opened 2023-12-08
fix(component): theme flickering

pengx17:
Add color-scheme to html so that the web will use system scheme when scripts has not being loaded;
Use `(prefers-color-scheme: dark)` so that the css vars do not defer load based on `data-theme`, which is set by next-themes (which is deferred).

### DIFF
--- a/packages/frontend/component/src/theme/theme.css.ts
+++ b/packages/frontend/component/src/theme/theme.css.ts
@@ -11,8 +11,12 @@ globalStyle('html', {
   vars: lightCssVariables,
 });
 
-globalStyle('html[data-theme="dark"]', {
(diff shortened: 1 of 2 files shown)


pengx17:
* canary
  * **PR #5238** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5238?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5238?utm_source=stack-comment).

CatsJuice (approved):

PR #5274, opened 2023-12-12
fix(core): polling to search in cmdk

EYHN:
This is a temporary solution until https://github.com/toeverything/blocksuite/issues/5668 be solved.

### DIFF
--- a/packages/frontend/core/src/components/pure/cmdk/data.tsx
+++ b/packages/frontend/core/src/components/pure/cmdk/data.tsx
@@ -25,7 +25,7 @@ import {
 } from '@toeverything/infra/command';
 import { atom, useAtomValue } from 'jotai';
 import { groupBy } from 'lodash-es';
-import { useCallback, useMemo } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 
 import {
   openQuickSearchModalAtom,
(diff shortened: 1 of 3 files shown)


EYHN:
* canary
  * **PR #5274** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5274?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5274?utm_source=stack-comment).

packages/frontend/core/src/components/pure/cmdk/data.tsx
```diff
@@ -216,7 +224,25 @@ export const usePageCommands = () => {
   const navigationHelper = useNavigateHelper();
   const t = useAFFiNEI18N();
 
+  const [searchTime, setSearchTime] = useState<number>(0);
+
+  // HACK: blocksuite indexer is async,
+  // so we need to re-search after it has been updated
+  useEffect(() => {
```
pengx17, line 231: I think making the `query` a debounced value could also implicitly solve the issue.
Also we should make query debounced to improve the searching performance.

EYHN: The indexer may delay by 1 second,
or the content may not have been synced, which may take longer.

debounce cannot solve this, this is a temporary solution until https://github.com/toeverything/blocksuite/issues/5668 be solved.

pengx17: Oh I just noticed that this function is essentially a setInterval.

pengx17 (approved):

PR #5295, opened 2023-12-13
feat: only follow `serverUrlPrefix` at redirect to client

darkskygit:
hotfix for insider domain migration


Brooooooklyn (approved):

packages/frontend/core/src/components/affine/share-page-modal/share-menu/use-share-url.ts
```diff
@@ -20,7 +20,7 @@ export const generateUrl = ({
   // or https://affine.app/share/123/456?mode=edgeless
 
   const url = new URL(
-    `${runtimeConfig.serverUrlPrefix}/${urlType}/${workspaceId}/${pageId}`
+    `${location.protocol}//${location.hostname}/${urlType}/${workspaceId}/${pageId}`
```
forehalo: `port` maybe

PR #5306, opened 2023-12-14
fix(core): page mode switch sometimes not working

pengx17:
Should not pass inline object without memo into `InternalLottie`.
https://github.com/toeverything/AFFiNE/blob/cdc96876b0a764ccf4f8ae9b7877ba8ae6774acb/packages/frontend/component/src/components/internal-lottie/index.tsx#L77

In the detail page when during syncing on the cloud, the detail page will be re-rendered constantly because of `useCurrentSyncEngineStatus` hook, which will then cause `PageSwitchItem` to re-render and forcing the internal lottie state to reset. As a result the click event may not be captured somehow.

### DIFF
--- a/packages/frontend/core/src/components/blocksuite/block-suite-mode-switch/switch-items.tsx
+++ b/packages/frontend/core/src/components/blocksuite/block-suite-mode-switch/switch-items.tsx
@@ -45,21 +45,30 @@ const HoverAnimateController = ({
   );
 };
 
+const pageLottieOptions = {
+  loop: false,
+  autoplay: false,
+  animationData: pageHover,
+  rendererSettings: {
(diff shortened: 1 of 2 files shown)


pengx17:
* `canary`
  * **PR #5306** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5306?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5306?utm_source=stack-comment).

CatsJuice (approved):

PR #5345, opened 2023-12-20
feat(electron): handle onboarding-window moved between screens

CatsJuice:
- when onboarding window is moved to another screen, resize it to fit screen.
- set window size to fullscreen on macOS(including dock area)
- disable round corner
- update hacked window `background` to fit design

### DIFF
--- a/packages/frontend/electron/src/main/onboarding.ts
+++ b/packages/frontend/electron/src/main/onboarding.ts
@@ -1,9 +1,15 @@
-import { BrowserWindow, screen } from 'electron';
+import { BrowserWindow, type Display, screen } from 'electron';
 import { join } from 'path';
 
+import { isMacOS } from '../shared/utils';
 import { mainWindowOrigin } from './constants';
 // import { getExposedMeta } from './exposed';
 import { logger } from './logger';
+
+const getScreenSize = (display: Display) => {
+  const { width, height } = isMacOS() ? display.bounds : display.workArea;
+  return { width, height };
+};
 
 // todo: not all window need all of the exposed meta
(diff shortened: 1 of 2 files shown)


CatsJuice:
* `canary`
  * **PR #5345** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5345?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5345?utm_source=stack-comment).

joooye34 (approved):

packages/frontend/electron/src/main/onboarding.ts
```diff
@@ -62,6 +69,20 @@ async function createOnboardingWindow(additionalArguments: string[]) {
...
   });
 
+  // When moved to another screen, resize to fit the screen
+  browserWindow.on('moved', () => {
+    const position = browserWindow.getPosition();
+    const size = browserWindow.getSize();
+    const currentDisplay = screen.getDisplayNearestPoint({
+      x: position[0] + size[0] / 2,
+      y: position[1] + size[1] / 2,
+    });
```
pengx17: do we need to use the same code for initial size/pos as well?

CatsJuice: Yes I think, it's uncertain which screen the window will open on.

pengx17 (approved):

joooye34 (approved):

PR #5358, opened 2023-12-20
ratelimitechange

AbhishekMaddheshiya919293:


darkskygit:
hi, we already have rate limit module: https://github.com/toeverything/AFFiNE/blob/aeec68b0d775364be3a00df27c550d43686d7427/packages/backend/server/src/throttler.ts#L50-L84

If you want to modify the rate limiter related configuration, please consider submitting an issue to discuss its feasibility first, because it involves server security issues.

PR #5366, opened 2023-12-21
feat: add useBlocksuiteEditor hooks

JimmFly:


JimmFly:
* `canary`
  * **PR #5366** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5366?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5366?utm_source=stack-comment).

EYHN (approved):

CatsJuice (approved):

packages/frontend/hooks/src/use-block-suite-editor.ts
```diff
@@ -0,0 +1,30 @@
+import type { AffineEditorContainer } from '@blocksuite/presets';
+import { atom, useAtom } from 'jotai';
+import { useCallback, useMemo } from 'react';
+
+export const editorContainerAtom = atom<{
```
pengx17: I think if you expose hooks, there is no point of exposing the atom any longer.
Also the API does not consider the case that two editors share the same pageId

packages/frontend/hooks/src/use-block-suite-editor.ts
```diff
@@ -0,0 +1,30 @@
...
+        [pageId]: editorContainer,
+      }));
+    },
+    [setEditors, pageId]
+  );
+
+  return useMemo(() => {
+    return {
+      editor,
+      setEditorContainer,
```
pengx17: the name of the return value may use `useState` pattern that returns a 2-tuple.
also the name of the first param should be `editorContainer`

packages/frontend/component/src/components/block-suite-editor/index.tsx
```diff
@@ -151,7 +148,7 @@ const BlockSuiteEditorImpl = ({
 }: EditorProps): ReactElement => {
   usePageRoot(page);
 
-  const [, setEditorContainer] = useAtom(editorContainerAtom);
+  const { setEditorContainer } = useBlocksuiteEditor(page.id);
```
pengx17: Should remove the editor instance when unmount

pengx17 (changes requested):
Please resolve my comments

pengx17:
So the new impl does not support multiple editors?

JimmFly:
> So the new impl does not support multiple editors?

Apologies, it seems I’ve misconstrued your meaning. I interpreted the comment as suggesting that the editor atom didn’t require correspondence with the pageid. @pengx17 

>Also the API does not consider the case that two editors share the same pageId

pengx17 (approved):

PR #5372, opened 2023-12-21
fix: websocket prefix

darkskygit:


joooye34 (approved):

packages/frontend/workspace/src/utils/affine-io.ts
```diff
@@ -2,21 +2,25 @@ import { Manager } from 'socket.io-client';
 
 let ioManager: Manager | null = null;
 
+function getBaseUrl(): string {
+  if (environment.isDesktop) {
+    return runtimeConfig.serverUrlPrefix;
+  }
+  const { protocol, hostname, port } = window.location;
+  return `${protocol === 'https:' ? 'wss' : 'ws'}://${hostname}${
```
EYHN, line 10: should we use wss in desktop version?

packages/frontend/workspace/src/utils/server-config.ts
```diff
@@ -0,0 +1,45 @@
+import { type QueryResponse, serverConfigQuery } from '@affine/graphql';
```
EYHN: delete this file?

joooye34 (approved):

PR #5390, opened 2023-12-25
fix(server): blob id may contains slash

forehalo:


forehalo:
* `canary`
  * **PR #5390** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5390?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5390?utm_source=stack-comment).

darkskygit:
blob ids is url safe and will not contains slash
https://github.com/toeverything/blocksuite/blob/f637299f4ab4e685d3a5bb93dd513536291de126/packages/store/src/persistence/blob/utils.ts#L3-L8

PR #5407, opened 2023-12-26
feat: bump blocksuite

doodlewind:
Change history: https://github.com/toeverything/blocksuite/compare/e3abcbb...3822a25

PR #5412, opened 2023-12-26
refactor(core): onboarding using new transformer api

pengx17:
Use new `transformer` to import onboarding json templates.

The json files are generated via this gist
https://gist.github.com/pengx17/ef92c305ac23123803a1a6a20e31f822

Not using the all-in-one `ZipTransformer` to import onboarding via a zip file. 
1. The main concerns is that we still need to serve the blob resources via CDN to reduce user's blob usage. Otherwise the user will get the onboarding images being uploaded to cloud server every time he creates a new workspace. In this PR we extracted parts of the code from `ZipTransformer` in blocksuite and mute some code for uploading blobs.
2. it maybe not necessary to use zip for loading snapshots.

This PR is a short term solution. whether or not to tune the transformer api design may need further discussions.

fix TOV-264


pengx17:
* `canary`
  * **PR #5412** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5412?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5412?utm_source=stack-comment).

pengx17:
blocks https://github.com/toeverything/AFFiNE/pull/5407

Saul-Mirone (approved):
The API call of transformers and middlewares looks good to me. I think we need to avoid hack the assets manager in the future. Maybe we can load the image in middleware.

packages/frontend/templates/package.json
```diff
@@ -1,11 +1,13 @@
...
   "private": true,
-  "sideEffect": false,
-  "exports": {
-    "./*.md": "./*.md",
-    "./v1/*.json": "./v1/*.json",
-    "./preloading.json": "./preloading.json"
+  "sideEffect": true,
+  "version": "0.11.0",
+  "scripts": {
+    "gen-templates": "node ./build.mjs"
```
joooye34: can we use postinstall to gen this file and ignore it, not commit to repo.

packages/frontend/templates/build.mjs
```diff
@@ -0,0 +1,26 @@
...
+        JSON.parse(
+          fs.readFileSync(path.resolve(__dirname, 'onboarding', file), 'utf8')
+        ),
+      ];
+    })
+);
+
+fs.writeFileSync(
+  path.resolve(__dirname, 'templates.gen.ts'),
+  `export const onboarding = ${JSON.stringify(content)};`
```
joooye34:
```suggestion
  `export const onboarding = ${JSON.stringify(content, null, 2)};`
```

let the generated file readable.

joooye34: Or, better code is importing json directly, not read all json content.

packages/frontend/templates/package.json
```diff
@@ -1,11 +1,13 @@
 {
   "name": "@affine/templates",
```
joooye34, line 2: we need a README.md to explain how to maintain templates.

tests/affine-local/e2e/all-page.spec.ts
```diff
@@ -85,7 +85,8 @@ test('allow creation of filters by favorite', async ({ page }) => {
   );
 });
 
-test('allow creation of filters by created time', async ({ page }) => {
+// new onboarding create pages that are all today
+test.skip('allow creation of filters by created time', async ({ page }) => {
```
joooye34: continue to skip or code forget to delete?

packages/common/infra/src/blocksuite/initialization/middleware.ts
```diff
@@ -0,0 +1,103 @@
+// TODO: remove this file after blocksuite exposed it
+
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
+import { assertExists } from '@blocksuite/global/utils';
+import type { JobMiddleware } from '@blocksuite/store';
+
+export const replaceIdMiddleware: JobMiddleware = ({ slots, workspace }) => {
```
joooye34, line 14: if we do not change the id, what problem will happen?

packages/frontend/templates/package.json
pengx17: I tried using postinstall but it seems the file does not get generated after `yarn install`. I will take a look

packages/common/infra/src/blocksuite/initialization/middleware.ts
pengx17: For now in idb all page docs will be stored in the same table with the key = pageid. This could be a disaster if we do not replace the ids.

joooye34: ok

packages/frontend/templates/package.json
pengx17: It seems yarn will only run postinstall when the dependency changed. So that's why I added the template to git.

packages/common/infra/src/blocksuite/initialization/middleware.ts
joooye34: image.png links are broken, maybe cause by id change ?

joooye34: image

packages/frontend/templates/build.mjs
pengx17: done

tests/affine-local/e2e/all-page.spec.ts
pengx17: 🤔 yeah, maybe I can just remove them

packages/common/infra/src/blocksuite/initialization/middleware.ts
pengx17: Thanks! This has been fixed. Also submitted an upstream PR:
https://github.com/toeverything/blocksuite/pull/5883

joooye34 (approved):

PR #5443, opened 2023-12-28
feat(electron): use release api to filter draft release

joooye34:


joooye34:
* `canary`
  * **PR #5439** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5439?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5443** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5443?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5443?utm_source=stack-comment).

packages/frontend/electron/src/main/updater/custom-github-provider.ts
```diff
@@ -176,6 +198,50 @@ export class CustomGitHubProvider extends BaseGitHubProvider<GithubUpdateInfo> {
     return `/${this.options.owner}/${this.options.repo}/releases`;
   }
 
+  /**
+   * Use release api to get latest version to filter draft version.
+   * But this api have low request limit, use this to help, not depend on it
```
pengx17: Maybe we could add a cf worker for fetching the latest releases from github to mitigate this limitation.

pengx17 (approved):

packages/frontend/electron/src/main/updater/custom-github-provider.ts
joooye34: Not sure how to do it. Github api has 60-times/1-hour limitation. We can get detail from rate_limit api https://api.github.com/rate_limit.

pengx17: for example, when requesting the cf worker will fetch & cache the latest versions from github; new requests within 5 min will use the cached results.

Also I think github api has a higher rate limit if we use a token instead of anonymous; and we could have a token pool if it still hit the limits.

joooye34: Yeah, I got it. Maybe not so complex, we can design a counter to use release API first, and fallback to feed API. If we get release info, the information is enough, no need to request feed again.

PR #5497, opened 2024-01-02
feat(infra): di container

EYHN:
docs: https://insider.affine.pro/share/055f9c4b-497a-43ec-a1c9-29d5baf184b9/N785YJ__oLMb2fUaLOv-k

### DIFF
--- a/packages/common/infra/src/di/__tests__/di.spec.ts
+++ b/packages/common/infra/src/di/__tests__/di.spec.ts
@@ -1,0 +1,357 @@
+import { describe, expect, test } from 'vitest';
+
+import {
+  CircularDependencyError,
+  createIdentifier,
+  createScope,
+  DuplicateServiceDefinitionError,
+  MissingDependencyError,
+  RecursionLimitError,
+  ServiceCollection,
+  ServiceNotFoundError,
+  ServiceProvider,
+} from '../';
+
+describe('di', () => {
+  test('basic', () => {
+    const serviceCollection = new ServiceCollection();
+    class TestService {
+      a = 'b';
+    }
+
+    serviceCollection.add(TestService);
+
+    const provider = serviceCollection.provider();
+    expect(provider.get(TestService)).toEqual({ a: 'b' });
+  });
+
+  test('size', () => {
+    const serviceCollection = new ServiceCollection();
+    class TestService {
+      a = 'b';
+    }
+
+    serviceCollection.add(TestService);
+
+    expect(serviceCollection.size).toEqual(1);
+  });
+
+  test('dependency', () => {
+    const serviceCollection = new ServiceCollection();
+
+    class A {
+      value = 'hello world';
+    }
+
+    class B {
+      constructor(public a: A) {}
+    }
+
+    class C {
+      constructor(public b: B) {}
+    }
+
+    serviceCollection.add(A).add(B, [A]).add(C, [B]);
+
+    const provider = serviceCollection.provider();
+
+    expect(provider.get(C).b.a.value).toEqual('hello world');
+  });
+
+  test('identifier', () => {
+    interface Animal {
+      name: string;
+    }
+    const Animal = createIdentifier<Animal>('Animal');
+
+    class Cat {
+      constructor() {}
+      name = 'cat';
+    }
+
+    class Zoo {
+      constructor(public animal: Animal) {}
+    }
+
+    const serviceCollection = new ServiceCollection();
+    serviceCollection.addImpl(Animal, Cat).add(Zoo, [Animal]);
+
+    const provider = serviceCollection.provider();
+    expect(provider.get(Zoo).animal.name).toEqual('cat');
+  });
+
+  test('variant', () => {
+    const serviceCollection = new ServiceCollection();
+
+    interface USB {
+      speed: number;
+    }
+
+    const USB = createIdentifier<USB>('USB');
+
+    class TypeA implements USB {
+      speed = 100;
+    }
+    class TypeC implements USB {
+      speed = 300;
+    }
+
+    class PC {
+      constructor(
+        public typeA: USB,
+        public ports: USB[]
+      ) {}
+    }
+
+    serviceCollection
+      .addImpl(USB('A'), TypeA)
+      .addImpl(USB('C'), TypeC)
+      .add(PC, [USB('A'), [USB]]);
+
+    const provider = serviceCollection.provider();
+    expect(provider.get(USB('A')).speed).toEqual(100);
+    expect(provider.get(USB('C')).speed).toEqual(300);
+    expect(provider.get(PC).typeA.speed).toEqual(100);
+    expect(provider.get(PC).ports.length).toEqual(2);
+  });
+
+  test('lazy initialization', () => {
+    const serviceCollection = new ServiceCollection();
+    interface Command {
+      shortcut: string;
+      callback: () => void;
+    }
+    const Command = createIdentifier<Command>('command');
+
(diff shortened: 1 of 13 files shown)


EYHN:
* `canary`
  * **PR #5562** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5562?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5595** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5595?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5497** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5497?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
      * **PR #5563** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5563?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #5564** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5564?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #5693** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5693?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
            * **PR #5617** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5617?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
              * **PR #5618** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5618?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
                * **PR #5565** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5565?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5497?utm_source=stack-comment).

packages/common/infra/src/di/core/collection.ts
```diff
@@ -0,0 +1,459 @@
...
+    string,
+    Map<string, Map<ServiceVariant, ServiceFactory>>
+  > = new Map();
+
+  /**
+   * Create an empty service collection.
+   *
+   * same as `new ServiceCollection()`
+   */
+  static get EMPTY() {
```
pengx17, line 112: it looks dangerous to me to return a new instance on access. Maybe cache the returned value?

packages/common/infra/src/di/core/provider.ts
```diff
@@ -0,0 +1,216 @@
...
+    super();
+    this.collection = collection.clone();
+    this.collection.addValue(ServiceProvider, this, {
+      scope: scope,
+      override: true,
+    });
+  }
+
+  getRaw(identifier: ServiceIdentifierValue, options?: ResolveOptions) {
+    const resolver = new ServiceResolver(this);
```
pengx17, line 205: I have a question: why add another `ServiceResolver` abstraction? It seems to me that `ServiceResolver` can be merged into BasicServiceProvider

pengx17 (approved):

packages/common/infra/src/di/core/collection.ts
```diff
@@ -0,0 +1,459 @@
...
+   * @example
+   *
+   * ```ts
+   * const ScopeA = createScope('a');
+   *
+   * services.scope(ScopeA).add(XXXService, ...);
+   * ```
+   */
+  scope = (scope: ServiceScope): ServiceCollectionEditor => {
+    this.currentScope = scope;
```
pengx17, line 409: It looks like scope might be misused. E.g.:

```ts
// a.ts
import { services } from './services'
services.scope('foo').add(xxx)

// then in b.ts
import { services } from './services'
services.add(yyy); // may or may not be scope 'foo', depending on import order
```

If a service collection ref being used in two places, there might be an issue that the scope is changed but it is not known to the others.

packages/common/infra/src/di/core/collection.ts
EYHN: What if someone modified it?

packages/common/infra/src/di/core/provider.ts
EYHN: We will create a ServiceResolver for each ServiceFactory.

see https://github.com/toeverything/AFFiNE/blob/fc2bcc8ad42c03c1ab1db9eb24550320773bdb28/packages/common/infra/src/di/core/provider.ts#L180-L183

The dependency depth and service dependency stack will be saved on ServiceResolver to detect circular dependencies.

EYHN: `BasicServiceProvider` is used to save instantiated services
`ServiceResolver` is used to instantiate services

But users don't need to care, users only know `ServiceProvider`

packages/common/infra/src/di/core/collection.ts
EYHN: In fact in your example we always get

```
foo:
  xxx
root:
  yyy
```

when you call `add` / `addImpl` / `scope` / `override` on `ServiceCollection`, a `ServiceCollectionEditor` will be created, and your `scope("foo")` actually only acts on this `Editor` and not on `ServiceCollection`.

https://github.com/toeverything/AFFiNE/blob/fc2bcc8ad42c03c1ab1db9eb24550320773bdb28/packages/common/infra/src/di/core/collection.ts#L132-L155

packages/common/infra/src/di/core/provider.ts
pengx17: makes sense now, thank you!

packages/common/infra/src/di/core/collection.ts
pengx17: I guess I fell into the immutable mindset now but the design of DI is built upon mutable structures.
Resolving now.

packages/common/infra/src/di/__tests__/di.spec.ts
```diff
@@ -0,0 +1,357 @@
...
+    const root = services.provider();
+    expect(root.get(System).appName).toEqual('affine');
+    expect(() => root.get(Workspace)).toThrowError(ServiceNotFoundError);
+
+    const workspace = services.provider(workspaceScope, root);
+    expect(workspace.get(Workspace).name).toEqual('workspace');
+    expect(workspace.get(System).appName).toEqual('affine');
+    expect(() => root.get(Page)).toThrowError(ServiceNotFoundError);
+
+    const page = services.provider(pageScope, workspace);
```
forehalo, line 249: since we explictly pass `workspace` as `page` scope's parent when creating it, is it possible to ignore the parent scope parameter when getting the provider?

packages/common/infra/src/di/__tests__/di.spec.ts
```diff
@@ -0,0 +1,357 @@
...
+    expect(() => provider.get(C)).toThrowError(CircularDependencyError);
+  });
+
+  test('duplicate service definition', () => {
+    const serviceCollection = new ServiceCollection();
+
+    class A {}
+
+    serviceCollection.add(A);
+    expect(() => serviceCollection.add(A)).toThrowError(
```
forehalo, line 313: what if overriding?

packages/common/infra/src/di/__tests__/di.spec.ts
EYHN: No, the workspace will have multiple instances, and it must be clear which instance the page is derived from.

PR #5527, opened 2024-01-06
fix: properly split dep and dev deps, fix phantom deps

lewisl9029:
Hi AFFiNE team! I've been working on adding [Reflame](https://reflame.app/)'s component testing product (you can think of it as a Chromatic alternative that gives you screenshot diffs within _single digit seconds_ of a push, instead of minutes) to the AFFiNE repo in a fork.

You can see the WIP results [here](https://github.com/lewisl9029/AFFiNE/pull/1/checks?check_run_id=20217535991), where we deploy, take screenshots, and diff 68 stories in ~7 seconds! :)

I've split this PR out of the larger WIP branch to make things easier to merge and review for y'all. This one only deals with a few issues I found in package.json dependencies:

- Some runtime dependencies were being specified as dev dependencies
- Some dev dependencies were being specified as runtime dependencies
- Some dependencies were used in packages without being specified in package.json (a.k.a. [phantom dependencies](https://rushjs.io/pages/advanced/phantom_deps/))

Reflame needs to be a lot stricter when it comes to these distinctions in order to install, build, and deploy the minimum set of deps possible for your app/tests to keep everything fast. It also doesn't hurt to have better hygine around these distinctions in general to avoid future headaches (especially around phantom deps, which have been known to cause really insidious issues), so I wanted to see if you were open to merging this separately.

Wdyt? Cheers!


packages/common/cmdk/package.json
```diff
@@ -8,5 +8,8 @@
     "react": "18.2.0",
     "react-dom": "18.2.0"
   },
+  "dependencies": {
+    "@radix-ui/react-dialog": "^1.0.5"
+  },
```
lewisl9029, line 13: This was used in https://github.com/toeverything/AFFiNE/blob/9650a5a6a1d1c7a144e101ba270f29f605c7ff5a/packages/common/cmdk/src/index.tsx#L3 without having been specified.

packages/common/env/package.json
```diff
@@ -7,8 +7,7 @@
     "@blocksuite/store": "0.11.0-nightly-202401020419-752a5b8",
     "react": "18.2.0",
     "react-dom": "18.2.0",
-    "vitest": "1.1.3",
-    "zod": "^3.22.4"
```
lewisl9029, line 11: This was being used at runtime even though it was specified as a dev dependency.

packages/common/env/package.json
```diff
@@ -21,10 +20,11 @@
   },
   "peerDependencies": {
     "@affine/templates": "workspace:*",
-    "@blocksuite/global": "0.0.0-20230409084303-221991d4-nightly"
```
lewisl9029, line 24: This was making it impossible for Reflame to resolve `@blocksuite/global` to a valid version, since every package that uses this has a strict dependency on version `0.11.0-nightly-202401020419-752a5b8`.

Reflame's npm package resolution algorithm is built on npm's [arborist](https://github.com/npm/cli/tree/latest/workspaces/arborist) library, which treats this as an error, while yarn only seems to output a warning.

packages/common/infra/package.json
```diff
@@ -18,7 +18,9 @@
     "@blocksuite/store": "0.11.0-nightly-202401020419-752a5b8",
     "jotai": "^2.5.1",
     "jotai-effect": "^0.2.3",
+    "nanoid": "^5.0.3",
     "tinykeys": "^2.1.0",
+    "yjs": "^13.6.10",
```
lewisl9029, line 23: These were both used at runtime.

packages/common/y-indexeddb/package.json
```diff
@@ -32,6 +32,7 @@
     }
   },
   "dependencies": {
+    "@blocksuite/global": "0.11.0-nightly-202401020419-752a5b8",
```
lewisl9029, line 35: This was used at runtime without being specified.

packages/common/y-provider/package.json
```diff
@@ -31,6 +31,7 @@
     "yjs": "^13.6.10"
   },
   "peerDependencies": {
+    "@blocksuite/global": "*",
```
lewisl9029, line 34: This was also used at runtime without being specified. Added it as a peer dep since it didn't have any direct runtime deps before.

packages/frontend/component/package.json
```diff
@@ -31,6 +31,8 @@
     "@emotion/react": "^11.11.1",
     "@emotion/server": "^11.11.0",
     "@emotion/styled": "^11.11.0",
+    "@mui/base": "5.0.0-beta.19",
+    "@mui/material": "^5.14.14",
```
lewisl9029, line 35: These were used at runtime.

packages/frontend/core/package.json
```diff
@@ -68,6 +68,7 @@
     "jotai-devtools": "^0.7.0",
     "jotai-effect": "^0.2.3",
     "lit": "^3.0.2",
+    "lodash-es": "^4.17.21",
```
lewisl9029, line 70: Also used but not specified.

tests/storybook/package.json
```diff
@@ -9,26 +9,16 @@
...
-    "concurrently": "^8.2.2",
-    "jest-mock": "^29.7.0",
+    "foxact": "^0.2.26",
+    "jotai": "^2.5.1",
+    "lodash-es": "^4.17.21",
     "nanoid": "^5.0.3",
-    "serve": "^14.2.1",
-    "storybook": "^7.5.3",
-    "storybook-dark-mode": "^3.0.1",
-    "wait-on": "^7.2.0"
```
lewisl9029, line 31: Moved all of these down to dev deps, since they weren't used at runtime.

tests/storybook/package.json
```diff
@@ -39,15 +29,28 @@
...
+    "@vitejs/plugin-react": "^4.2.0",
     "chromatic": "^10.0.0",
-    "foxact": "^0.2.26",
-    "jotai": "^2.5.1",
+    "concurrently": "^8.2.2",
+    "jest-mock": "^29.7.0",
     "react": "18.2.0",
     "react-dom": "18.2.0",
-    "react-router-dom": "^6.19.0",
-    "storybook-addon-react-router-v6": "^2.0.10"
```
lewisl9029, line 50: Moved these up into deps since they were used at runtime.

Brooooooklyn:
Thanks @lewisl9029 , let me fix the rest of the stuff on your branch

Brooooooklyn:
* `canary`
  * **#5527** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5527?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

 Join @lewisl9029 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/common/cmdk/package.json
Brooooooklyn: That's strange, why didn't lint detect this issue? Let me double check.

Brooooooklyn (approved):

packages/common/cmdk/package.json
lewisl9029: I think this is working currently because @radix-ui/react-dialog is specified in two other packages in the same workspace: https://github.com/toeverything/AFFiNE/blob/7c71c5457ae4e4cc98d487220e3b4614cc6dfe28/packages/frontend/core/package.json#L46
https://github.com/toeverything/AFFiNE/blob/7c71c5457ae4e4cc98d487220e3b4614cc6dfe28/packages/frontend/component/package.json#L37

So yarn hoists the package up to the root node_modules folder, which makes it resolvable by this package as well, even though the package itself doesn't specify it as a dependency. The node resolution algorithm just keeps looking up node_modules folders recursively until it reaches `/`, and doesn't actually care if it's in package.json or not. This is known as a "phantom dependency", here's a good writeup on the problems that might cause: https://rushjs.io/pages/advanced/phantom_deps/

I'm not aware of a lint rule that guards against phantom deps, but pnpm does guard against it by default (by symlinking only the specified deps for each package), and both npm and yarn have options you can use to opt-in to similar behavior:

- For npm it's [`--install-strategy=linked`](https://docs.npmjs.com/cli/v9/commands/npm-install#install-strategy).
- For yarn it's [`nodeLinker: pnpm`](https://yarnpkg.com/configuration/yarnrc#nodeLinker).

lewisl9029: Oh and I forgot to mention, Reflame of course will also guard against phantom deps. That's how I discovered these. :)

Brooooooklyn: we have  [`no-extraneous-dependencies`](https://github.com/toeverything/AFFiNE/blob/canary/.eslintrc.js#L271) rule for phantom deps. But it seems like the cmdk related code has disabled this rule, I will open another issue to record this problem.

> Oh and I forgot to mention, Reflame of course will also guard against phantom deps. That's how I discovered these. :)

Great!

PR #5535, opened 2024-01-08
feat(workspace): add blob and storage limit

JimmFly:
close TOV-343 AFF-508 TOV-461 TOV-460 TOV-419

Add `isOverCapacity ` status to detect if blob usage exceeds limits.
Add `onCapacityChange` and `onBlobSet` to monitor if the storage or blob exceeds the capacity limit.
Global modals `LocalQuotaModal` and `CloudQuotaModal` have been added, with the upload size of the blob being limited within the modal components.
The notification component has been adjusted, now you can pass in `action` click events and `actionLabel` .


JimmFly:
* `canary`
  * **PR #5535** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5535?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5535?utm_source=stack-comment).

packages/common/workspace/src/engine/blob.ts
```diff
@@ -12,6 +12,7 @@ const logger = new DebugLogger('affine:blob-engine');
  */
 export class BlobEngine {
   private abort: AbortController | null = null;
+  private isOverCapacity: boolean = false;
```
EYHN: You need an event to tell the outside about this change.

example: https://github.com/toeverything/AFFiNE/blob/93a2398101e3300828822d87144b0c559b4f4885/packages/common/workspace/src/engine/sync/engine.ts#L58https://github.com/toeverything/AFFiNE/blob/93a2398101e3300828822d87144b0c559b4f4885/packages/common/workspace/src/engine/sync/engine.ts#L58

JimmFly: Do the latest changes meet your expectations?

packages/frontend/core/src/components/pure/workspace-slider-bar/workspace-card/index.tsx
```diff
@@ -108,10 +152,18 @@ const useSyncEngineSyncProgress = () => {
         }
       )
     );
+    const disposableBlobSet = currentWorkspace.engine.blob.onBlobSet.on(
+      blobSetEventArgs => {
+        console.log('blob set', blobSetEventArgs);
```
EYHN: use debug logger ?

JimmFly: Alright, I'll change to a debug logger. It's currently a draft, so please review again once it's turned into a formal PR.

packages/backend/server/src/modules/quota/types.ts
```diff
@@ -42,13 +42,13 @@ export type Quota = z.infer<typeof QuotaSchema>;
 
 @ObjectType()
 export class QuotaQueryType {
-  @Field(() => Int)
+  @Field(() => Float)
```
EYHN: Maybe we can use custom scalar
https://the-guild.dev/graphql/scalars/docs/scalars/safe-int

darkskygit: replaced

packages/common/workspace/src/engine/blob.ts
```diff
@@ -169,6 +212,10 @@ export class BlobEngine {
 
     return Array.from(blobList);
   }
+
+  isStorageOverCapacity() {
```
joooye34: why not a public getter for isOverCapacity?

packages/common/workspace/src/engine/blob.ts
```diff
@@ -12,6 +19,16 @@ const logger = new DebugLogger('affine:blob-engine');
  */
 export class BlobEngine {
   private abort: AbortController | null = null;
+  private _isOverCapacity: boolean = false;
+  onCapacityChange = new Slot<boolean>();
+  onBlobSet = new Slot<BlobSetEventArgs>();
+
+  private set isOverCapacity(value: boolean) {
```
joooye34: cc @EYHN

packages/frontend/core/src/components/affine/quota-reached-modal/cloud-quota-modal.tsx
```diff
@@ -0,0 +1,99 @@
...
+    userQuota,
+    workspaceQuota.humanReadable.blobLimit,
+  ]);
+
+  useEffect(() => {
+    const disposable = currentWorkspace.engine.blob.onBlobSet.on(
+      blobSetEventArgs => {
+        if (!checkBlobSize(blobSetEventArgs.value)) {
+          setOpen(true);
+          blobSetEventArgs.updateShouldProceed(false);
```
joooye34: it is not a good idea to change the inner state through an event function

packages/common/workspace/src/engine/blob.ts
```diff
@@ -78,10 +95,16 @@ export class BlobEngine {
               await remote.set(key, data);
             }
           } catch (err) {
-            logger.error(
-              `error when sync ${key} from [${this.local.name}] to [${remote.name}]`,
-              err
-            );
+            const code = (err as any)[0].extensions.code;
```
EYHN:
```suggestion
            if (err instanceof GraphQLError) {
            const code = (err as any)[0].extensions.code;
```

packages/common/workspace/src/engine/blob.ts
```diff
@@ -122,6 +145,19 @@ export class BlobEngine {
...
+
+    const updateShouldProceed = (proceed: boolean) => {
+      shouldProceed = proceed;
+    };
+
+    this.onBlobSet.emit({ key, value, updateShouldProceed });
+
+    if (!shouldProceed) {
+      logger.error('blob over limit, abort set');
+      return key;
```
EYHN, line 153: should we throw a error, and maybe blocksuite will show error view?

packages/common/workspace/src/engine/blob.ts
```diff
@@ -131,7 +167,13 @@ export class BlobEngine {
         .filter(r => !r.readonly)
         .map(peer =>
           peer.set(key, value).catch(err => {
-            logger.error('error when upload to peer', err);
+            const code = err[0].extensions.code;
```
EYHN: check `err instanceof GraphQLError`

packages/common/workspace/src/engine/blob.ts
```diff
@@ -141,6 +183,7 @@ export class BlobEngine {
             `blob ${key} update finish, but some peers failed to update`
           );
         } else {
+          this.isOverCapacity = false;
```
EYHN: Since our current capacity is only increasing but not decreasing, it is best not to set false here?

If the capacity has been expanded, users can refresh the page or app.

packages/common/workspace/src/engine/blob.ts
EYHN: is same thing as isOverCapacity ?

packages/common/workspace/src/engine/blob.ts
```diff
@@ -12,6 +19,16 @@ const logger = new DebugLogger('affine:blob-engine');
  */
 export class BlobEngine {
   private abort: AbortController | null = null;
+  private _isOverCapacity: boolean = false;
+  onCapacityChange = new Slot<boolean>();
```
EYHN: In order to be consistent with other engines, maybe you can change this to

```ts
interface BlobStatus {
  isOverCapacity: boolean
}

_status: BlobStatus = { isOverCapacity: false };
onStatusChange = new Slot<BlobStatus>;
```

packages/frontend/core/src/components/affine/quota-reached-modal/cloud-quota-modal.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const workspaceQuota = useWorkspaceQuota(currentWorkspace.id);
+  const isOwner = useIsWorkspaceOwner(currentWorkspace.meta);
+  const userQuota = useUserQuota();
+  const isFreePlanOwner = useMemo(() => {
+    return isOwner && userQuota?.humanReadable.name.toLowerCase() === 'free';
+  }, [isOwner, userQuota?.humanReadable.name]);
+
+  const checkBlobSize = useCallback(
+    (blob: Blob) => {
+      const size = BigInt(blob.size);
```
EYHN: BigInt is not necessary?

packages/frontend/core/src/components/affine/quota-reached-modal/local-quota-modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { useAtom, useAtomValue } from 'jotai';
+import { useCallback, useEffect } from 'react';
+
+export const LocalQuotaModal = () => {
+  const t = useAFFiNEI18N();
+  const currentWorkspace = useAtomValue(waitForCurrentWorkspaceAtom);
+  const [open, setOpen] = useAtom(openQuotaModalAtom);
+
+  const checkBlobSize = useCallback((blob: Blob) => {
+    const size = BigInt(blob.size);
```
EYHN:
```suggestion
    const size = blob.size;
```

packages/frontend/core/src/components/affine/quota-reached-modal/local-quota-modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+    return size < bytes('100MB');
+  }, []);
+
+  const onConfirm = useCallback(() => {
+    setOpen(false);
+  }, [setOpen]);
+
+  useEffect(() => {
+    const disposable = currentWorkspace.engine.blob.onBlobSet.on(
```
EYHN: We listen to `currentWorkspace.engine.blob.isOverCapacity` here to open the model.

packages/common/workspace/src/engine/blob.ts
```diff
@@ -1,6 +1,13 @@
 import { DebugLogger } from '@affine/debug';
+import { Slot } from '@blocksuite/global/utils';
 import { difference } from 'lodash-es';
 
+export interface BlobSetEventArgs {
+  key: string;
+  value: Blob;
+  updateShouldProceed: (proceed: boolean) => void;
```
EYHN: Passing a side-effect function in the event doesn't looks very good to me.

Perhaps we can add a `singleBlobSizeLimit` field to limit the maximum size of a single file, and then set `blobEngine.singleBlobSizeLimit = 100mb` in react.

Then add an `onAbortLargeFile` event to trigger the size limit model.

```ts
singleBlobSizeLimit: number = 100mb
onAbortLargeBlob = new Slot<Blob>()
```

packages/common/workspace/src/engine/blob.ts
EYHN: This part of the error check is only for affine cloud, while BloeEngine is a general class

we can define a `BlobStorageOverCapacity` error,

Then check the status code === 413 in `affineCloudBlobStorage`, and then throw a `BlobStorageOverCapacity`.

```ts
if (err instanceof BlobStorageOverCapacity) {
  this.isOverCapacity = true
}
```

packages/common/workspace/src/engine/blob.ts
JimmFly: What do you think about this? @regischen

packages/frontend/core/src/components/affine/quota-reached-modal/local-quota-modal.tsx
JimmFly: In terms of design, we should limit the blob size when it is uploaded to the local. If it seems that it cannot be checked using isOverCapacity?

EYHN: Okay you can ignore this

packages/frontend/core/src/components/affine/quota-reached-modal/cloud-quota-modal.tsx
darkskygit: not needed when we used `SafeInt`

packages/common/workspace/src/engine/blob.ts
JimmFly: How does it look now? Does it match your idea?

packages/common/workspace/src/engine/blob.ts
```diff
@@ -122,6 +147,18 @@ export class BlobEngine {
       throw new Error('local peer is readonly');
     }
 
+    let shouldProceed = true;
```
EYHN: is useless, just use `value.size > this.singleBlobSizeLimit`

packages/common/workspace/src/engine/blob.ts
```diff
@@ -78,6 +98,11 @@ export class BlobEngine {
               await remote.set(key, data);
```
EYHN, line 98: <i class='graphite__hidden'>[Re: lines 76 to 76]<br/><br/></i>

We should add a condition here

```
if (this.status. isStorageOverCapacity) {
  return;
}
```<!--__GRAPHITE_HTML_TAG_START__--><span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5535?utm_source=unchanged-line-comment">Graphite</a>.</span>

<!--__GRAPHITE_HTML_TAG_END__-->

File: packages/common/workspace/src/engine/blob.ts (hunk 1 above)
Comment (EYHN) [reply packages/common/workspace/src/engine/blob.ts]: exactly

File: packages/common/workspace/src/engine/error.ts [hunk 18]
```diff
@@ -0,0 +1,14 @@
+export class BlobStorageOverCapacity extends Error {
+  constructor(
+    public override readonly message: string,
```
EYHN: These parameters are not needed here.

```ts
constructor(public originError?: any) {
  super("Blob storage over capacity.")
}
```

throw with:

```ts
throw new BlobStorageOverCapacity(graphqlError)
```

packages/frontend/workspace-impl/src/cloud/blob.ts
```diff
@@ -31,14 +32,26 @@ export const createAffineCloudBlobStorage = (
...
         variables: {
           workspaceId,
           blob: new File([value], key),
         },
-      });
-      return result.setBlob;
+      })
+        .then(res => res.setBlob)
+        .catch(err => {
+          if (err[0].extensions && err[0].extensions.code === 413) {
```
EYHN: You still need to add a condition here `if (err instanceof GraphQLError)`

If `err` is not a `GraphQLError` here [0] may also raise other errors.

packages/common/workspace/src/engine/blob.ts
regischen: I feel that it mainly depends on the product side.

regischen: Now, if a blob is not found, there will be a corresponding style. I feel that it might also be necessary to stop creating blocks after exceeding the limit.

JimmFly: Feels that a reminder should suffice, there's no need to prevent the creation of blocks.

joooye34 (approved):

PR #5561, opened 2024-01-10
feat(core): page info adapter for journal

pengx17:
Page info adapter + schema.
Adapted for journal features.

image


pengx17:
* `canary`
  * **PR #5561** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5561?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #5657** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5657?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5606** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5606?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #5638** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5638?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #5614** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5614?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5561?utm_source=stack-comment).

packages/frontend/core/src/modules/workspace/properties/schema.ts
```diff
@@ -0,0 +1,101 @@
...
+export const PageInfoTagsItemSchema = PageInfoItemSchema.extend({
+  id: z.literal(PageSystemPropertyId.Tags),
+  value: z.array(z.string()),
+});
+
+// ====== workspace properties schema ======
+export const WorkspaceFavoriteItemSchema = z.object({
+  id: z.string(),
+  order: z.number(),
+  type: z.enum(['page', 'collection']),
```
CatsJuice, line 63: Is the term defined here as `Favorite` actually supposed to cover both `Favorite` and `Collection`? Wouldn't this cause some ambiguity and lead to confusion?

```ts
adapter.isFavouriated('id', 'collection');
```

packages/frontend/core/src/modules/workspace/properties/adapter.ts
```diff
@@ -0,0 +1,161 @@
...
+    this.ensurePageProperties(id);
+    return !!this.getJournalPageDate(id);
+  }
+
+  getJournalPageDate(id: string) {
+    this.ensurePageProperties(id);
+    return this.pageProperties[id].system[PageSystemPropertyId.Journal].value;
+  }
+
+  setJournalPageDate(id: string, date: string) {
```
CatsJuice: ~~Maybe handle journal date here? Force validate and format date. I think we can do it later, but the date format should be determined as soon as possible, and should not be easily changed.~~

Maybe it's better to handle in `Journal Module` later.

packages/frontend/core/src/pages/workspace/detail-page/detail-page.tsx
```diff
@@ -179,6 +182,10 @@ const DetailPageImpl = memo(function DetailPageImpl({ page }: { page: Page }) {
         main={
           // Add a key to force rerender when page changed, to avoid error boundary persisting.
           <AffineErrorBoundary key={currentPageId}>
+            {adapter
```
CatsJuice: Is this for debug purpose and should be removed?

packages/frontend/core/package.json
```diff
@@ -48,6 +48,8 @@
     "@radix-ui/react-select": "^2.0.0",
     "@radix-ui/react-toolbar": "^1.0.4",
     "@react-hookz/web": "^24.0.0",
+    "@reactivedata/react": "^0.2.2",
```
CatsJuice: Did these dependencies used?

pengx17: No they are not. Will be removed.

packages/frontend/core/src/modules/workspace/properties/schema.ts
pengx17: I haven't thought through, but can you please elaborate more on the ambiguity?
Do you mean that there is a very low chance that the same id os used for two different object?

CatsJuice: In my opinion, A 'page' can exist in both 'Favorite' and 'Collection', so there are two concepts. However, here 'Favorite' is used to represent both, which makes the naming a bit strange.

packages/frontend/core/src/modules/workspace/properties/adapter.ts
pengx17: I haven't decided how to make more fine-graind abstraction for this yet. It could follow some approach of that in blocksuite, but maybe it is a bit early at this stage

packages/frontend/core/src/modules/workspace/properties/schema.ts
pengx17: I think it is not the correct case.

In the new design we will have data like
```ts
const foo: WorkspaceAffineProperties['favouriates'] = {
  pageId_0: {
    id: 'pageId_0',
    order: 0,
    type: 'page',
  },
  collectionId_0: {
    id: 'collectionId_0',
    order: 1,
    type: 'collection',
  },
};
```
page `pageId_0` could exist in `collectionId_0`, but it is not relevant to the this cases. The page -> collection mapping is handled elsewhere

packages/frontend/core/src/hooks/use-journal.ts
```diff
@@ -36,9 +39,17 @@ export const useJournalHelper = (workspace: BlockSuiteWorkspace) => {
...
+      adapter.setJournalPageDateString(page.id, title);
       return page;
     },
-    [bsWorkspaceHelper]
+    [adapter, bsWorkspaceHelper]
+  );
+
+  const isPageJournal = useCallback(
+    (pageId: string) => {
+      return isJournalString(adapter.getJournalPageDateString(pageId));
```
CatsJuice: Maybe we don't need to validate format which is already stored on `pageInfo`?

packages/frontend/core/src/hooks/use-affine-adapter.ts
CatsJuice, line 1: Seems duplicated with `packages/frontend/core/src/hooks/use-affine-properties.ts`, is it resolved in later PRs?

pengx17: Oh, this should be removed. Thanks!

packages/frontend/core/src/hooks/use-journal.ts
pengx17: it does no harm with an additional check while it makes the implementation a bit more robust

CatsJuice: 😆 Yes, I'm just worry about that in case we change the format later, then there could be multiple formats. Anyway, that not matters, let's keep it this way!

packages/frontend/core/src/modules/workspace/properties/adapter.ts
```diff
@@ -0,0 +1,156 @@
...
+              id: PageSystemPropertyId.Tags,
+              name: 'Tags',
+              source: 'system',
+              type: PagePropertyType.Tags,
+              options: this.workspace.meta.properties.tags?.options ?? [], // better use a one time migration
+            },
+          },
+        },
+      },
+      favouriates: {},
```
joooye34: favorite or favourite ?

pengx17: Our App is actually en-GB so favourite > favorite. But it is a matter of the display texts. Not sure if we need to follow the same rules for the internal impl?

packages/frontend/core/src/hooks/use-journal.ts
joooye34: Is the isJournalString function still needed?

joooye34 (approved):

PR #5562, opened 2024-01-10
feat(infra): livedata

EYHN:
LiveData is a reactive data type.
 
## basic usage
 
@example
```ts
const livedata = new LiveData(0); // create livedata with initial value
 
livedata.next(1); // update value
 
console.log(livedata.value); // get current value
 
livedata.subscribe(v => { // subscribe to value changes
 console.log(v); // 1
});
```
 
## observable
 
LiveData is a rxjs observable, you can use rxjs operators.
 
@example
```ts
new LiveData(0).pipe(
  map(v => v + 1),
  filter(v => v > 1),
  ...
)
```
 
NOTICE: different from normal observable, LiveData will always emit the latest value when you subscribe to it.
 
## from observable
 
LiveData can be created from observable or from other livedata.
 
@example
```ts
const A = LiveData.from(
  of(1, 2, 3, 4), // from observable
  0 // initial value
);
 
const B = LiveData.from(
  A.pipe(map(v => 'from a ' + v)), // from other livedata
  '' // initial value
);
```
 
NOTICE: LiveData.from will not complete when the observable completes, you can use `spreadComplete` option to change
this behavior.
 
## Why is it called LiveData
 
This API is very similar to LiveData in Android, as both are based on Observable, so I named it LiveData.


EYHN:
* `canary`
  * **PR #5562** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5562?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #5595** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5595?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5497** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5497?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #5563** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5563?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #5564** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5564?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #5693** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5693?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
            * **PR #5617** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5617?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
              * **PR #5618** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5618?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
                * **PR #5565** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5565?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5562?utm_source=stack-comment).

packages/common/infra/src/livedata/index.ts
```diff
@@ -0,0 +1,105 @@
...
+ * this behavior.
+ *
+ * ## Why is it called LiveData
+ *
+ * This API is very similar to LiveData in Android, as both are based on Observable, so I named it LiveData.
+ *
+ * @see {@link https://rxjs.dev/api/index/class/BehaviorSubject}
+ * @see {@link https://developer.android.com/topic/libraries/architecture/livedata}
+ */
+export class LiveData<T = unknown> extends BehaviorSubject<T> {
```
pengx17: ~~I have no knowledge on Android's LiveData at all and after reading some brief intro it seems to me that LiveData in JetPack enables lifecycle bindings to views. It seems to me that if we borrow the LiveData term, it might be used together in a hook.~~
Oh I see. There is a `useLiveData` hook 😂

packages/common/infra/src/livedata/react.ts
```diff
@@ -0,0 +1,41 @@
+import { use } from 'foxact/use';
+import { useEffect, useState } from 'react';
+
+import type { LiveData } from '.';
```
Brooooooklyn: Invalid import path, use `./index` here

packages/common/infra/src/livedata/react.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * subscribe LiveData and return the value. If the value is nullish, will suspends until the value is not nullish.
+ */
+export function useEnsureLiveData<T>(liveData: LiveData<T>): NonNullable<T> {
+  const data = useLiveData(liveData);
+
+  if (data === null || data === undefined) {
+    return use(
+      new Promise(resolve => {
+        liveData.subscribe(value => {
```
Brooooooklyn: error handling

pengx17: Please unsub when resolved

pengx17:
Who is this @example?

packages/common/infra/src/livedata/index.ts
EYHN: Actually, the LiveData currently implemented is not bound to any lifecycle, because the few LiveData used so far are pure, so there is no need for destruction.

If the Observable inside the LiveData has side effects, it should be unsubscribe when there are no subscribers. I plan to do this later.

pengx17: Not only about observable unsubscription, the parts on data-binding for view updates looks more relevant for LiveData abstraction. Regarding side-effects, not sure if we want to directly add side effects in the piped operators - maybe we should have another abstraction over it?

pengx17 (approved):

packages/common/infra/src/livedata/index.ts
```diff
@@ -0,0 +1,293 @@
...
+  Observable,
+  type Observer,
+  of,
+  type OperatorFunction,
+  scan,
+  skip,
+  type Subscription,
+  switchMap,
+} from 'rxjs';
+import { BehaviorSubject, Subject } from 'rxjs';
```
forehalo, line 17: clean the imports, better to move type imports to a separated one?

EYHN: We need to add a lint rule to check, there are too many such codes

PR #5582, opened 2024-01-12
fix: improve self-host convenience

forehalo:


forehalo:
* `canary`
  * **PR #5582** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5582?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5582?utm_source=stack-comment).

packages/backend/server/.env.example
```diff
@@ -1,8 +1,4 @@
-DATABASE_URL="postgresql://affine@localhost:5432/affine"
-NEXTAUTH_URL="http://localhost:8080"
-OAUTH_EMAIL_SENDER="noreply@toeverything.info"
-OAUTH_EMAIL_LOGIN=""
-OAUTH_EMAIL_PASSWORD=""
-ENABLE_LOCAL_EMAIL="true"
-STRIPE_API_KEY=
-STRIPE_WEBHOOK_KEY=
+# AFFINE_SERVER_PORT=3010
```
darkskygit, line 1: .env.example is a showcase for developer, do we make a copy to use to self-hosted version like `.env.example.self-hosted` ?

forehalo: there won't be too much difference for self hosted server, I don't think it's necessary to make a separated one.

im thinking about a even better way to provide available env vars list, to solve the problem that the dotenv file is always outdated.

darkskygit (approved):

PR #5590, opened 2024-01-13
style: import from './index' instead of '.'

lewisl9029:
This PR replaces all the `import x from '.'`s in the codebase with `import x from './index'`.

`import x from '.'` is a bit of a dead-end, since `'.'` is not a valid import specifier in browsers. As far as I'm aware, there's no way to even use import maps to map them into the right "index.js" files, so it will likely never become supported in browsers.

Reflame's React Fast Refresh implementation in development mode makes use of import maps to avoid having to invalidate more modules than necessary on updates, which is how I discovered these.


Brooooooklyn (approved):
Thanks!

PR #5613, opened 2024-01-17
feat: use `SafeInt` replace `Float`

darkskygit:


darkskygit:
* `canary`
  * **PR #5613** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5613?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5613?utm_source=stack-comment).

packages/frontend/graphql/codegen.yml
```diff
@@ -16,7 +16,7 @@ config:
     Decimal: number
     UUID: string
     ID: string
-    JSON: any
+    JSON: string
```
EYHN, line 19: I think JSON: any is right?

and we should add SafeInt to here

darkskygit: Currently there is no place to use the JSON type in server, and graphql-scalers contains a JSON type: https://the-guild.dev/graphql/scalars/docs/scalars/json
And the original type of JSON in graphql-scalers is defined as String: https://the-guild.dev/graphql/scalars/docs/quick-start#graphql-code-generator-integration
We can use JSON type in graphql-scalers directly in future

Brooooooklyn (approved):

PR #5625, opened 2024-01-18
fix(core): reserve space for the editor scrollbar

JimmFly:
Because the space for the scroll bar on the right is reserved, in order to make the editor symmetrical, padding corresponding to the width of the scroll bar is added to the left.

### DIFF
--- a/packages/frontend/core/src/components/page-detail-editor.css.ts
+++ b/packages/frontend/core/src/components/page-detail-editor.css.ts
@@ -18,6 +18,8 @@ export const editor = style({
 
 globalStyle(`${editor} .affine-doc-viewport`, {
   paddingBottom: '150px',
+  paddingLeft: '20px',
+  scrollbarGutter: 'stable',
 });
 
 globalStyle('.is-public-page page-meta-tags', {


JimmFly:
* `canary`
  * **PR #5625** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5625?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5625?utm_source=stack-comment).

pengx17 (approved):

packages/frontend/core/src/components/page-detail-editor.css.ts
```diff
@@ -18,6 +18,8 @@ export const editor = style({
 
 globalStyle(`${editor} .affine-doc-viewport`, {
   paddingBottom: '150px',
+  paddingLeft: '20px',
```
pengx17, line 21: I am not sure - is `stable both-edges` for this case?

JimmFly: If this parameter is used, drag handle will overlap with toggle, so it is not considered to be used for the time being.

PR #5630, opened 2024-01-18
refactor(server): plugin modules

forehalo:
- [x] separates modules into `fundamental`, `core`, `plugins`
- [x] optional modules with `@OptionalModule` decorator to install modules with requirements met(`requires`, `if`)
- [x] `module.contributesTo` defines optional features that will be enabled if module registered
- [x] `AFFiNE.plugins.use('payment', {})` to enable a optional/plugin module
- [x] `PaymentModule` is the first plugin module
- [x] GraphQLSchema will not be generated for non-included modules
- [x] Frontend can use `ServerConfigType` query to detect which features are enabled
- [x] override existing provider globally


forehalo:
* `canary`
  * **PR #5630** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5630?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5630?utm_source=stack-comment).

packages/backend/server/src/fundamentals/config/def.ts
```diff
@@ -52,15 +51,28 @@ export interface AFFiNEConfig {
   /**
    * Server Identity
    */
-  readonly serverId: string;
+  serverId: string;
+
+  /**
+   * Name may show on the UI
+   */
+  serverName: string;
```
darkskygit, line 59: seems will not change in runtime? i think it can be readonly

darkskygit (approved):

packages/backend/server/src/fundamentals/config/def.ts
forehalo: user may want to update it in selfhosted mode, like 'Ly's CLoud'

PR #5718, opened 2024-01-26
feat(core): add starAFFiNE and issueFeedback modal

JimmFly:
close TOV-482

https://github.com/toeverything/AFFiNE/assets/102217452/da1f74bc-4b8d-4d7f-987d-f53da98d92fe


JimmFly:
* `canary`
  * **PR #5718** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5718?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #5733** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5733?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5718?utm_source=stack-comment).

packages/frontend/component/src/components/overlay-modal/overlay-modal.css.tsx
```diff
@@ -0,0 +1,27 @@
+import { style } from '@vanilla-extract/css';
+
+export const title = style({
+  padding: '20px 24px 8px 24px',
+  fontSize: 'var(--affine-font-h-6)',
```
CatsJuice: Now it's better to use `cssVar` from `@toeverything/theme` for theme variables.

packages/frontend/component/src/components/overlay-modal/overlay-modal.tsx
CatsJuice, line 1: This component should be moved  to `ui`, since it is generic. And onboarding modal should be replaced with this component later.

packages/frontend/component/src/components/overlay-modal/overlay-modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+  style: {
+    padding: 0,
+    overflow: 'hidden',
+    boxShadow: 'var(--affine-menu-shadow)',
+  },
+};
+const defaultOverlayOptions: ModalProps['overlayOptions'] = {
+  style: {
+    background: 'var(--affine-white-80)',
+    backdropFilter: 'blur(2px)',
```
CatsJuice: Seems unnecessary, because of blurring `#app` directly below in `useEffect`.

packages/frontend/component/src/components/overlay-modal/overlay-modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+  },
+};
+
+export interface OverlayModalProps extends ModalProps {
```
CatsJuice: I think it's better to provide a `to` prop, to make the confirm button behave as a `<Link />`. Also we don't need to handle the open url behavior outside component.

packages/frontend/component/src/components/overlay-modal/overlay-modal.tsx
JimmFly: In this ui component, useEffect seems to be unnecessary, because the background blur processed by useEffect does not change the background of the modal. I would use it alone for onboarding modal.

packages/frontend/component/src/components/overlay-modal/overlay-modal.tsx
JimmFly: I will add `to` prop, but I will keep the `onConfirm` in order to pass in different click events.

packages/frontend/component/src/components/overlay-modal/overlay-modal.tsx
JimmFly: image

packages/frontend/component/src/components/overlay-modal/overlay-modal.tsx
JimmFly: For example, click to open enable cloud modal

packages/frontend/component/src/ui/overlay-modal/overlay-modal.tsx
```diff
@@ -0,0 +1,80 @@
...
+      {topImage}
+      <div className={styles.title}>{title}</div>
+      <div className={styles.content}>{description}</div>
+      <div className={styles.footer}>
+        {!withoutCancelButton ? (
+          <DialogTrigger asChild>
+            <Button {...cancelButtonOptions}>{cancelText}</Button>
+          </DialogTrigger>
+        ) : null}
+        <Button onClick={handleConfirm} {...confirmButtonOptions}></Button>
```
CatsJuice: I mean that it's better to use a `<Link>` component when `to` is provided, so the button can behave as native `<a>` tag, and we don't need to open url with JS.

packages/frontend/component/src/ui/overlay-modal/overlay-modal.tsx
CatsJuice, line 1: Add the corresponding story

packages/frontend/core/src/components/affine/setting-modal/style.css.ts
```diff
@@ -50,4 +50,12 @@ export const footer = style({
   justifyContent: 'center',
   alignItems: 'center',
   paddingBottom: '20px',
+  gap: '4px',
+  fontSize: 'var(--affine-font-xs)',
```
CatsJuice: use `cssVar` instead

CatsJuice (approved):

PR #5728, opened 2024-01-27
fix(core): replace most --affine with cssVar

pengx17:
using a [babel plugin](https://gist.github.com/pengx17/49e24ae8a5a609bdaff122ee8c679d1c) to transform all var(--affine-xxx) to cssVar 

Some issues:
- tried ast-grep but it seems to be not easy to add imports conditionally
- current work does not work well with ts with types because babel will strip them out

### DIFF
--- a/packages/frontend/component/src/components/affine-banner/index.css.ts
+++ b/packages/frontend/component/src/components/affine-banner/index.css.ts
@@ -1,10 +1,10 @@
+import { cssVar } from '@toeverything/theme';
 import { style } from '@vanilla-extract/css';
-
 export const browserWarningStyle = style({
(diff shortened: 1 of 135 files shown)


pengx17:
* `canary`
  * **PR #5728** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5728?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5728?utm_source=stack-comment).

CatsJuice (approved):

PR #5729, opened 2024-01-28
feat(core): page info ui

pengx17:
this PR includes the main table view in the page detail page

### DIFF
--- a/packages/frontend/core/src/components/affine/page-properties/icons-mapping.tsx
+++ b/packages/frontend/core/src/components/affine/page-properties/icons-mapping.tsx
@@ -1,0 +1,57 @@
+import type { PagePropertyType } from '@affine/core/modules/workspace/properties/schema';
+import * as icons from '@blocksuite/icons';
+import type { SVGProps } from 'react';
+
+type IconType = (props: SVGProps<SVGSVGElement>) => JSX.Element;
+
+// todo: this breaks tree-shaking, and we should fix it (using dynamic imports?)
+const IconsMapping = {
+  text: icons.TextIcon,
+  tag: icons.TagIcon,
+  dateTime: icons.DateTimeIcon,
+  progress: icons.ProgressIcon,
+  checkbox: icons.CheckBoxCheckLinearIcon,
+  number: icons.NumberIcon,
+  // todo: add more icons
+} satisfies Record<string, IconType>;
+
+export type PagePropertyIcon = keyof typeof IconsMapping;
+
+export const getDefaultIconName = (
+  type: PagePropertyType
+): PagePropertyIcon => {
+  switch (type) {
+    case 'text':
+      return 'text';
+    case 'tags':
+      return 'tag';
+    case 'date':
+      return 'dateTime';
+    case 'progress':
+      return 'progress';
+    case 'checkbox':
+      return 'checkbox';
+    case 'number':
+      return 'number';
+    default:
+      return 'text';
+  }
+};
+
+// fixme: this function may break if icons are imported twice
+export const IconToIconName = (icon: IconType) => {
+  const iconKey = Object.entries(IconsMapping).find(([_, candidate]) => {
+    return candidate === icon;
+  })?.[0];
+  return iconKey;
+};
+
+export const nameToIcon = (
+  iconName: string,
+  type: PagePropertyType
+): IconType => {
+  return (
+    IconsMapping[iconName as keyof typeof IconsMapping] ??
+    getDefaultIconName(type)
+  );
+};
--- a/packages/frontend/core/src/components/affine/page-properties/table.tsx
+++ b/packages/frontend/core/src/components/affine/page-properties/table.tsx
@@ -1,0 +1,892 @@
+import {
+  Button,
+  IconButton,
+  Menu,
+  MenuIcon,
+  MenuItem,
+  Scrollable,
+  Tooltip,
+} from '@affine/component';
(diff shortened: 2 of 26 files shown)


pengx17:
* **#5809** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5809?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5769** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5769?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5748** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5748?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5739** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5739?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5729** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5729?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#5804** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5804?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5803** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5803?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5729?utm_source=stack-comment).

packages/frontend/core/src/components/affine/page-properties/icons-mapping.tsx
```diff
@@ -0,0 +1,57 @@
...
+  number: icons.NumberIcon,
+  // todo: add more icons
+} satisfies Record<string, IconType>;
+
+export type PagePropertyIcon = keyof typeof IconsMapping;
+
+export const getDefaultIconName = (
+  type: PagePropertyType
+): PagePropertyIcon => {
+  switch (type) {
```
CatsJuice, line 23: I prefer to use a mapping object, which is more intuitive and provides better type hints.

```ts
const propertyTypeIconMap: Record<PagePropertyType, PagePropertyIcon> = {}
```

Then once one of the types changes, we must update the mapping object.

packages/frontend/core/src/components/affine/page-properties/table.tsx
```diff
@@ -0,0 +1,892 @@
...
+    </SortablePropertyRow>
+  );
+};
+
+interface PagePropertiesTableBodyProps {
+  className?: string;
+  style?: React.CSSProperties;
+}
+
+const modifiers = [restrictToParentElement, restrictToVerticalAxis];
```
CatsJuice, line 634: 🤔 Lift this up? Since it's only used at line 155.

packages/frontend/core/src/components/affine/page-properties/table.tsx
```diff
@@ -0,0 +1,892 @@
...
+    (event: DragEndEvent) => {
+      if (readonly) {
+        return;
+      }
+      const { active, over } = event;
+      const fromIndex = properties.findIndex(p => p.id === active.id);
+      const toIndex = properties.findIndex(p => p.id === over?.id);
+
+      if (fromIndex !== -1 && toIndex !== -1 && fromIndex !== toIndex) {
+        const newOrdered = arrayMove(properties, fromIndex, toIndex);
```
CatsJuice, line 102: It seems that with this implementation, the order of each property needs to be updated each time when sorting, even if the order of a certain item has not changed.

tests/affine-local/e2e/local-first-collections-items.spec.ts
```diff
@@ -137,7 +137,7 @@ test('edit collection and change filter date', async ({ page }) => {
   expect(await first.textContent()).toBe('123');
 });
 
-test('create temporary filter by click tag', async ({ page }) => {
+test.skip('create temporary filter by click tag', async ({ page }) => {
```
Brooooooklyn, line 140: skip as expect?

pengx17: it is fixed in a later stack

packages/frontend/core/src/components/affine/page-properties/table.tsx
pengx17: Yeah, it was moved in a later stack https://github.com/toeverything/AFFiNE/blob/fbe25bc837419a7ef954a7262d0ea30355609752/packages/frontend/core/src/components/affine/page-properties/table.tsx#L97

packages/frontend/core/src/components/affine/page-properties/icons-mapping.tsx
pengx17: This gives an opportunity for possible fallback. Also, it is mostly type-wise  equalvant.

packages/frontend/core/src/components/affine/page-properties/table.tsx
pengx17: It should be fine to change the `order` property only. If we store values in an array and move items in the array, it may cause some unexpected issues since the items at the same index may change completely.

Brooooooklyn (approved):

packages/frontend/core/src/components/affine/page-properties/table.tsx
pengx17: fixed in #5809

PR #5750, opened 2024-01-31
feat: move templates into AFFiNE 

doouding:
Related to https://github.com/toeverything/blocksuite/pull/6156

### Change
Move the edgeless templates to AFFiNE. All templates are stored as zip files. Run `node build-edgeless.mjs` in `@affine/templates` to generate JSON-format templates and importing script. The template will be generated automatically during building and dev (`yarn dev`).


packages/common/env/package.json
```diff
@@ -3,8 +3,8 @@
   "private": true,
   "type": "module",
   "devDependencies": {
-    "@blocksuite/global": "0.12.0-nightly-202401290223-b6302df",
-    "@blocksuite/store": "0.12.0-nightly-202401290223-b6302df",
+    "@blocksuite/global": "0.12.0-nightly-202401312339-b178f26",
+    "@blocksuite/store": "0.12.0-nightly-202401312339-b178f26",
```
pengx17: please move blocksuite bump as a parent stack of this pr

packages/frontend/templates/build-edgeless.mjs
```diff
@@ -0,0 +1,289 @@
...
+
+import JSZip from 'jszip';
+
+const __dirname = join(fileURLToPath(import.meta.url), '..');
+const ZIP_PATH = join(__dirname, './edgeless-snapshot');
+const ASSETS_PREFIX = `/static/templates`;
+const ASSETS_PATH = join(__dirname, '../core/public/', ASSETS_PREFIX);
+const TEMPLATE_PATH = join(__dirname, './edgeless');
+
+const zipFiles = promisify(fs.readdir)(ZIP_PATH).then(files => {
```
pengx17: you can use readdir from `node:fs/promises` or just use`readdirSync`

packages/frontend/templates/build-edgeless.mjs
```diff
@@ -0,0 +1,289 @@
...
+ * @property {object} props
+ * @property {string} props.sourceId
+ */
+
+/**
+ * @param {Block} block
+ */
+const convertSourceId = block => {
+  if (block.props?.sourceId) {
+    block.props.sourceId = `${ASSETS_PREFIX}/${block.props.sourceId}.png`;
```
pengx17: I guess the assets might always not be png files

packages/frontend/core/src/components/page-detail-editor.tsx
```diff
@@ -39,6 +44,14 @@ function useRouterHash() {
   return useLocation().hash.substring(1);
 }
 
+let extended = false;
+function useEdgelessTemplate() {
+  if (!extended) {
+    EdgelessTemplatePanel.templates.extend(builtInTemplates as TemplateManager);
+    extended = true;
+  }
+}
```
pengx17: No need to use a hook here since it is not dependent on any react states

packages/frontend/templates/build-edgeless.mjs
```diff
@@ -0,0 +1,289 @@
...
+    .map(template => {
+      return `import ${toVariableName(
+        template
+      )} from './edgeless/${template}.json';`;
+    })
+    .join('\n');
+
+  const code = `${importStatements}
+
+const templates = {
```
pengx17, line 250: can you reorg the files stored in edgeless-snapshots in a pattern and then enumerate the template meta, instead of maintain these object explicity? I think in the long term what we want is to replace snapshot files only without touching this script

packages/frontend/core/src/components/page-detail-editor.tsx
doouding: > No need to use a hook here since it is not dependent on any react states

Not so sure about the code style and structure. Where is the correct place to put this extension call?

packages/frontend/templates/build-edgeless.mjs
doouding: Good idea.

packages/frontend/core/src/components/page-detail-editor.tsx
pengx17: I think you can put it into `packages/frontend/core/src/bootstrap/register-blocksuite-components.ts`

packages/frontend/templates/edgeless-snapshot/4P Marketing Matrix.zip
Brooooooklyn, line 1: Is this zip file necessary?

doouding: I had a discussion with @joooye34 and @pengx17 and concluded that zip files are more convenient and intuitive.

Brooooooklyn (approved):

PR #5752, opened 2024-01-31
refactor(core): use custom scrollbar for editor and adjust shared page style

JimmFly:
Close TOV-481
- Use a custom scrollbar component for editor
- Modified the header of the share page and added a new footer


JimmFly:
* **#5752** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5752?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5752?utm_source=stack-comment).

packages/frontend/core/src/components/cloud/share-header-right-item/authenticated-item.tsx
```diff
@@ -1,24 +1,39 @@
...
   const members = useMembers(workspaceId, 0);
   const isMember = members.some(m => m.id === user.id);
   const t = useAFFiNEI18N();
   const { jumpToPage } = useNavigateHelper();
 
+  const handleEdit = useCallback(() => {
+    jumpToPage(workspaceId, pageId);
+  }, [workspaceId, pageId, jumpToPage]);
+
+  useEffect(() => {
```
CatsJuice, line 26: I think we can get this value in parent component so we don't need this effect. Is there any reason that I've not considered why it's necessary?

JimmFly: Because `useMember` sends a request to the database and must be `authenticated` to be used, `useMember` cannot extract to its parent component.

CatsJuice (approved):

Brooooooklyn (approved):

PR #5798, opened 2024-02-04
fix(core): prevent data loss (hot-fix)

EYHN:


Brooooooklyn (approved):

packages/frontend/core/src/pages/workspace/index.tsx
```diff
@@ -67,12 +73,34 @@ export const Component = (): ReactElement => {
     localStorage.setItem('last_workspace_id', workspace.id);
   }, [setCurrentWorkspace, meta, workspaceManager, workspace]);
 
+  const [workspaceIsLoading, setWorkspaceIsLoading] = useState(true);
+
+  // hotfix: avoid doing operation, before workspace is loaded
```
pengx17, line 78: maybe combine this with doc.whenSynced?

PR #5831, opened 2024-02-18
refactor(core): refactor atom to use di

EYHN:
To support multiple instances, this PR removes some atoms and implements them using the new DI system.

removed atom

- `pageSettingsAtom`
- `currentPageIdAtom`
- `currentModeAtom`


EYHN:
* **#5837** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5837?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5832** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5832?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5831** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5831?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5831?utm_source=stack-comment).

packages/common/infra/src/livedata/react.ts
```diff
@@ -3,13 +3,29 @@ import { useSyncExternalStore } from 'react';
...
+
+function noopGetSnapshot() {
+  return null;
+}
+
 /**
  * subscribe LiveData and return the value.
  */
-export function useLiveData<T>(liveData: LiveData<T>): T {
+export function useLiveData<Input extends LiveData<any> | null | undefined>(
```
forehalo, line 17: in which scenario the LiveData passed to this hook is nullable?

packages/frontend/core/src/modules/services.ts
```diff
@@ -23,5 +29,12 @@ export function configureBusinessServices(services: ServiceCollection) {
 }
 
 export function configureWebInfraServices(services: ServiceCollection) {
-  services.addImpl(GlobalCache, LocalStorageGlobalCache);
+  services
+    .addImpl(GlobalCache, LocalStorageGlobalCache)
+    .addImpl(GlobalState, LocalStorageGlobalState)
+    .scope(WorkspaceScope)
```
forehalo: is it better to be put in `configureWorkspaceServices`, because I saw an overriding of `WorkspaceLocalState` in testing environment, but I didn't see it in normal environment. It makes me confused at first until I search with it globally

packages/frontend/core/src/bootstrap/first-app-data.ts
```diff
@@ -1,43 +1,34 @@
...
-          },
-        });
-      } else {
+  if (runtimeConfig.enablePreloading) {
+    const workspaceMetadata = await buildShowcaseWorkspace(
+      workspaceManager,
+      WorkspaceFlavour.LOCAL,
+      DEFAULT_WORKSPACE_NAME
+    );
+    console.info('create first workspace', workspaceMetadata);
```
forehalo: logger?

packages/frontend/core/src/components/affine/affine-error-boundary/error-basic/info-logger.tsx
```diff
@@ -18,17 +16,17 @@ export const DumpInfo = (_props: DumpInfoProps) => {
   const currentWorkspace = useLiveData(
     useService(CurrentWorkspaceService).currentWorkspace
   );
-  const currentPageId = useAtomValue(currentPageIdAtom);
+  const currentPage = useServiceOptional(Page);
```
forehalo, line 19: we should call it `useOptionalService` or `useService(S, { optional: true })`, `useOptionalService` could be an alias of `useService()`

[confused](https://app.graphite.dev/macros/toeverything)

packages/frontend/core/src/hooks/affine/use-register-blocksuite-editor-commands.tsx
```diff
@@ -18,10 +18,10 @@ import { useBlockSuiteMetaHelper } from './use-block-suite-meta-helper';
 import { useExportPage } from './use-export-page';
 import { useTrashModalHelper } from './use-trash-modal-helper';
 
-export function useRegisterBlocksuiteEditorCommands(
-  pageId: string,
-  mode: 'page' | 'edgeless'
-) {
+export function useRegisterBlocksuiteEditorCommands() {
+  const page = useService(Page);
```
forehalo, line 22: is `useService` reactive? if not, the hooks below use `pageId` as dependency would introduce unexpected behavior

packages/common/infra/src/livedata/react.ts
EYHN: before:

```ts
const page = useServiceOptional(Page);
const mode = useLiveData(page?.mode ?? new LiveData(null));
```

after:

```ts
const page = useServiceOptional(Page);
const mode = useLiveData(page?.mode);
```

packages/frontend/core/src/hooks/affine/use-register-blocksuite-editor-commands.tsx
EYHN: useService is reactive

packages/common/infra/src/livedata/react.ts
forehalo: but if the page service is reactive here, why do we need another `useLiveDate` on fetching `mode`?

packages/frontend/core/src/components/affine/affine-error-boundary/error-basic/info-logger.tsx
EYHN: `useService(S, { optional: true })` is hard in typescript...

EYHN: emm I feel there is no difference, may be no need to change it.

packages/common/infra/src/livedata/react.ts
EYHN: page service is reactive means when we change page, page service will change. (actually is service provider changed)

page is internally variable, page internal is not reactive

packages/common/infra/src/page/record.ts
```diff
@@ -0,0 +1,65 @@
...
+  setMode(mode: PageMode) {
+    this.localState.set(`page:${this.id}:mode`, mode);
+  }
+
+  toggleMode() {
+    this.setMode(this.mode.value === 'edgeless' ? 'page' : 'edgeless');
+    return this.mode.value;
+  }
+
+  title = this.meta.map(meta => meta.title);
```
pengx17, line 64: Change this to getter? Title here is a static value after instaniate

packages/common/infra/src/page/record.ts
```diff
@@ -0,0 +1,65 @@
...
+      return () => {
+        dispose();
+      };
+    }),
+    {
+      id: this.id,
+      title: '',
+      tags: [],
+      createDate: 0,
+    }
```
pengx17, line 43: It seems to me the default value can be omitted since there is a `emit` call in the observable body

packages/common/infra/src/page/record.ts
EYHN: I believe title is variable here

packages/common/infra/src/page/record.ts
EYHN: Although there is an emit call, if pageMeta does not exist, this default value will still be used.

packages/frontend/core/src/pages/workspace/detail-page/detail-page.tsx
```diff
@@ -301,6 +299,14 @@ export const DetailPage = ({ pageId }: { pageId: string }): ReactElement => {
     currentWorkspace.setPriorityRule(id => id.endsWith(pageId));
   }, [pageId, currentWorkspace]);
 
+  const jumpOnce = useLiveData(pageRecord?.meta.map(meta => meta.jumpOnce));
```
pengx17, line 305: I think the `jumpOnce` flag can be moved to PageRecord now since it is a side effect transient flag and not meant to be synced or persisted

EYHN: I feel jumpOnce is a bad design and I want to remove it in the future

pengx17: Agreed.

pengx17 (approved):

PR #5832, opened 2024-02-18
refactor(core): remove collection atom

EYHN:

PR #5837, opened 2024-02-19
feat(core): workbench system

EYHN:
# Workbench

CleanShot 2024-02-20 at 10 15 46@2x

# router

```
RootRouter
└── WorkspaceLayout
    └── WorkBenchRoot
        ├── View1
        │   └── MemoryRouter
        ├── View2
        │   └── MemoryRouter
        └── View3
            └── MemoryRouter
```


EYHN:
* **#5837** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5837?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#5832** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5832?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5831** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5831?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5837?utm_source=stack-comment).

packages/frontend/core/src/modules/workbench/workbench.ts
```diff
@@ -0,0 +1,91 @@
...
+  open(to: To, { at = 'active' }: { at?: WorkbenchPosition } = {}) {
+    let view = this.viewAt(at);
+    if (!view) {
+      const newIndex = this.createView(at);
+      view = this.viewAt(newIndex);
+      if (!view) {
+        throw new Unreachable();
+      }
+    }
+    view.history.push(to);
```
pengx17: replace will not be supported?

packages/frontend/core/src/modules/workbench/browser-adapter.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  const view = useLiveData(workbench.activeView);
+
+  useEffect(() => {
+    return view.history.listen(update => {
+      if (update.action === 'POP' || update.action === 'REPLACE') {
+        throw new Error(
+          'POP and REPLACE view history is not allowed on browser'
+        );
+      }
```
pengx17: Mind explain why replace is not supported? It is useful for cases like redirect

packages/frontend/core/src/modules/workbench/desktop-adapter.ts
```diff
@@ -0,0 +1,38 @@
...
+    const newLocation = browserLocationToViewLocation(
+      browserLocation,
+      basename
+    );
+    if (newLocation === null) {
+      return;
+    }
+
+    workbench.open(newLocation);
+  }, [basename, browserLocation, workbench]);
```
pengx17, line 35: I think `browserLocation` from `useLocation` will be a new reference on route change. May not use it in effect deps

pengx17 (commented):
I feel like we may need to add some new baseline e2e tests for this stack of PR

packages/frontend/core/src/modules/workbench/workbench.ts
pengx17: In some cases, a route view may redirect to another view on landing, which creates a unwanted item in the history stack.
e.g., `goto("root.a") =>  "root.a" redirected to "root.a.foo"`
By using `replace`, "root.a" will not be on the stack. If not using `replace` and using route backwards at `root.a.foo`, the app will bounce back from `root.a` because it triggers redirect again

pengx17 (approved):
LGTM!

packages/frontend/core/src/modules/workbench/desktop-adapter.ts
EYHN: In desktop, we not really care about the browser history, we only listen it, and never modify it.

Only when the legacy navigation api is used (in most cases now), the browser route will change.

So in the case of a browser route change, we simply push the new route into the active view.

pengx17: I am a bit concerned that putting route navigate in use effect may result in unwanted history items in the stack. Perhaps skipping route navigate in `workbench.open` if new route is exactly the same with the previous one?

packages/frontend/core/src/modules/workbench/browser-adapter.ts
EYHN: replace is now supported !

packages/frontend/core/src/modules/workbench/workbench.ts
EYHN: replace is now supported !

CatsJuice (approved):

PR #5885, opened 2024-02-23
fix(core): fix app boot speed

EYHN:


EYHN:
* **#5885** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5885?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5885?utm_source=stack-comment).

Brooooooklyn (approved):

packages/frontend/core/src/providers/modal-provider.tsx
```diff
@@ -216,7 +216,9 @@ export const SignOutConfirmModal = () => {
 
   const onConfirm = useAsyncCallback(async () => {
     setOpen(false);
-    await signOutCloud();
+    await signOutCloud({
+      redirect: false,
+    });
```
pengx17, line 221: is this relevant?

pengx17 (approved):

packages/frontend/core/src/providers/modal-provider.tsx
EYHN: There is jump logic below this line, If redirected here, the code below will not be run.

EYHN: The previously added `workspaceListStatus.loading` judgment is to solve the page not found error of jumping to cloud workspace after sign out.

PR #5951, opened 2024-02-29
feat(core): new onboarding template

EYHN:


EYHN:
* **#5951** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5951?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/5951?utm_source=stack-comment).

packages/frontend/templates/onboarding/blob.json
pengx17, line 1: Not sure about this but importing blob as base64 will increase the memory footprint.
But I guess the template solution will be refactored later using a standalone repo tho 🤔

pengx17 (approved):

packages/frontend/templates/onboarding/blob.json
EYHN: I found that the previous method of placing the images in the /static/ folder caused the old version client can't load the images created by the new version client.

So I did some processing and let the blob of the onboarding template be saved to blobstorage for synchronization.

I believe this is just a temporary solution.

Brooooooklyn (approved):

PR #5961, opened 2024-02-29
feat(server): make captcha modular

darkskygit:


darkskygit:
* **#6002** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6002?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5961** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5961?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#8032** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8032?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7999** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7999?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#8019](https://github.com/toeverything/AFFiNE/pull/8019) <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8019?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#7994** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7994?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`

<h2></h2>

packages/backend/server/src/config/affine.ts
```diff
@@ -80,21 +80,29 @@ AFFiNE.port = 3010;
 // ###############################################################
 // ##                        Plugins settings                   ##
 // ###############################################################
-//
-// /* Redis Plugin */
-// /* Provide caching and session storing backed by Redis. */
-// /* Useful when you deploy AFFiNE server in a cluster. */
+
```
forehalo: I think it's ok to disable all the plugins by default, enable the one affine required in affine.self.ts

packages/backend/server/src/plugins/captcha/service.ts
```diff
@@ -39,11 +49,71 @@ export class CaptchaService {
...
-  async verifyChallengeResponse(response: any, resource: string) {
+  private async verifyChallengeResponse(response: any, resource: string) {
     return verifyChallengeResponse(
       response,
       this.captcha.challenge.bits,
       resource
     );
   }
+
+  async getChallengeToken() {
```
forehalo, line 67: use auth/TokenService

packages/backend/server/src/core/auth/controller.ts
```diff
@@ -156,6 +150,8 @@ export class AuthController {
 
     const valid = await this.token.verifyToken(TokenType.SignIn, token, {
       credential: email,
+      // don't revoke the token to avoid email client prefetching touch link
+      keep: true,
```
forehalo: then when will we delete the token?

packages/backend/server/src/plugins/captcha/types.ts
```diff
@@ -0,0 +1,32 @@
+import { Field, ObjectType } from '@nestjs/graphql';
+
+export interface CaptchaConfig {
+  /**
+   * whether to enable captcha
+   */
+  enable: boolean;
```
forehalo: `enable` is not needed

packages/backend/server/src/core/auth/controller.ts
darkskygit: will deleted after click `open affine`/ `submit questionnaire` in onboarding page
frontend change will commit later

forehalo: that makes no sense. the token should be expired once the user open the link.

I forget the mailer prefetching stuff. I will make a new page with and sending a POST request to login to replace this api

forehalo: I checked next-auth's email callback impl, but I didn't see any protect against mailer link prefetching stuff. I will double check the necessity

packages/backend/server/src/core/auth/token.ts
```diff
@@ -47,34 +49,42 @@ export class TokenService {
...
-      return null;
-    }
-
-    const expired = record.expiresAt <= new Date();
-    const valid =
-      !expired && (!record.credential || record.credential === credential);
-
-    if ((expired || valid) && !keep) {
-      await this.db.verificationToken.delete({
+    return await this.db.$transaction(async tx => {
```
forehalo: revoke the changes and I will bring it to upstream commit.

packages/backend/server/src/core/auth/controller.ts
darkskygit: most email clients provide link preview feature:
image

and the browser's preloading function may also preload it:
image

previously, some users report that the email login link will be invalid randomly, and this has also occurred in our internal tests

therefore, we should postpone the timing of revoking the token and providing the session until after the page is fully loaded in a normal browser

forehalo: cool, I will do the workaround, you can revoke it here

packages/backend/server/src/config/affine.ts
```diff
@@ -80,21 +80,27 @@ AFFiNE.port = 3010;
...
+/* Payment Plugin Default Config */
 AFFiNE.plugins.use('payment', {
   stripe: { keys: {}, apiVersion: '2023-10-16' },
 });
-//
-//
+
+/* Captcha Plugin Default Config */
+AFFiNE.plugins.use('captcha', {
```
forehalo: disable here and use it in `affine.self.ts`

forehalo (commented):
frontend adaptation?

darkskygit:
> frontend adaptation?

fixed

packages/backend/server/src/plugins/captcha/service.ts
```diff
@@ -0,0 +1,119 @@
...
+    res.status(status);
+    res.json({
+      url: `${this.config.baseUrl}/api/auth/error?${new URLSearchParams({
+        error,
+      }).toString()}`,
+      error,
+    });
+  }
+
+  async verifyRequest(req: Request, res: Response): Promise<boolean> {
```
forehalo: the backend verification logic is missing?

packages/backend/server/src/core/auth/resolver.ts
```diff
@@ -48,7 +49,8 @@ export class AuthResolver {
   constructor(
     private readonly config: Config,
     private readonly auth: AuthService,
-    private readonly token: TokenService
+    private readonly token: TokenService,
+    private readonly captcha?: CaptchaService
```
forehalo: I think the core module should not be aware of plugin modules
let me find a better solution for this

darkskygit: how do you think about this usage:

``` ts
// register at global
CHALLENGES.register(Captcha);

// guard request in resolver
@UseChallengeGuard()
async somethingWithChallenge() {
}
```

forehalo: or, maybe it's ok to make captcha a standard service and let cloudflare captcha override it.
the captcha verification could receive the whole headers as parameter instead of specified parameters.

forehalo: it's cool but it doesn't has to be registered manually. the injection could be `@Optional()`.
and also I don't like the way which by extending the args of APIs need to be protected

darkskygit: because the interface has been removed, the problem no longer exists

packages/backend/server/src/core/auth/controller.ts
```diff
@@ -52,6 +52,8 @@ export class AuthController {
   ) {}
 
   @Public()
+  @UseBasicGuard()
+  @Guard('captcha')
```
forehalo: looks verbose and we can't register more than one guard implementation for basic guard. I think it could be merged as `@CaptchaGuard`

darkskygit: now merged into single decorator

packages/frontend/core/src/components/affine/auth/use-captcha.tsx
```diff
@@ -87,9 +98,9 @@ export const useCaptcha = (): [string | undefined, string?] => {
           console.error('Error getting challenge response:', err);
         });
     }
-  }, [challenge, setResponse]);
+  }, [challenge, hasCaptchaFeature, setResponse]);
 
-  if (!runtimeConfig.enableCaptcha) {
+  if (!hasCaptchaFeature) {
```
forehalo, line 103: should we avoid cloudflare script loading if captcha feature is not enabled?

packages/backend/server/src/core/auth/token.ts
```diff
@@ -86,6 +84,16 @@ export class TokenService {
     return valid ? record : null;
   }
 
+  async revokeToken(type: TokenType, token: string, tx?: Transaction) {
+    const client = tx || this.db;
```
forehalo: in which case transaction is reuqired

darkskygit: legacy code, let me cleanup

darkskygit: done

forehalo (commented):
will merge in iteration 0.17 together with login refactoring

forehalo (approved):

PR #5968, opened 2024-02-29
feat(core): set journal's really created time

### LINKED ISSUE #5958: Journals create time error​
### What happened?

Journals create time error, I create it on 14:24, but the app shows that the create time is 00:00

image

### Distribution version

Linux

Chrome

### Relevant log output

_No response_

### Anything else?

_No response_

- [X] Yes I'd like to help by submitting a PR!

CatsJuice:
closes #5958;

This makes the creation date of the journal correspond to the date, and the creation time is the real creation time.

### DIFF
--- a/packages/frontend/core/src/hooks/use-journal.ts
+++ b/packages/frontend/core/src/hooks/use-journal.ts
@@ -31,11 +31,17 @@ export const useJournalHelper = (workspace: BlockSuiteWorkspace) => {
    */
   const _createJournal = useCallback(
     (maybeDate: MaybeDate) => {
-      const title = dayjs(maybeDate).format(JOURNAL_DATE_FORMAT);
+      const day = dayjs(maybeDate);
+      const title = day.format(JOURNAL_DATE_FORMAT);
       const page = bsWorkspaceHelper.createDoc();
       // set created date to match the journal date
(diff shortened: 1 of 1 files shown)


CatsJuice:
* **#5968** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5968?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

EYHN (commented):
I think it's a bit strange to modify the creation time. Do we really need to modify the creation time? @hwangdev97

hwangdev97:
> I think it's a bit strange to modify the creation time. Do we really need to modify the creation time? @hwangdev97

after hidden the empty journal on page list, maybe we can show the real creation time 🤔. Do you think so @CatsJuice

CatsJuice:
> > I think it's a bit strange to modify the creation time. Do we really need to modify the creation time? @hwangdev97
> 
> after hidden the empty journal on page list, maybe we can show the real creation time 🤔. Do you think so @CatsJuice 

Even if we use the real creation time, there still could be multiple journals created in one day. I think our purpose to modify created time is to avoid this situation.

Brooooooklyn (approved):

PR #5971, opened 2024-02-29
feat(core): split right sidebar

EYHN:
https://github.com/toeverything/AFFiNE/assets/13579374/c846c069-aa32-445d-b59b-b773a9b05ced

Now each view has a general container, the yellow area is the general container part, and the green part is the routing specific part.


EYHN:
* **#5994** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5994?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5971** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5971?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/modules/right-sidebar/entities/right-sidebar.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  toggle() {
+    this.isOpen.next(!this.isOpen.value);
+  }
+
+  close() {
+    this.isOpen.next(false);
+  }
+
+  _append(view: RightSidebarView) {
```
pengx17, line 28: Why prefixed with `_`?

packages/frontend/core/src/components/blocksuite/block-suite-editor/blocksuite-editor-container.tsx
```diff
@@ -88,7 +89,7 @@ const useBlockElementById = (
     return () => {
       canceled = true;
     };
-  }, [container, blockId, timeout]);
+  }, [blockId, containerRef, timeout]);
```
pengx17, line 92: I guess `containerRef` should be `containerRef.current`
```suggestion
  }, [blockId, containerRef.current, timeout]);
```

packages/frontend/core/src/utils/island.tsx
```diff
@@ -0,0 +1,41 @@
+import { LiveData, useLiveData } from '@toeverything/infra/livedata';
+import { useEffect, useRef } from 'react';
+import { createPortal } from 'react-dom';
+
+export const createIsland = () => {
```
pengx17, line 5: Wow clever

packages/frontend/core/src/utils/island.tsx
```diff
@@ -0,0 +1,41 @@
...
+          targetLiveData.next(null);
+        };
+      }, []);
+      return <div {...other} ref={target}></div>;
+    },
+    Provider: ({ children }: React.PropsWithChildren) => {
+      const target = useLiveData(targetLiveData);
+      useEffect(() => {
+        if (provided === true) {
+          throw new Error('Island should not be provided more than once');
```
pengx17, line 29: I guess this could be a dev only error

pengx17 (commented):
LGTM overall. It seems some UI details may need to be fixed later.
Also, perhaps adding a couple of more tests?

packages/frontend/component/src/components/workspace/index.css.ts
```diff
@@ -66,9 +65,6 @@ export const mainContainerStyle = style({
         },
       },
     },
-    '&[data-show-padding="true"][data-is-macos="true"]': {
```
CatsJuice, line 69: This shouldn't be removed, since it should be rounded for macOS when clientBorder is enabled

packages/frontend/component/src/theme/global.css
```diff
@@ -266,14 +266,6 @@ affine-block-hub {
   }
 }
 
-button,
```
CatsJuice, line 269: Why to remove?

packages/frontend/component/src/components/workspace/index.css.ts
EYHN: This rounded corner should be implemented on the view, not here.

packages/frontend/component/src/theme/global.css
EYHN: CleanShot 2024-03-04 at 09 56 48
Because there is a hidden button that needs to be able to drag.
So i deleted this global rule. and found that no-drag is defined locally in most places, so this global rule does not affect the features.

packages/frontend/component/src/components/workspace/index.css.ts
EYHN: But I haven’t implemented rounded corners yet, it needs to change with the border settings and os, you can implement it in your PR

CatsJuice: I've used `inherit` in view so that it don't need to care about whether `clientBorder` is enabled. But if you remove it here, I think it's also ok to manage this in view and I'll do that later.

packages/frontend/component/src/theme/global.css
CatsJuice: Maybe it is better to override for this button? I'm not sure whether `pointer-events: none` worked here. I remember that @pengx17  has added this to the `button`, but maybe not it's parent wrapper.

pengx17: This was added for buttons or alike that are in blocksuite. Please do not remove these rules.

packages/frontend/core/src/components/blocksuite/block-suite-editor/blocksuite-editor-container.tsx
EYHN: eslint will warning when `containerRef.current`, there are same, as containerRef.current never change in this the component lifecycle.

packages/frontend/core/src/modules/right-sidebar/entities/right-sidebar.ts
EYHN: This means that this function should not be called outside this "module".
should use RightSidebarViewIsland instead.

EYHN: I have added some comment here.

packages/frontend/component/src/theme/global.css
pengx17: This may be reverted

pengx17 (approved):

CatsJuice (approved):

PR #5979, opened 2024-03-01
fix(server): wrap read-modify-write apis with distributed lock

darkskygit:
impl new distributed lock with `using` syntax:

```typescript
{
   // lock is acquired here
   await using lock = await mutex.lock('resource-key');
   if (lock) {
     // do something
   } else {
     // failed to lock
   }
}
// lock is released here
```


darkskygit:
* **#5979** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/5979?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

forehalo:
this is a bad idea, make all the db queries must be aware of query engine.

darkskygit:
> this is a bad idea, make all the db queries must be aware of query engine.

currently prisma only support get a transaction by Prisma.$transaction

but i just found a prisma [transaction plugin](https://papooch.github.io/nestjs-cls/plugins/available-plugins/transactional) for nestjs, what are your thoughts on this usage?

forehalo:
what about a resource lock for mutating data, for example:

```js
function invite(workspaceId, userId) {
  // this is good because it tell us there may be concurrency issue if the resource not locked
  // we can safely loop query the lock in a given time limit like 100ms(is 100ms enough for any table mutating?), and block the lock until the last use get released.
  // if the time limit passes by, force rewrite the lock and go ahead
  const lock = await this.mutex.lock('members:' + workspaceId)
  
  try {
    await addUserToWorkspace(workspaceId, userId)
  } finally {
    lock.release()
  }

  // ...or
  await this.mutex.lock('members:' + workspaceId, async () => {
  //                    ^ maybe strong typed so we won't get lost if different callers request for a same lock
    await addUserToWorkspace(workspaceId, userId)
  })
}

```

darkskygit:
> what about a resource lock for mutating data, for example:
> 
> ```js
> function invite(workspaceId, userId) {
>   // this is good because it tell us there may be concurrency issue if the resource not locked
>   // we can safely loop query the lock in a given time limit like 100ms(is 100ms enough for any table mutating?), and block the lock until the last use get released.
>   // if the time limit passes by, force rewrite the lock and go ahead
>   const lock = await this.mutex.lock('members:' + workspaceId)
>   
>   try {
>     await addUserToWorkspace(workspaceId, userId)
>   } finally {
>     lock.release()
>   }
> 
>   // ...or
>   await this.mutex.lock('members:' + workspaceId, async () => {
>   //                    ^ maybe strong typed so we won't get lost if different callers request for a same lock
>     await addUserToWorkspace(workspaceId, userId)
>   })
> }
> ```

if we use mutex, then it needs to be a distributed lock
invite include querying and writing to multiple table, directly locking the entire resource can indeed make the logic simpler

but for single table read-modify-write(e.g. https://github.com/toeverything/AFFiNE/blob/aeb9f69b59f150b4baf2bd62e918bd4689baaec8/packages/backend/server/src/core/features/service.ts#L70), i still prefer use transaction, that allow database to optmizie queries

packages/backend/server/src/core/workspaces/resolvers/workspace.ts
```diff
@@ -336,6 +338,11 @@ export class WorkspaceResolver {
       throw new ForbiddenException('Cannot change owner');
     }
 
+    const lockFlag = `invite:${workspaceId}`;
+    if (!(await this.mutex.lock(lockFlag))) {
```
forehalo: make the execution after lock acquired as a callback? there are sometimes lots of return paths. I concerns there might be chances to forget to release the locks.

darkskygit:
consider that the auth module is refacting, restack the throttler refactor branch and develop it based on the auth branch but not this branch

packages/backend/server/src/fundamentals/mutex/index.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (locked) {
+      let result: R;
+      try {
+        result = await cb();
+      } finally {
+        await this.unlock(key);
+      }
+      return result;
+    } else {
+      throw new ForbiddenException('Failed to acquire lock');
```
forehalo: This error would be confusing because it will be directly shown to user. 
What if we pass lock fail error as parameter to callback just in the old school way, or you have a better idea?

packages/backend/server/src/fundamentals/mutex/index.ts
```diff
@@ -0,0 +1,87 @@
...
+      const current = this.bucket.get(key);
+      if (current && current !== id) {
+        this.logger.warn(
+          `Failed to fetch lock ${key}, retrying in ${MUTEX_WAIT} ms`
+        );
+        await sleep(MUTEX_WAIT * (MUTEX_RETRY - retry + 1));
+        return fetchLock(retry - 1);
+      }
+      this.bucket.set(key, id);
+      console.error('success lock', key);
```
forehalo: Console logs should be cleaned up

packages/backend/server/src/app.module.ts
```diff
@@ -39,6 +41,12 @@ export const FunctionalityModules = [
   ScheduleModule.forRoot(),
   EventModule,
   CacheModule,
+  ClsModule.forRoot({
```
forehalo: Provide some description about how to use it? And I didn't see any usage yet, should we import clsmodule right now?

packages/backend/server/src/fundamentals/mutex/index.ts
darkskygit: i think errors thrown by service should still be directly related to the error
but we can capturing error in the API  and return an easy-to-understand message to the user (such as "Server is busy, please try again later")
and the original error thrown by service can be process after (for example, output as log):

``` js
try {
    await this.mutex.lockWith('some scope', ()=> {
        // do something
    })
} catch(e) {
    this.logger.error('log something', e.message)
    return new TooManyRequestsException('Server is busy')
}
```

packages/backend/server/src/app.module.ts
darkskygit: its a nestjs binding for [async local storage](https://nodejs.org/api/async_context.html)
that can allow use share some info across the async context, like thread local storage
we can use this context to distinguish multiple async requests that accessing same resource
https://github.com/toeverything/AFFiNE/blob/5c5208bee17ddec642085e033a26706c635cda18/packages/backend/server/src/plugins/redis/mutex.ts#L31-L45

forehalo: I know what it is. If it's not used by now, we should not only import the module

packages/backend/server/src/fundamentals/utils/utils.ts
```diff
@@ -0,0 +1,3 @@
+export function sleep(ms: number): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, ms));
```
Brooooooklyn: `import { setTimeout } from 'node:timers/promise'`

`await setTimeout(100)`

Brooooooklyn:
How about expose `using` API:

```ts
await using _ = await this.mutex.lock(`key`);
```

darkskygit:
> How about expose `using` API:
> 
> ```ts
> await using _ = await this.mutex.lock(`key`);
> ```

looks more elegant than the existing solution, I will add support for this usage

packages/backend/server/src/core/auth/service.ts
```diff
@@ -231,7 +231,7 @@ export class AuthService {
 
     return this.prisma.user.create({
       data: {
-        name: 'Unnamed',
+        name: email.split('@')[0],
```
Brooooooklyn: we need to run a script to make all exists `Unnamed` data align to this behavior

packages/backend/server/src/core/workspaces/resolvers/workspace.ts
```diff
@@ -336,72 +339,85 @@ export class WorkspaceResolver {
...
-        where: { workspaceId },
-      }),
-      this.quota.getWorkspaceUsage(workspaceId),
-    ]);
-    if (memberCount >= quota.memberLimit) {
-      throw new PayloadTooLargeException('Workspace member limit reached.');
-    }
+    try {
+      const lockFlag = `invite:${workspaceId}`;
+      return this.mutex.lockWith(lockFlag, async () => {
```
Brooooooklyn: We already have an X-Request-Id header on Request, and it appears to be retrievable from that location. I’m wondering if we could use this existing ID instead of generating a new CLS UUID, as relying on AsyncLocalStorage could significantly impact performance: https://github.com/nodejs/node/issues/34493

packages/backend/server/src/core/auth/service.ts
darkskygit: already have a migration script about this, i will rerun after depoly to stable: https://github.com/toeverything/AFFiNE/blob/canary/packages/backend/server/src/data/migrations/1703756315970-unamed-account.ts

packages/backend/server/src/core/workspaces/resolvers/workspace.ts
darkskygit: `X-Request-Id` is generate by client, to avoid replay attacks we should use server generated ids here
i would check if there is any other way to generate ids that can differentiate between requests

Brooooooklyn: So we can generate uuid in the middleware level rather than the cls, I still worrying about the performance impact of introducing by cls

Brooooooklyn (approved):

PR #6034, opened 2024-03-06
fix: backlinks jump

regischen:
Fix [AFF-656](https://linear.app/affine-design/issue/AFF-656/back-links-看起来都坏了)


packages/frontend/core/src/components/affine/reference-link/index.tsx
```diff
@@ -62,10 +62,7 @@ export function AffinePageReference({
   });
 
   return (
-    <Link
-      to={`/workspace/${workspace.id}/${pageId}`}
-      className={styles.pageReferenceLink}
-    >
+    <Link to={`/${pageId}`} className={styles.pageReferenceLink}>
```
EYHN: pls use `<WorkbenchLink />`

regischen: done

EYHN (approved):

Brooooooklyn (approved):

PR #6056, opened 2024-03-10
chore(component): simplify code

kamuiiiii:


packages/frontend/component/src/components/setting-components/workspace-detail-skeleton.tsx
```diff
@@ -8,19 +8,17 @@ export const WorkspaceDetailSkeleton = () => {
     <>
       <SettingHeader title={<Skeleton />} subtitle={<Skeleton />} />
 
-      {Array.from({ length: 3 })
```
Brooooooklyn, line 11: you can also move the `map` fn into the second argument of `Array#from` to make it shorter

pengx17 (approved):

PR #6072, opened 2024-03-12
fix(core): fix page priority load

EYHN:
When refactoring, I planned to put the logic of loading current pages priority in `PageManager.open`, but I forgot, so the current page will not be loaded priority.

But currently `PageManager.open` has been used in many places, but only one priority page can be set, so the logic of setting the priority page is still placed in the detail page.


EYHN:
* **#6072** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6072?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

pengx17 (approved):

PR #6079, opened 2024-03-12
refactor(core): refactor tag to use di

JimmFly:
use case
```
const tagService = useService(TagService);
const tags = useLiveData(tagService.tags);
const currentTagLiveData = tagService.tagByTagId(tagId);
const currentTag = useLiveData(currentTagLiveData);

```


JimmFly:
* **#6079** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6079?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/modules/tag/service.ts
```diff
@@ -0,0 +1,34 @@
+import type { Workspace } from '@toeverything/infra';
```
EYHN: move to 'packages/frontend/core/src/modules/tag/service/tag.ts'

packages/frontend/core/src/modules/tag/service.ts
```diff
@@ -0,0 +1,34 @@
...
+import type { WorkspaceLegacyProperties } from '../workspace';
+import { Tag } from './entities/tag';
+
+export class TagService {
+  constructor(
+    private readonly properties: WorkspaceLegacyProperties,
+    private readonly workspace: Workspace
+  ) {}
+
+  readonly tags = this.properties.tagOptions$;
```
EYHN: services should return entity instance

```suggestion
  readonly tags = this.properties.tagOptions$.map(tags => 
    tags.map(
      tag => new Tag(tag.id, tag.name, this.properties, this.workspace)
    )
  );
```

packages/frontend/core/src/modules/tag/service/tag.ts
```diff
@@ -0,0 +1,72 @@
...
+  constructor(
+    private readonly properties: WorkspaceLegacyProperties,
+    private readonly pageRecordList: PageRecordList
+  ) {}
+
+  readonly tags = this.properties.tagOptions$.map(tags =>
+    tags.map(tag => new Tag(tag.id, this.properties, this.pageRecordList))
+  );
+
+  createTag(value: string, color: string, pageId?: string) {
```
EYHN: This api can be divided into 2 steps?

```
const newTag = createTag("xxx", "#fff")
newTag.tag(pageId)
```

Lets keep the API as simple as possible

packages/frontend/core/src/components/page-list/tags/create-tag.tsx
```diff
@@ -33,30 +36,54 @@ export const CreateOrEditTag = ({
...
   const [menuOpen, setMenuOpen] = useState(false);
-  const [tagName, setTagName] = useState(tagMeta?.title || '');
-  const [activeTagIcon, setActiveTagIcon] = useState(() => {
-    return (
-      tagColors.find(([_, color]) => color === tagMeta?.color) ||
-      randomTagColor()
-    );
-  });
+
+  const tagNameLiveData = useMemo(
```
EYHN: why not useState ?

```suggestion
const [tagName, setTagName] = useState(tagMeta?.title)
```

packages/frontend/core/src/components/page-list/tags/create-tag.tsx
```diff
@@ -33,30 +36,54 @@ export const CreateOrEditTag = ({
...
+  );
+  const tagName = useLiveData(tagNameLiveData);
+  const handleChangeName = useCallback(
+    (value: string) => {
+      tagNameLiveData.next(value);
+    },
+    [tagNameLiveData]
+  );
+
+  const tagIconLiveData = useMemo(
```
EYHN: I feel like useState can also be used here instead of livedata

packages/frontend/core/src/components/page-list/docs/page-tags.tsx
```diff
@@ -26,12 +30,17 @@ interface TagItemProps {
 
 export const TagItem = ({
   tag,
+  tagId,
```
EYHN: You only need tag to get the following props. Why do you need to add these props?

packages/frontend/core/src/components/page-list/docs/page-list-header.tsx
```diff
@@ -95,8 +96,17 @@ export const TagPageListHeader = ({
...
+  const service = useService(TagService);
+  const tagMetas: TagMeta[] = useLiveData(service.tagMetas);
+
+  const liveData = LiveData.computed(get => {
+    return {
+      tagColor: get(tag.color),
+      tagTitle: get(tag.value),
+    };
+  });
+  const { tagColor, tagTitle } = useLiveData(liveData);
```
EYHN:
```suggestion
  const tagColor = useLiveData(tag.color);
  const tagTitle = useLiveData(tag.value);
```

packages/frontend/core/src/components/page-list/docs/page-list-header.tsx
```diff
@@ -175,24 +185,26 @@ export const TagPageListHeader = ({
...
 
-export const TagsEditor = ({ options, onClick }: TagsEditorProps) => {
+export const TagsEditor = ({ tagMetas, onClick }: TagsEditorProps) => {
   const t = useAFFiNEI18N();
   const [inputValue, setInputValue] = useState('');
   const filteredOptions = useMemo(
     () =>
-      options.filter(o => (inputValue ? filterOption(o, inputValue) : true)),
-    [inputValue, options]
+      tagMetas.filter(tagMeta =>
```
EYHN: you can put the filter function into TagService

```ts
class TagService {
  filterByName(name: string) {
    this.tags.map((tags) => {
      ...do filter
      return tags
    })
  }
}
```

then this `<TagsEditor/>` dont need `tagMetas` props

packages/frontend/core/src/components/page-list/docs/page-list-header.tsx
```diff
@@ -175,24 +185,26 @@ export const TagPageListHeader = ({
...
 };
 
 interface TagsEditorProps {
-  options: Tag[];
+  tagMetas: TagMeta[];
   onClick: (open: boolean) => void;
 }
 
-export const TagsEditor = ({ options, onClick }: TagsEditorProps) => {
+export const TagsEditor = ({ tagMetas, onClick }: TagsEditorProps) => {
```
EYHN: rename this

packages/frontend/core/src/components/page-list/tags/create-tag.tsx
```diff
@@ -33,30 +36,54 @@ export const CreateOrEditTag = ({
...
+    () =>
+      new LiveData(
+        tagColors.find(
+          ([_, color]) => color === tagColorMap(tagMeta?.color || '')
```
EYHN: put this to 'Tag' entity

packages/frontend/core/src/modules/tag/service/tag.ts
```diff
@@ -0,0 +1,72 @@
...
+import type { WorkspaceLegacyProperties } from '../../workspace';
+import { Tag } from '../entities/tag';
+
+export class TagService {
+  constructor(
+    private readonly properties: WorkspaceLegacyProperties,
+    private readonly pageRecordList: PageRecordList
+  ) {}
+
+  readonly tags = this.properties.tagOptions$.map(tags =>
```
pengx17, line 13: Have we agreed upon livedata var/field naming? @EYHN

packages/frontend/core/src/modules/tag/service/tag.ts
```diff
@@ -0,0 +1,72 @@
...
+    return LiveData.computed(get => {
+      const pageRecord = get(this.pageRecordList.record(pageId));
+      if (!pageRecord) return [];
+      const tagIds = get(pageRecord.meta).tags;
+
+      return get(this.tags).filter(tag => tagIds.includes(tag.id));
+    });
+  }
+
+  tagByTagId(tagId?: string) {
```
pengx17, line 51: I think this should be this.tags.map ... instead of using computed if there is only a single livedata is involved

packages/frontend/core/src/components/affine/page-properties/tags-inline-editor.tsx
```diff
@@ -171,25 +170,49 @@ export const EditTagMenu = ({
...
 
   return <Menu {...menuProps}>{children}</Menu>;
 };
 
 export const TagsEditor = ({
+  pageId,
   options,
   value,
-  onChange,
-  onOptionsChange,
```
pengx17: I believe you do not need to change TagsEditor & InlineTagsList at all. You can just wrap the TagService into onChange & onOptionsChange where TagsEditor is being used

packages/frontend/core/src/modules/tag/service/tag.ts
EYHN: Yes, I will rename them in later pr.

packages/frontend/core/src/components/affine/page-properties/tags-inline-editor.tsx
EYHN: I think it is reasonable to use service and tag entity to make modifications.

The entity contains the business logic, rules, and verification of how to modify the tag.

In crdt applications we'd better clarify the business logic of each operation, such as
`Create tag`, `Rename tag`, `Change tag color`, then convert it into a suitable crdt operation at the domain layer.

And `onOptionsChange([...oldTags, { ...newTag }])` just meaning nothing in business logic.  And it will overwrite the modifications of other clients.

pengx17: Portability-wise it might not be a good idea to call entity methods deeply inside of an UI view.
IMO it might be worth adding a glue layer that converts Tag entity into the required callbacks/APIs that required by TagsEditor. If there is anything that is not that well fit into good CRDT practices, we can make changes to the exposed props of TagsEditor instead, like exposing various callbacks like `onCreateTag`, `onRenamingTag` etc

EYHN: Oh, I agree with this.. 

But there's also the question here is we should lifting up the side effect to which level.

If it is just lift up to PagePropertyValue (the upper level of this file), it seems to be no big difference.

It would be nice to lift up to a higher level and make the PageProperty pure.

But currently our tag is not stored as a regular page property, so it is not easy to implement..

EYHN (approved):

PR #6082, opened 2024-03-12
refactor(core): split web entry from core

Brooooooklyn:
This pr is trying to split `web` and `electron` entries from `core`. It allows more platform-related optimization to be addressed in each entry.
We should remove all browser/electron only codes from `core` eventually, this is the very first step for that.


packages/frontend/core/src/app.tsx
```diff
@@ -61,7 +61,7 @@ const services = new ServiceCollection();
 configureWebServices(services);
 const serviceProvider = services.provider();
 
-export const App = memo(function App() {
+export const App = function App() {
```
forehalo:
```diff
- export const App = function App() {
+ export function App() {
```

forehalo: I think it's unnecessary now. moved to `web/app.tsx`?

pengx17 (approved):

PR #6084, opened 2024-03-12
feat(infra): livedata from yjs

EYHN:

PR #6083, opened 2024-03-12
feat(infra): livedata flatten

EYHN:

PR #6091, opened 2024-03-12
feat(infra): computed livedata

EYHN:


EYHN:
* **#6091** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6091?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6083** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6083?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6084](https://github.com/toeverything/AFFiNE/pull/6084) <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6084?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `canary`

<h2></h2>

packages/common/infra/src/livedata/index.ts
```diff
@@ -170,6 +170,67 @@ export class LiveData<T = unknown> implements InteropObservable<T> {
...
+    compute: (get: <L>(data: LiveData<L>) => L) => T
+  ): LiveData<T> {
+    return LiveData.from(
+      new Observable(subscribe => {
+        const execute = (next: () => void) => {
+          const subscriptions: Subscription[] = [];
+          const getfn = <L>(data: LiveData<L>) => {
+            let value = null as L;
+            let first = true;
+            subscriptions.push(
```
CatsJuice, line 161: Can we deduplicate the same deps?

EYHN: Subscribing to the same livedata multiple times actually shares the same subscription internally, so it is safe.

CatsJuice (approved):

packages/common/infra/src/livedata/index.ts
EYHN: see this test
https://github.com/toeverything/AFFiNE/pull/6091/files#diff-8260747d906aadc2441485759654a60aac8a7d00449af79ddbbb61b77307f180R286

packages/common/infra/src/livedata/index.ts
```diff
@@ -170,6 +170,67 @@ export class LiveData<T = unknown> implements InteropObservable<T> {
...
+   */
+  static computed<T>(
+    compute: (get: <L>(data: LiveData<L>) => L) => T
+  ): LiveData<T> {
+    return LiveData.from(
+      new Observable(subscribe => {
+        const execute = (next: () => void) => {
+          const subscriptions: Subscription[] = [];
+          const getfn = <L>(data: LiveData<L>) => {
+            let value = null as L;
```
pengx17, line 159: init it with data.value?

packages/common/infra/src/livedata/index.ts
```diff
@@ -170,6 +170,67 @@ export class LiveData<T = unknown> implements InteropObservable<T> {
...
+            let first = true;
+            subscriptions.push(
+              data.subscribe(v => {
+                value = v;
+                if (!first) {
+                  next();
```
pengx17: not sure if this will cause infinite loops, saying two computed data depends on each other:

```js
      let c;
      const b = LiveData.computed(get => get(c) + 1);
      c = LiveData.computed(get => get(b) + 1);
```

this could be not that obvious in real scenarios. maybe we could break and report for errors when reaching a number of loops

EYHN: I added a test and will throw error in this case

packages/common/infra/src/livedata/index.ts
EYHN: Since livedata.subscribe will output a value immediately, it is the same.

pengx17 (approved):

packages/common/infra/src/livedata/index.ts
```diff
@@ -132,10 +132,100 @@ export class LiveData<T = unknown> implements InteropObservable<T> {
...
+   * const v2 = new LiveData(200);
+   *
+   * const v = LiveData.computed(get => {
+   *   return get(a) === 'v1' ? get(v1) : get(v2);
+   * });
+   *
+   * expect(v.value).toBe(100);
+   * ```
+   */
+  static computed<T>(
```
CatsJuice, line 151: Will this compare computed res with previous? Since in some case, although some deps changed, but the computed res is same.

EYHN: currently no compare,

I think we can do compare here

PR #6093, opened 2024-03-12
feat(core): add split view to experimental features settings

pengx17:


pengx17:
* **#6113** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6113?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6093** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6093?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6090** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6090?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`

<h2></h2>

packages/backend/server/src/core/user/resolver.ts
```diff
@@ -108,6 +108,15 @@ export class UserResolver {
     });
   }
 
+  @Throttle({ default: { limit: 10, ttl: 60 } })
+  @ResolveField(() => Boolean, {
+    name: 'earlyAccess',
+    description: 'Is user in early access',
+  })
+  async earlyAccess(@CurrentUser() user: CurrentUser) {
```
forehalo: why not add it to `availableFeatures`?

pengx17: I think earlyAccess is a rule based attribute for a user, but availableFeatures in current design is for a certain cloud workspace. They may be related somehow but I guess it may not strictly belongs to availableFeatures or the other way around.

darkskygit: its recommand use `this.feature.getUserFeatures(...).then(features=>features.filter(f=>f.activated).map(f=>f.feature.name))`, return the feature name to frontend, then check in frontend
in future we may have more user feature

packages/backend/server/src/core/workspaces/management.ts
```diff
@@ -117,7 +117,7 @@ export class WorkspaceManagementResolver {
   async availableFeatures(
     @CurrentUser() user: CurrentUser
   ): Promise<FeatureType[]> {
-    const isEarlyAccessUser = await this.feature.isEarlyAccessUser(user.email);
+    const isEarlyAccessUser = await this.feature.canEarlyAccess(user.email);
```
darkskygit: why use `canEarlyAccess`, this method will always true if `earlyAccessPreview` flag not enabled

packages/frontend/core/src/modules/workbench/view/workbench-link.tsx
```diff
@@ -1,9 +1,16 @@
+import { useAppSettingHelper } from '@affine/core/hooks/affine/use-app-setting-helper';
 import { useService } from '@toeverything/infra/di';
 import type { To } from 'history';
 import { useCallback } from 'react';
 
 import { Workbench } from '../entities/workbench';
 
+declare global {
```
CatsJuice: I think we can remove `window.enableMultiView` directly

packages/backend/server/src/core/user/resolver.ts
pengx17: Talked with @darkskygit offline. We will expose a userFeatures query instead

EYHN (commented):
This weird that it is set in workspace setting but stored in appsetting, I think you can use `useService(WorkspaceLocalState)` , It is persistent, workspace-level storage

forehalo (approved):

PR #6094, opened 2024-03-12
fix: update docs

pengx17:


pengx17:
* **#6094** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6094?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

docs/developing-server.md
```diff
@@ -61,6 +61,8 @@ MAILER_USER="auth"
 MAILER_PASSWORD="auth"
 MAILER_HOST="localhost"
 MAILER_PORT="1025"
+STRIPE_API_KEY=sk_live_1
```
forehalo, line 64: add a comment "To debug Payment features"

pengx17: This was added to make sure the generated `schema.gql` will include payment related. Not sure if we should ignore schema.gql and related auto-generated files from git?

forehalo: that doesn't make sense as well, the schema generation in frontend requires `schema.gql`

pengx17: should be related to https://github.com/toeverything/AFFiNE/blob/282b788258ae6f3699cd82a751aeda182fd8cc3d/packages/backend/server/src/plugins/payment/index.ts#L21-L24

Without these two env var, when running `yarn dev` the schema.gql will regenerated and exclude the ones that defined in payment module.

forehalo (approved):

PR #6115, opened 2024-03-14
feat(server): allow prefetch doc stats before sync

forehalo:


forehalo:
* **#6118** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6118?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6115** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6115?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

EYHN (approved):

packages/backend/server/src/core/doc/manager.ts
```diff
@@ -303,9 +317,56 @@ export class DocManager implements OnModuleInit, OnModuleDestroy {
...
+   * Get latest timestamp of all docs in the workspace.
+   */
+  @CallTimer('doc', 'get_stats')
+  async getStats(workspaceId: string, after: number | undefined = 0) {
+    const snapshots = await this.db.snapshot.findMany({
+      where: {
+        workspaceId,
+        updatedAt: {
+          gt: new Date(after),
+        },
```
darkskygit, line 336: we can add a (workspaceId, updatedAt) index pair to Snapshot table to improve this query

packages/backend/server/src/core/doc/manager.ts
```diff
@@ -303,9 +317,56 @@ export class DocManager implements OnModuleInit, OnModuleDestroy {
...
+        workspaceId,
+        createdAt: {
+          gt: new Date(after),
+        },
+      },
+      by: ['workspaceId', 'id'],
+      _max: {
+        createdAt: true,
+      },
+    });
```
darkskygit: can improve by `create index idx_update_created_at on update (workspaceId, createAt desc)`

darkskygit (approved):

packages/backend/server/src/core/doc/manager.ts
forehalo: workspaceId index will narrow down the query to a very small set(page list)

packages/backend/server/src/core/doc/manager.ts
forehalo: if so, it should be `Index(workspaceId, guid, createdAt)`

packages/backend/server/src/core/doc/manager.ts
darkskygit: my mistake, i just noticed that prisma has built a separate index for `workspaceId`

packages/backend/server/src/core/doc/manager.ts
darkskygit: agree, add guid can speed up query for specific subdocs, but i still recommend add `desc` sorting for craeteAt -- there should be no scenario where we need to find the oldest snapshot

forehalo: that make sense, but I wondering whether it worth to add such an index, because the stats query won't be too frequent.

forehalo: and what we hope is that updates will be merged as soon as possible so there won't be too many rows in updates table, which decrease the necessity again

PR #6202, opened 2024-03-19
chore(core): add mixpanel track

Brooooooklyn:


Brooooooklyn:
* **#6203** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6203?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6267** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6267?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6202** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6202?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

 Join @Brooooooklyn and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/backend/server/src/schema.gql
```diff
@@ -2,14 +2,6 @@
 # THIS FILE WAS AUTOMATICALLY GENERATED (DO NOT MODIFY)
 # ------------------------------------------------------
 
-input CreateCheckoutSessionInput {
```
pengx17: changes in this file should not be committed.

packages/backend/server/src/app.ts
```diff
@@ -43,5 +43,12 @@ export async function createApp() {
     app.useWebSocketAdapter(adapter);
   }
 
+  if (AFFiNE.isSelfhosted && AFFiNE.telemetry.enabled) {
```
pengx17, line 46: `!AFFiNE.isSelfhosted`?

packages/frontend/core/src/router.tsx
```diff
@@ -1,59 +1,81 @@
...
-    path: '/onboarding',
-    lazy: () => import('./pages/onboarding'),
-  },
-  {
-    path: '*',
-    lazy: () => import('./pages/404'),
+import { mixpanel } from './utils';
+
+function RootRouter() {
+  const location = useLocation();
```
pengx17, line 13: Due to the two layers router design in desktop app, this location change side effect will not be tracked in desktop.
I guess the correct way is to `useLiveData(workbench.location)`? cc @EYHN

packages/frontend/core/src/router.tsx
```diff
@@ -1,59 +1,81 @@
...
+  useEffect(() => {
+    mixpanel.track_pageview({
```
pengx17, line 15: looks like it will reports error if mixpanel is not enabled. maybe we need to check MIXPANEL_TOKEN env before calling this method

pengx17, line 15: looks like it will reports error if mixpanel is not enabled. maybe we need to check MIXPANEL_TOKEN env before calling this method

packages/backend/server/src/app.ts
darkskygit, line 46: only for selfhosted?

Brooooooklyn: yes

packages/frontend/core/src/router.tsx
Brooooooklyn: `mixpanel` here is a proxied object, and it will print a log if `MIXPANEL_TOKEN` is not provided

packages/frontend/core/src/router.tsx
Brooooooklyn: It doesn't matter, let's ship it first

pengx17 (approved):

EYHN (approved):

PR #6205, opened 2024-03-19
feat(infra): new doc sync engine

EYHN:
https://github.com/toeverything/AFFiNE/blob/eyhn/feat/new-sync/packages/common/infra/src/workspace/engine/doc/README.md


EYHN:
* **#6205** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6205?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#6248](https://github.com/toeverything/AFFiNE/pull/6248) <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6248?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#6250](https://github.com/toeverything/AFFiNE/pull/6250) <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6250?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈
* `canary`

<h2></h2>

packages/common/infra/src/storage/kv.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export class ReadonlyByteKV extends MemoryByteKV implements ByteKV {
+  override transaction<T>(
+    _cb: (transaction: ByteKVBehavior) => Promise<T>
+  ): Promise<T> {
+    return Promise.resolve(null as T);
+  }
+  override set(_key: string, _value: Uint8Array): Promise<void> {
+    return Promise.resolve();
+  }
```
forehalo, line 78: override `del` and `clear`?

forehalo (commented):
too complex
I need a architecture design to understand the whole

packages/common/infra/src/storage/kv.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export class MemoryByteKV implements ByteKV {
+  readonly lock = new AsyncLock();
+
+  constructor(readonly db = new Map<string, Uint8Array>()) {}
+
+  async transaction<T>(cb: (transaction: ByteKVBehavior) => Promise<T>) {
+    using _lock = await this.lock.acquire();
+    return cb({
```
pengx17: I think this should be `return await cb(`

Check this example:
https://stackblitz.com/edit/stackblitz-starters-kkgykp?file=package.json,index.ts

packages/frontend/workspace-impl/src/local/doc-indexeddb.ts
```diff
@@ -0,0 +1,243 @@
...
+    return await cb({
+      async get(docId) {
+        const data = await store.get(docId);
+
+        if (!data) {
+          return null;
+        }
+
+        const { updates } = data;
+        const update = mergeUpdates(updates.map(({ update }) => update));
```
pengx17, line 126: I forget the reason why not to use mergeUpdates from yjs at the first place. Is it safe to do so now?

forehalo: there was a bug made the merged binary broken. Actually I don't know whether it was fixed or not

packages/common/infra/src/workspace/engine/doc/__tests__/sync.spec.ts
```diff
@@ -0,0 +1,256 @@
...
+      this.server.listeners.delete(listener);
+    };
+  }
+
+  async waitingForConnectServer(): Promise<void> {}
+  disconnectServer(): void {}
+  onInterrupted(_cb: (reason: string) => void): void {}
+}
+
+describe('sync', () => {
```
doodlewind, line 114: These test cases looks intuitive, can we think about them as examples of the new API surface of the sync engine?

packages/common/infra/src/workspace/engine/doc/index.ts
```diff
@@ -0,0 +1,181 @@
...
+      });
+    }
+  }
+
+  setPriority(docId: string, priority: number) {
+    this.localPart.setPriority(docId, priority);
+    this.remotePart?.setPriority(docId, priority);
+  }
+
+  waitForSaved() {
```
doodlewind, line 135: Wondering how to distinguish the difference between `saved` / `synced` / `ready`? Are they implementation details, or should be sensible for engine users?

packages/common/infra/src/storage/kv.ts
EYHN: It's truly an easy mistake to make

darkskygit: I remember the `no-floating-promises` rule has been enabled, did it not detect this problem? 🤔

packages/common/infra/src/workspace/engine/doc/index.ts
EYHN: * saved: ydoc changes have been saved to storage, and the browser can be safely closed without losing data.

* synced: is fully synchronized with the server.

* ready: is a strange flag, which means that the doc has been loaded and the data can be modified. (is not force, you can still modify it if you know you are creating some new data)
  
  This is a temporary solution to deal with the yjs overwrite issue.
  
  if content is loaded from storage
or if content is pulled from the server, it will be true, otherwise be false.
  
  For example, when opening a doc that is not in storage, ready = false until the content is pulled from the server.

packages/common/infra/src/storage/kv.ts
pengx17: I guess since it is returned it will not be considered floating, but it might be the case for a using clause

EYHN:
> too complex I need a architecture design to understand the whole

https://github.com/toeverything/AFFiNE/blob/eyhn/feat/new-sync/packages/common/infra/src/workspace/engine/doc/README.md

packages/common/infra/src/workspace/engine/doc/__tests__/sync.spec.ts
EYHN: Yes, this is an API example

packages/frontend/workspace-impl/src/local/doc-indexeddb.ts
EYHN: https://github.com/yjs/yjs/issues/558

Looks like it's been fixed

packages/common/infra/src/workspace/engine/doc/__tests__/sync.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+    }) => void
+  ): Promise<() => void> {
+    const listener = { cb, clientId: this.id };
+    this.server.listeners.add(listener);
+    return () => {
+      this.server.listeners.delete(listener);
+    };
+  }
+
+  async waitingForConnectServer(): Promise<void> {}
```
pengx17: I think the name should be `waitForConnectingServer`

packages/backend/server/src/core/workspaces/controller.ts
```diff
@@ -51,7 +51,7 @@ export class WorkspacesController {
     // metadata should always exists if body is not null
     if (metadata) {
       res.setHeader('content-type', metadata.contentType);
-      res.setHeader('last-modified', metadata.lastModified.toISOString());
+      res.setHeader('last-modified', metadata.lastModified.toUTCString());
```
pengx17, line 54: seems last modified has a standard format
https://arc.net/l/quote/wradjpwu
better to follow it?

pengx17 (approved):

packages/backend/server/src/core/workspaces/controller.ts
EYHN: Looks like that format is the same as toUTCString

packages/common/infra/src/workspace/engine/doc/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+The typical workflow is:
+
+1. load data from storage, apply to `YDoc` instance.
+2. track `YDoc` changes
+3. write the changes back to storage.
+
+### SeqNum
+
+There is a `SeqNum` on each Doc data in `Storage`. Every time `LocalPart` writes data, `SeqNum` will be +1.
```
forehalo, line 89: why not using timestamps in local. 
a isomorphism storage layer are more easy to be maintained, think about the p2p sync we promised

packages/frontend/workspace-impl/src/local/doc-indexeddb.ts
darkskygit: image
the reproduction method was provided in the issue on aug. 29, and on sept. 18, the author stated that it had been fixed. I am not sure which commit fixed the problem. I will make some tests later. If it can be successfully reproduced, it will be reported in slack.

packages/common/infra/src/workspace/engine/doc/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+Represents the local storage used, Specific implementations are replaceable, such as `IndexedDBDocStorage` on the `browser` and `SqliteDocStorage` on the `desktop`.
+
+### DocEventBus
+
+Each `DocStorage` contains a `DocEventBus`, which is used to communicate with other engines that share the same storage.
+
+With `DocEventBus` we can sync updates between engines without connecting to the server.
+
+For example, on the `browser`, we have multiple tabs, all tabs share the same `IndexedDBDocStorage`, so we use `BroadcastChannel` to implement `DocEventBus`, which allows us to broadcast events to all tabs.
```
forehalo, line 32: will the two tabs get the same update at the same time from server if others pushed one?

packages/common/infra/src/workspace/engine/doc/README.md
forehalo: in my first design, all data providers are treated equally, this is why I chose naming them `main` and `followers`, or even just a providers list with priority instead of `local` and `remote`.

each provider comes with a storage, it will be easy understandable of you think of the storage as a "sidecar", and the storage should be standardized.

packages/common/infra/src/workspace/engine/doc/README.md
darkskygit: technically, we can allow receive updates from the server and `BroadcastChannel` at the same time, and they can be merged.

however, if we accept updates through `BroadcastChannel`, we better elect a tab to stay connected to the server, and all other tabs should be disconnected from the server and idb to avoid wasted bandwidth and multiple writes (idb)

darkskygit: after checking the code, currently no logic to disconnect ws when bc is connected, so updates will be received from the server twice and write to idb twice

packages/common/infra/src/workspace/engine/doc/README.md
EYHN: seqNum here is mainly to work with pushedSeqNum.

* Imagine you have two tabs writing updates to storage at the same time. `seqNum = 2`
* The two tabs are each responsible for sending their updates to the server.
* One tab is successfully sent, `pushedSeqNum += 1`, and the other tab fails to be sent, `pushedSeqNum` remains unchanged.
* At this time `pushedSeqNum < seqNum`, we can know that something is not uploaded to the server.

According to the current design, the local and server have been defined that they cannot be equal.

packages/common/infra/src/workspace/engine/doc/README.md
EYHN: All tabs have their own ws links, 

Based on the current design, we will not choose one tab as the leader. All tabs are equal and can be closed at any time.

packages/common/infra/src/workspace/engine/doc/README.md
EYHN: There are 2 modes of push in RemotePart,

One is partial push, which means only pushing the currently updated part. In this case `pushedSeqNum + 1`

The other is full push, which we do pull and then push, we can ensure that all updates are sent to the server.
In this case, you can safely set `pushedSeqNum = seqNum`

If you use timestamps, you cannot implement a partial push.

packages/common/infra/src/workspace/engine/doc/README.md
EYHN: If in the future we have a `service worker` we can just do the ws connection in the `service worker`.

darkskygit (approved):

Issue #6195, opened 2024-03-19
Add About shortcut for macOS Menubar.


qpomelo:
### Description

Expected behavior: click "about" menu item and open about page in settings.
<img width="574" alt="Screenshot 2024-03-19 at 3 08 07 PM" src="https://github.com/toeverything/AFFiNE/assets/35286331/a16c315d-7ca3-4e63-a9ba-81940755fcfa">

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

Aniruddha775:
hey can i work on this issue?

qpomelo:
Hi Aniruddha,


I hope this message finds you well. I wanted to take a moment to express
our gratitude for your interest in the AFFiNE project. It’s always exciting
to see new faces eager to contribute, and we’re thrilled to have you join
our community!


We believe that with your assistance, we can make AFFiNE even better. As
you get started, please take a look at our contribution guidelines and
project license to ensure a smooth collaboration. If you have any questions
or need further clarification on any points, feel free to reach out.


Thanks once again for being a part of this journey. We can’t wait to see
the great things we will achieve together!


Best regards,


QPomelo


Aniruddha Dhar Chowdhury ***@***.***>：

PR #6245, opened 2024-03-21
feat: open about page in setting modal when click about menu

BABA983:
For #6195


packages/frontend/core/src/components/root-app-sidebar/index.tsx
```diff
@@ -130,6 +132,18 @@ export const RootAppSidebar = ({
     return;
   }, [onClickNewPage]);
 
+  useEffect(() => {
```
pengx17: better move this side effect to https://github.com/BABA983/AFFiNE/blob/pr/6245/packages/frontend/core/src/providers/modal-provider.tsx#L94

BABA983: ✅

EYHN (approved):

PR #6251, opened 2024-03-21
feat(core): refactor sidebar header

CatsJuice:
- Add user avatar
- Move sign-out/user settings link from workspace-modal to user avatar modal
- Modify the style of workspace list items
- Modify gap of navigation buttons 
- Animate Syncing/Offline/...


CatsJuice:
* **#6251** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6251?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

JimmFly (approved):

JimmFly:
Do we require a larger clickable area?
For instance, transitioning from a red zone to a green zone.
image

CatsJuice:
> Do we require a larger clickable area? For instance, transitioning from a red zone to a green zone. image

Actually, I think so too, but this is how it's designed now. cc @zanwei

pengx17:
image
The clicking area is determined by workspace name length, which is weird.

pengx17:
This thing rotates on reload. Is this expected?

https://github.com/toeverything/AFFiNE/assets/584378/6683a176-7429-42f1-8184-4748196d21bf

pengx17:
Looks like we no longer requires emotion after this pr. Maybe we can finally remove it completely

CatsJuice:
> This thing rotates on reload. Is this expected?
> 
>  Kapture.2024-03-25.at.15.05.37.mp4

Oh, not expected. Local workspace shouldn't have sync status.

CatsJuice:
> Looks like we no longer requires emotion after this pr. Maybe we can finally remove it completely

I'm afraid there still be a lot of  `styled` used

packages/frontend/core/src/components/pure/workspace-slider-bar/workspace-card/index.tsx
```diff
@@ -196,21 +190,47 @@ const useSyncEngineSyncProgress = () => {
       ) : (
         <LocalWorkspaceStatus />
       ),
+    active: syncing || retrying || isOverCapacity,
```
EYHN: Maybe we can add a debounce here

EYHN: Here used to have a debounce, but I deleted it during refactoring.

As a result, the sync status always flashes when you edit the docs.

EYHN: I want the sync here not flicker, and the sync progress bar not lag (the old debounce would cause this).

Do you have any good ideas?

CatsJuice: I think we can try using css transition delay, so that the animation will be triggered only when the status changes for a certain duration.

CatsJuice: https://github.com/toeverything/AFFiNE/blob/729f3f42da50f6597a5485c646db94a16365391c/packages/frontend/core/src/components/pure/workspace-slider-bar/workspace-card/styles.css.ts#L30

After adding this delay, animation won't play unless it takes more than 0.3s to sync.

EYHN (approved):

zanwei:
Feedback received. We might need to change the click area to full width. Th

cc @CatsJuice 

image

zanwei:
One more thing needs to be change that adjust the portrait size from 20px to 24px, thanks!

cc @CatsJuice

PR #6255, opened 2024-03-21
chore: improve password error message

fundon:
chore: improve error message

chore: add password minlength & maxlength i18n

chore: check max length

fix: i18n variables

feat: add CredentialsRequirementType


fundon:
* **#6255** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6255?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

 Join @fundon and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/backend/server/src/core/config.ts
```diff
@@ -22,6 +22,21 @@ export function ADD_ENABLED_FEATURES(feature: ServerFeature) {
...
+export class PasswordLimitsType {
+  @Field()
+  minLength!: number;
+
+  @Field()
+  maxLength!: number;
+}
+
+@ObjectType()
+export class AuthPrivilegeType {
```
EYHN:
```suggestion
export class AuthPolicyType {
```

I think to express overall authorization rules or policies, choosing `AuthPolicy` might be more appropriate. Whereas if you need to specifically describe the permissions or privileges of a particular user or entity, then you should use `AuthPrivilege`

packages/frontend/component/src/components/auth-components/password-input/index.tsx
```diff
@@ -27,30 +30,37 @@ export const PasswordInput: FC<
...
+        case 2:
+          setStatus('medium');
+          break;
+        case 3:
+          setStatus('strong');
+          break;
+      }
+    },
+    [passwordLimits]
+  );
```
darkskygit: can we write a zod type validator to simplify this logic? its too long

packages/backend/server/src/core/config.ts
```diff
@@ -22,6 +22,21 @@ export function ADD_ENABLED_FEATURES(feature: ServerFeature) {
...
+export class PasswordLimitsType {
+  @Field()
+  minLength!: number;
+
+  @Field()
+  maxLength!: number;
+}
+
+@ObjectType()
+export class AuthPolicyType {
```
forehalo: `CredentialsRequirementType`, `Policy` term might be used in further Permission module

packages/frontend/component/src/components/auth-components/password-input/index.tsx
```diff
@@ -1,104 +1,200 @@
...
+        ctx.addIssue({
+          code: ZodIssueCode.custom,
+          message: t['com.affine.auth.set.password.message.maxlength']({
+            max: String(maxLength),
+          }),
+          params: {
+            status: 'maximum',
+          },
+        });
+      }
```
darkskygit: why not use min().max() directly? error message can process in view after, and we can use error code `too_small` and `too_big` to identify the error:
image

fundon: We want to pass the status parameter, so use `ZodCustomIssue`.

packages/frontend/component/src/components/auth-components/password-input/index.tsx
```diff
@@ -1,104 +1,200 @@
...
+          value: 'medium',
+          minDiversity: 4,
+          minLength: 8,
+        },
+        {
+          id: 2,
+          value: 'strong',
+          minDiversity: 4,
+          minLength: 10,
+        },
```
darkskygit: extract as constant

packages/frontend/component/src/components/auth-components/password-input/index.tsx
darkskygit: where we use the message? i see this i18n call already in the view

darkskygit: where we use the message? i see this i18n call already in the view

fundon: Good catch. 
Removed.

forehalo (approved):

EYHN (approved):

PR #6261, opened 2024-03-21
Add ENV for nodemailer, rejectUnauthorized

xiaoland:
Hi!😘
I deploy the AFFiNE on my own server and use my own mail server.
So it requires the nodemailer to allow invalid TLS certificate.

Based on https://nodemailer.com/smtp/#3-allow-self-signed-certificates, I made this change.

It really simple! What a great design ---- the ENV mapping!


xiaoland:
BTW, this shouldn't be the best solution for self-signed certificate.
I recommend to add the self-signed certificate into trusted.

Anyway, this is a solution too.

forehalo:
we don't recommend to add all config-env mappings to `affine.env.ts`, don't even talk about such a rare one.

users could access there `~/.affine/self-host/config/affine.js` on host machine to update the mailer config to allow untrusted certifications.

PR #6284, opened 2024-03-24
feat(core): auto focus the delete confirm button

### LINKED ISSUE #5813: Auto focus the delete confirm button
### Description

It would be better if the delete button in the confirmation modal is autoFocussed so that I can use the enter key instead of clicking the button to delete

### Use case

_No response_

### Anything else?

_No response_

ronitrajfr:
## What does this PR do?
This PR auto focus the delete confirm button

Fixes #5813 

##Type of change
Bug fix(non-breaking change which fixes the issue)


forehalo:
unfortunately, it doesn't work

PR #6285, opened 2024-03-24
refactor(core): favorite adapter

pengx17:
1. abstraction over favourites that supports different type of resources
2. sorting abstraction

### DIFF
--- a/packages/frontend/core/src/components/blocksuite/block-suite-header/favorite/index.tsx
+++ b/packages/frontend/core/src/components/blocksuite/block-suite-header/favorite/index.tsx
@@ -1,11 +1,10 @@
 import { FavoriteTag } from '@affine/core/components/page-list';
-import { useBlockSuiteMetaHelper } from '@affine/core/hooks/affine/use-block-suite-meta-helper';
-import { useBlockSuiteDocMeta } from '@affine/core/hooks/use-block-suite-page-meta';
+import { FavoriteItemsAdapter } from '@affine/core/modules/workspace';
 import { toast } from '@affine/core/utils';
 import { useAFFiNEI18N } from '@affine/i18n/hooks';
 import { assertExists } from '@blocksuite/global/utils';
-import { useService, Workspace } from '@toeverything/infra';
-import { useCallback } from 'react';
+import { useLiveData, useService, Workspace } from '@toeverything/infra';
+import { useCallback, useMemo } from 'react';
 
 export interface FavoriteButtonProps {
   pageId: string;
@@ -16,24 +15,21 @@ export const useFavorite = (pageId: string) => {
   const workspace = useService(Workspace);
   const docCollection = workspace.docCollection;
   const currentPage = docCollection.getDoc(pageId);
+  const favAdapter = useService(FavoriteItemsAdapter);
   assertExists(currentPage);
 
-  const pageMeta = useBlockSuiteDocMeta(docCollection).find(
-    meta => meta.id === pageId
+  const favorite = useLiveData(
+    useMemo(() => favAdapter.isFavorite$(pageId, 'doc'), [favAdapter, pageId])
   );
-  const favorite = pageMeta?.favorite ?? false;
-
-  const { toggleFavorite: _toggleFavorite } =
-    useBlockSuiteMetaHelper(docCollection);
 
   const toggleFavorite = useCallback(() => {
-    _toggleFavorite(pageId);
+    favAdapter.toggle(pageId, 'doc');
     toast(
       favorite
         ? t['com.affine.toastMessage.removedFavorites']()
         : t['com.affine.toastMessage.addedFavorites']()
     );
-  }, [favorite, pageId, t, _toggleFavorite]);
+  }, [favorite, pageId, t, favAdapter]);
 
   return { favorite, toggleFavorite };
 };
--- a/packages/frontend/core/src/components/blocksuite/block-suite-header/favorite/index.tsx
+++ b/packages/frontend/core/src/components/blocksuite/block-suite-header/favorite/index.tsx
@@ -4,7 +4,7 @@ import { toast } from '@affine/core/utils';
 import { useAFFiNEI18N } from '@affine/i18n/hooks';
 import { assertExists } from '@blocksuite/global/utils';
 import { useLiveData, useService, Workspace } from '@toeverything/infra';
-import { useCallback, useMemo } from 'react';
+import { useCallback } from 'react';
 
 export interface FavoriteButtonProps {
   pageId: string;
@@ -18,9 +18,7 @@ export const useFavorite = (pageId: string) => {
   const favAdapter = useService(FavoriteItemsAdapter);
   assertExists(currentPage);
 
-  const favorite = useLiveData(
-    useMemo(() => favAdapter.isFavorite$(pageId, 'doc'), [favAdapter, pageId])
-  );
+  const favorite = useLiveData(favAdapter.isFavorite$(pageId, 'doc'));
 
   const toggleFavorite = useCallback(() => {
     favAdapter.toggle(pageId, 'doc');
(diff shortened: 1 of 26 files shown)


pengx17:
* **#6302** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6302?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6288** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6288?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6285** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6285?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/utils/fractional-indexing.ts
```diff
@@ -0,0 +1,51 @@
+import { generateKeyBetween } from 'fractional-indexing';
+
+// Using fractional-indexing managing orders of items in a list
+export abstract class FractionalIndexingSortableItems<T> {
```
EYHN: I don't think this need to be an abstract class.

just providing some tool functions?

packages/frontend/core/src/modules/workspace/properties/adapter.ts
```diff
@@ -36,9 +38,18 @@ export class WorkspacePropertiesAdapter {
     const rootDoc = workspace.docCollection.doc;
     this.properties = rootDoc.getMap(AFFINE_PROPERTIES_ID);
     this.proxy = createYProxy(this.properties);
+
+    this.properties$ = LiveData.from(
+      new Observable(observer => {
```
EYHN, line 43: The observable in livedata should emit a value as soon as the listening starts.

packages/frontend/core/src/modules/workspace/properties/adapter.ts
```diff
@@ -36,9 +38,18 @@ export class WorkspacePropertiesAdapter {
...
+        this.properties.observeDeep(() => {
+          requestAnimationFrame(() => observer.next(this.proxy));
```
EYHN: why use requestAnimationFrame, 

Performance optimization to avoid rendering should be done in react

EYHN: This may cause issue if this.proxy is a reference that never changes.

packages/frontend/core/src/components/pure/workspace-slider-bar/favorite/favorite-list.tsx
```diff
@@ -15,11 +17,18 @@ export const FavoriteList = ({
   docCollection: workspace,
 }: FavoriteListProps) => {
   const metas = useBlockSuiteDocMeta(workspace);
+  const favAdapter = useService(FavoriteItemsAdapter);
   const dropItemId = getDropItemId('favorites');
 
+  const favourites = useLiveData(favAdapter.favorites$);
```
EYHN: You can derive one livedata from another using `map`.

This can prevent the filtered parts causing react to rerender.

```
favAdapter.favorites$.map((favourites) => {
    return favourites.filter(fav => {
      const meta = metas.find(m => m.id === fav.id);
      return meta && !meta.trash;
    })
})
```

packages/frontend/core/src/components/page-list/operation-cell.tsx
```diff
@@ -37,37 +44,63 @@ import type { AllPageListConfig } from './view';
...
   onRemoveFromAllowList,
 }: PageOperationCellProps) => {
   const t = useAFFiNEI18N();
+  const currentWorkspace = useService(Workspace);
   const { appSettings } = useAppSettingHelper();
+  const { setTrashModal } = useTrashModalHelper(currentWorkspace.docCollection);
   const [openDisableShared, setOpenDisableShared] = useState(false);
+  const favAdapter = useService(FavoriteItemsAdapter);
+  const favourite = useLiveData(
+    useMemo(() => favAdapter.isFavorite$(page.id, 'doc'), [favAdapter, page.id])
```
EYHN: should no need useMemo

packages/frontend/core/src/components/blocksuite/block-suite-header/favorite/index.tsx
```diff
@@ -16,24 +15,21 @@ export const useFavorite = (pageId: string) => {
   const workspace = useService(Workspace);
   const docCollection = workspace.docCollection;
   const currentPage = docCollection.getDoc(pageId);
+  const favAdapter = useService(FavoriteItemsAdapter);
   assertExists(currentPage);
 
-  const pageMeta = useBlockSuiteDocMeta(docCollection).find(
-    meta => meta.id === pageId
+  const favorite = useLiveData(
+    useMemo(() => favAdapter.isFavorite$(pageId, 'doc'), [favAdapter, pageId])
```
EYHN:
```suggestion
    favAdapter.isFavorite$(pageId, 'doc')
```

packages/frontend/core/src/components/page-list/operation-cell.tsx
pengx17: calling `favAdapter.isFavorite$` without memo will return a new observable on re-render, and then by combining with `useLiveData` there will be constant sub/unsubs under the hood. I think there will be some performance issues thus why I added use memo

packages/frontend/core/src/modules/workspace/properties/adapter.ts
pengx17: Let me check. I added `requestAnimationFrame` because there is some UI view inconsistent issues. It seems like to be related to proxy here

packages/frontend/core/src/components/pure/workspace-slider-bar/favorite/favorite-list.tsx
pengx17: I did not do this because of the same reason of performance concerns about sub/unsub on rerender.
What do you think?

packages/frontend/core/src/utils/fractional-indexing.ts
pengx17: changed to createFractionalIndexingSortableHelper

packages/frontend/core/src/modules/workspace/properties/adapter.ts
pengx17: It seems that observeDeep is called in the same tick when doc data is changed. The change is somehow ignored and will not trigger react rerender. however requestAnimationFrame will get rid of the issue. will take some more time digging out why.

packages/frontend/core/src/components/pure/workspace-slider-bar/favorite/favorite-list.tsx
EYHN: sub unsub should be fast. Much faster than react rerender

packages/frontend/core/src/components/page-list/operation-cell.tsx
EYHN: sub/unsub should be fast. This should not cause performance issues.

packages/frontend/core/src/modules/workspace/properties/adapter.ts
EYHN: I think you can keep it, I think we need to refactor this proxy in the future

pengx17: Oh I got it why it does not work - the proxy will firstly fire `observeDeep` and then set the value, which is the reason why in the callback it always gets the outdated data in the same tick: https://github.com/toeverything/blocksuite/blob/94ec11725132f96694c1ee83c3cfa64d697bb58d/packages/framework/store/src/reactive/proxy.ts#L157

EYHN (approved):

PR #6318, opened 2024-03-26
feat(core): add shortcut for multi select docs

JimmFly:
close TOV-701
In All Doc, `Shift + Click` has been added to activate the multiple selection state.

### DIFF
--- a/packages/frontend/core/src/modules/workbench/view/workbench-link.tsx
+++ b/packages/frontend/core/src/modules/workbench/view/workbench-link.tsx
@@ -20,6 +20,9 @@ export const WorkbenchLink = ({
     (event: React.MouseEvent<HTMLAnchorElement>) => {
       event.preventDefault();
       event.stopPropagation();
+      if (onClick?.(event)) {
+        return;
+      }
 
       if (event.ctrlKey || event.metaKey) {
         if (appSettings.enableMultiView && environment.isDesktop) {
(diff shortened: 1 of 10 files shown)


JimmFly:
* **#6318** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6318?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/modules/workbench/view/workbench-link.tsx
```diff
@@ -19,9 +19,13 @@ export const WorkbenchLink = ({
   const handleClick = useCallback(
     (event: React.MouseEvent<HTMLAnchorElement>) => {
       event.preventDefault();
-      event.stopPropagation();
```
pengx17, line 22: Why remove?

packages/frontend/core/src/modules/workbench/view/workbench-link.tsx
```diff
@@ -20,6 +20,9 @@ export const WorkbenchLink = ({
     (event: React.MouseEvent<HTMLAnchorElement>) => {
       event.preventDefault();
       event.stopPropagation();
+      if (event.shiftKey && onClick) {
```
pengx17: I think you may want to change `onClick` handler type to 
`(e: React.MouseEvent) => boolean`. If the return value is true, then we can skip all the following work. Otherwise we will invoke the detail behaviors.

BTW I think you can remove line 40

pengx17 (approved):

PR #6331, opened 2024-03-26
fix(core): editor pgup/pgdn issues

### LINKED ISSUE #6232: UI Bug when using PgUp or PgDown
### What happened?

When using PgUp or PgDown keys the UI crashes for a millisecond. Then a black block appears on the right side of the app. Everything else is pushed to the left and even if I try to minimize, then maximize, is no use. The only way to fix it is restarting the application or by clicking the right sidebar toggle button from the top toolbar. Is not that is blocking me from using the app, as the solution is simple, but I think it worth to mention as it doesn't feel right.

image

### Distribution version

Windows x64

_No response_

### Are you self-hosting?

- [ ] Yes

### Relevant log output

_No response_

### Anything else?

_No response_

pengx17:
fix https://github.com/toeverything/AFFiNE/issues/6232

### DIFF
--- a/packages/frontend/core/src/pages/workspace/detail-page/detail-page.tsx
+++ b/packages/frontend/core/src/pages/workspace/detail-page/detail-page.tsx
@@ -27,13 +27,14 @@ import {
 } from '@toeverything/infra';
 import clsx from 'clsx';
 import { useSetAtom } from 'jotai';
-import type { ReactElement } from 'react';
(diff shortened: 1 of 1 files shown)


pengx17:
* **#6331** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6331?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

EYHN (approved):

PR #6361, opened 2024-03-28
chore: remove unnecessary extension

Dunqing:
The main reason is that these `*.js` files don't exist


Dunqing:
Since removing the extension caused many problems, I decided to close it.

Also, I'm very curious why the `typecheck` command is `tsc -b tsconfig.json` instead of `tsc -noEmit` Would this be faster?

PR #6365, opened 2024-03-28
chore: update typecheck script

Dunqing:


Brooooooklyn:
@Dunqing `-b` is required, because without `-b` the tsc will do nothing since the `include` in `tsconfig.json` is `[]`

Dunqing:
> @Dunqing `-b` is required, because without `-b` the tsc will do nothing since the `include` in `tsconfig.json` is `[]`

I see. Thank you for telling me!

PR #6534, opened 2024-04-12
fix(electron): close app while on fullscreen should be minimize instead

pengx17:


pengx17:
* **#6534** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6534?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/electron/src/main/main-window.ts
```diff
@@ -127,7 +127,11 @@ async function createWindow(additionalArguments: string[]) {
       //   - all browser windows will capture the "close" event
       //   - the hidden window will close all windows
       //   - "window-all-closed" event will be emitted and eventually quit the app
-      browserWindow.hide();
+      if (browserWindow.isFullScreen()) {
+        browserWindow.setFullScreen(false);
```
EYHN, line 134: can we do both `setFullscreen(false)`  and `hide()`

pengx17: It seems setting full screen to false will trigger an animation. During animation transitioning period, `hide` will not work.

EYHN: https://github.com/electron/electron/issues/6033#issuecomment-242023295

seems to be a solution

[confused](https://app.graphite.dev/settings/meme-library?org=toeverything)

EYHN (approved):

PR #6561, opened 2024-04-15
feat: check quota correctly

darkskygit:


darkskygit:
* **#6561** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6561?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/backend/server/src/core/features/management.ts
```diff
@@ -116,9 +116,14 @@ export class FeatureManagementService {
...
+  // ======== User Feature ========
   async getUserFeatures(userId: string): Promise<FeatureType[]> {
-    return (await this.feature.getUserFeatures(userId)).map(
-      f => f.feature.name
-    );
+    const features = await this.feature.getUserFeatures(userId);
+    const types = features.filter(f => f.activated).map(f => f.feature.name);
+    // todo(@darkskygit): remove this after support cascade feature
+    if (types.includes(FeatureType.EarlyAccess)) {
+      types.push(FeatureType.Copilot);
```
EYHN: If we want to give them an AI subscription, I think this should be done by giving a stripe coupon instead of implementing it in the code.

EYHN (commented):
You use copilot feature here to represent that the user has unlimited copilot permissions. I have a few questions.

* I don’t seem to see you giving users `copilot` feature in user subscription and unsubscription.
* We seem to have used `copilot` feature to refer to the access rights of the workspace's AI Poc. Is there some confusion?
* Such changes seem to result in us having two quota systems ( `user.copilot.quota` and `user.quota` ). I think we can reuse the previous quota service.

can we implement the quota as "free_ai_plan" and "paied_ai_plan",

When querying quota, merge ai plan with cloud plan:

```
{
  ...paied_ai_plan,
  ...pro_plan
}
```

darkskygit:
> You use copilot feature here to represent that the user has unlimited copilot permissions. I have a few questions.
> 
> * I don’t seem to see you giving users  feature in user subscription and unsubscription.`copilot`
> * We seem to have used  feature to refer to the access rights of the workspace's AI Poc. Is there some confusion?`copilot`
> * Such changes seem to result in us having two quota systems (  and  ). I think we can reuse the previous quota service.`user.copilot.quota``user.quota`
> 
> can we implement the quota as "free_ai_plan" and "paied_ai_plan",
> 
> When querying quota, merge ai plan with cloud plan:
> 
> ```
> {
>   ...paied_ai_plan,
>   ...pro_plan
> }
> ```

The feature was originally designed to declare that "a user has a copilot feature", but it was temporarily used as an aipoc declaration.

Because there was no serverConfig api before, the frontend could not know whether the server module was enabled (for example, in self hosted, captcha will be disable), so the `availableFeature` was added as an auxiliary judgment. When the frontend can rely on the server and other module activation status send requests, this api should be removed

packages/backend/server/src/core/features/management.ts
darkskygit: hack code, has been remove and add ea user migration

darkskygit:
> > You use copilot feature here to represent that the user has unlimited copilot permissions. I have a few questions.
> > 
> > * I don’t seem to see you giving users  feature in user subscription and unsubscription.`copilot`
> > * We seem to have used  feature to refer to the access rights of the workspace's AI Poc. Is there some confusion?`copilot`
> > * Such changes seem to result in us having two quota systems (  and  ). I think we can reuse the previous quota service.`` user.copilot.quota``user.quota ``
> > 
> > can we implement the quota as "free_ai_plan" and "paied_ai_plan",
> > When querying quota, merge ai plan with cloud plan:
> > ```
> > {
> >   ...paied_ai_plan,
> >   ...pro_plan
> > }
> > ```
> 
> The feature was originally designed to declare that "a user has a copilot feature", but it was temporarily used as an aipoc declaration.
> 
> Because there was no serverConfig api before, the frontend could not know whether the server module was enabled (for example, in self hosted, captcha will be disable), so the `availableFeature` was added as an auxiliary judgment. When the frontend can rely on the server and other module activation status send requests, this api should be removed

@pengx17  does fontend never used `availableFeatures`, planning to remove it

EYHN:
According to the previous code, `copilot` feature is linked to workspace, not user

packages/backend/server/src/data/migrations/1713164714634-copilot-feature.ts
```diff
@@ -0,0 +1,85 @@
...
+    const newUsers = userIds.filter(u => !hasCopilotUserIds.includes(u));
+
+    await tx.userFeatures.createMany({
+      data: newUsers.map(userId => ({
+        userId,
+        featureId,
+        reason: 'Early access bonus',
+        activated: true,
+        // 1 year
+        expiredAt: new Date(Date.now() + 365 * 24 * 60 * 60 * 1000),
```
EYHN: I think based on our discussion, we should not manually code to implement this logic, we should give users a $0 coupon to let them get the AI function for free.

darkskygit:
> According to the previous code, `copilot` feature is linked to workspace, not user

This can indeed cause confusion, already define a new feature

packages/backend/server/src/data/migrations/1713164714634-copilot-feature.ts
darkskygit: just see the conclusion update in doc, will update it

EYHN:
One last thing, you should give the user the feature when the subscription is activated.

https://github.com/toeverything/AFFiNE/blob/canary/packages/backend/server/src/core/quota/service.ts#L163

darkskygit:
> One last thing, you should give the user the feature when the subscription is activated.
> 
> https://github.com/toeverything/AFFiNE/blob/canary/packages/backend/server/src/core/quota/service.ts#L163

fixed

packages/backend/server/src/core/features/service.ts
```diff
@@ -157,6 +166,38 @@ export class FeatureService {
...
   }
 
+  async getActivatedUserFeatures(userId: string) {
+    const features = await this.prisma.userFeatures.findMany({
+      where: {
+        user: { id: userId },
+        feature: {
+          type: FeatureKind.Feature,
+        },
+        expiredAt: {
```
forehalo: duplicated with `OR`

packages/backend/server/src/core/features/service.ts
```diff
@@ -338,4 +379,38 @@ export class FeatureService {
...
+  // ======== Event Handler ========
+
+  @OnEvent('user.subscription.activated')
+  async onSubscriptionUpdated({
+    userId,
+    plan,
+  }: EventPayload<'user.subscription.activated'>) {
+    switch (plan) {
+      case SubscriptionPlan.AI:
+        await this.addUserFeature(
```
forehalo: why not make it a quota

packages/backend/server/src/plugins/copilot/session.ts
```diff
@@ -347,6 +352,35 @@ export class ChatSessionService {
       );
   }
 
+  async getQuota(userId: string) {
+    const quota = await this.quota.getUserQuota(userId);
```
forehalo: only send quota query if user doesn't have unlimited feature, a db
query can be saved

packages/backend/server/src/plugins/copilot/session.ts
```diff
@@ -347,6 +352,35 @@ export class ChatSessionService {
...
+    const hasCopilotFeature = await this.feature
+      .getActivatedUserFeatures(userId)
+      .then(f => f.includes(FeatureType.UnlimitedCopilot));
+    const limit = hasCopilotFeature
+      ? undefined
+      : quota.feature.copilotActionLimit;
+
+    const actions = await this.countUserActions(userId);
+    const chats = await this.listHistories(userId).then(histories =>
```
forehalo: `listHistories` is too complex than counting. use `Prisma.count` directly

forehalo (approved):

PR #6621, opened 2024-04-18
fix(server): fix unstable test

EYHN:


EYHN:
* **#6621** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6621?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/backend/server/tests/nestjs/throttler.spec.ts
```diff
@@ -156,7 +156,7 @@ test('should use default throttler for unauthenticated user when not specified',
 
   t.is(headers.limit, '120');
   t.is(headers.remaining, '119');
-  t.is(headers.reset, '60');
+  t.regex(headers.reset, /59|60/);
```
darkskygit, line 159: extract all header assert as a function?
and we can convert to number and check in range to make test more stable

EYHN: This should always be an integer string. so this should be enough imo

darkskygit (approved):

PR #6637, opened 2024-04-19
fix: pass doc to chat panel

regischen:
Upstream https://github.com/toeverything/blocksuite/pull/6825

### DIFF
--- a/packages/frontend/core/src/modules/multi-tab-sidebar/multi-tabs/tabs/chat.tsx
+++ b/packages/frontend/core/src/modules/multi-tab-sidebar/multi-tabs/tabs/chat.tsx
@@ -1,7 +1,7 @@
 import { assertExists } from '@blocksuite/global/utils';
 import { AiIcon } from '@blocksuite/icons';
 import { ChatPanel } from '@blocksuite/presets';
-import { useCallback, useRef } from 'react';
+import { useCallback, useEffect, useRef } from 'react';
 
 import type { SidebarTab, SidebarTabProps } from '../sidebar-tab';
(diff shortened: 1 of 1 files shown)


packages/frontend/core/src/modules/multi-tab-sidebar/multi-tabs/tabs/chat.tsx
```diff
@@ -27,6 +34,7 @@ const EditorChatPanel = ({ editor }: SidebarTabProps) => {
 
   if (editor !== chatPanelRef.current?.editor) {
     (chatPanelRef.current as ChatPanel).editor = editor;
```
pengx17, line 36: can we infer doc from editor instance?

regischen: yes, but when doc switched, chat panel don't get informed.

pengx17: I guess you need to hook into doc change event somehow. There could be other cases that changes the active doc

regischen: okay， I'll change it later

pengx17: Since you already have an editor instance, you can add a event to it when the bound doc reference is changed. Let me know if this solution does not work

PR #6639, opened 2024-04-19
feat: update chat prompt

darkskygit:
fix AFF-878
fix AFF-852 (maybe)

### DIFF
--- a/packages/backend/server/src/data/migrations/utils/prompts.ts
+++ b/packages/backend/server/src/data/migrations/utils/prompts.ts
@@ -17,7 +17,24 @@ export const prompts: Prompt[] = [
   {
     name: 'debug:chat:gpt4',
     model: 'gpt-4-turbo-preview',
-    messages: [],
+    messages: [
+      {
+        role: 'system',
+        content:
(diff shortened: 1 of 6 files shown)


darkskygit:
* **#6639** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6639?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/backend/server/src/data/migrations/utils/prompts.ts
```diff
@@ -17,7 +17,24 @@ export const prompts: Prompt[] = [
...
+      },
+    ],
+  },
+  {
+    name: 'chat:gpt4',
+    model: 'gpt-4-vision-preview',
```
pengx17, line 30: Any text input action may contain an image blob, as we can execute actions on multiple blocks where one or more images are present.
Do you think it's possible to implement a logic in the server to intelligently select the most suitable model based on user inputs?

darkskygit: already planned in next stage

pengx17 (approved):

packages/backend/server/src/data/migrations/utils/prompts.ts
pengx17: Ok then. Can you please also make all gpt-4-turbo to vision instead in this PR?

darkskygit: vision may not perform as well as the original version in pure text dialogue and oneshot(action)

pengx17: If it is the case, we need to drop image blocks from inputs of these text actions @donteatfriedrice

PR #6659, opened 2024-04-22
feat: provide closeStream

regischen:
Upstream https://github.com/toeverything/blocksuite/pull/6836

### DIFF
--- a/packages/frontend/core/src/components/blocksuite/block-suite-editor/ai/request.ts
+++ b/packages/frontend/core/src/components/blocksuite/block-suite-editor/ai/request.ts
@@ -28,6 +28,7 @@ export type TextToTextOptions = {
   params?: Record<string, string>;
   timeout?: number;
   stream?: boolean;
+  signal?: AbortSignal;
 };
 
 export function createChatSession({
@@ -102,6 +103,7 @@ export function textToText({
   params,
   sessionId,
   stream,
+  signal,
   timeout = TIMEOUT,
 }: TextToTextOptions) {
   if (stream) {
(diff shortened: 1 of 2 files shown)


packages/frontend/core/src/components/blocksuite/block-suite-editor/ai/request.ts
```diff
@@ -120,12 +121,16 @@ export function textToText({
...
+        this.eventSource = eventSource;
         for await (const event of toTextStream(eventSource, { timeout })) {
           if (event.type === 'message') {
             yield event.data;
           }
         }
       },
+      closeStream: function () {
+        this.eventSource?.close();
+      },
```
pengx17: Exposing EventSource is not a good idea. Better use a `signal` from `AbortController` instead.

packages/frontend/core/src/components/blocksuite/block-suite-editor/ai/request.ts
```diff
@@ -120,6 +122,11 @@ export function textToText({
           sessionId: message.sessionId,
           messageId: message.messageId,
         });
+        if (signal) {
+          signal.onabort = () => {
```
lawvs, line 130: It is necessary to verify the abort status as well.

```ts
if (signal.aborted) return;
signal.onabort = () => {
   // ...
}
```

packages/frontend/core/src/components/blocksuite/block-suite-editor/ai/request.ts
```diff
@@ -120,6 +122,15 @@ export function textToText({
           sessionId: message.sessionId,
           messageId: message.messageId,
         });
+        if (signal) {
+          if (signal.aborted) {
```
pengx17, line 126: Shall be `!signal.aborted`

EYHN (approved):

PR #6738, opened 2024-04-29
fix(core): update prompts

pengx17:
Update prompts:

When sending a message to openai, we actually send two messages. 
There are two issues here:

1. the role in the first message shall be user or system
2. the content in the first message is not being interpolated at all.

```
[
  {
    role: "assistant",
    createdAt: "2024-04-27T06:30:34.820Z",
    params: {
      language: "French",
    },
    content: "You are a translation expert, please translate the following content into French, and only perform the translation action, keeping the translated content in the same format as the original content:\n\n          \"\"\"\"\n\n          \n\n          \"\"\"\"",
  },
  {
    role: "user",
    content: "Basic things you should know:",
    attachments: undefined,
    params: {
      language: "French",
    },
    createdAt: "2024-04-29T09:42:11.466Z",
  },
]
```

In this PR I removed `content` from prompts and let open ai use the second message (user) as the input.

### DIFF
--- a/packages/backend/server/src/data/migrations/1714386922280-update-prompts.ts
+++ b/packages/backend/server/src/data/migrations/1714386922280-update-prompts.ts
@@ -1,0 +1,13 @@
+import { PrismaClient } from '@prisma/client';
+
+import { refreshPrompts } from './utils/prompts';
+
+export class UpdatePrompts1714386922280 {
+  // do the migration
+  static async up(db: PrismaClient) {
+    await refreshPrompts(db);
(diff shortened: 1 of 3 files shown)


pengx17:
* **#6738** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6738?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

EYHN (approved):

PR #6930, opened 2024-05-14
Enabled auto focus on delete button

SambhavSetia:
fixes##5813


forehalo:
unfortunately, it doesn't work

PR #6947, opened 2024-05-15
feat: add retry support for copilot

darkskygit:
fix AFF-1066


darkskygit:
* **#6947** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6947?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/backend/server/src/plugins/copilot/controller.ts
```diff
@@ -129,7 +136,6 @@ export class CopilotController {
     @CurrentUser() user: CurrentUser,
     @Req() req: Request,
     @Param('sessionId') sessionId: string,
-    @Query('messageId') messageId: string,
```
pengx17, line 132: is this a breaking change? Maybe older client will not work any longer with it. Shall we make it compatible for old client at least for one major version?

darkskygit: message id will get in `@Query() params: Record<string, string | string[]>`, no behavier change

packages/backend/server/tests/copilot.e2e.ts
```diff
@@ -228,6 +228,61 @@ test('should be able to chat with api', async t => {
...
+      app,
+      token,
+      id,
+      randomUUID(),
+      promptName
+    );
+    const messageId = await createCopilotMessage(app, token, sessionId);
+    // chat 2 times
+    await chatWithText(app, token, sessionId, messageId);
+    await chatWithText(app, token, sessionId, messageId);
```
pengx17, line 249: In the real world case (like in chatgpt), the user may want to **edit** the last message and then resend. I think the current impl does not cover this case?

darkskygit: currently pr only cover the case about retry action or latest chat to get different response
edit message will impl in branching timeline like the chatgpt

pengx17 (approved):

PR #6415, opened 2024-03-30
fix(electron): optimize bundle size by removing unused dependencies

pengx17:
Should greatly reduce the size of helper.js and could speed up the time on starting the client app.

fix https://github.com/toeverything/AFFiNE/issues/6312

Before:

After:

PR #6969, opened 2024-05-16
fix(electron): remove all migration code in electron

pengx17:
The migration code import blocksuite in the dependency tree and prevent affine from running because electron's helper process tries to run browser only code that is part of the side effect of `@blocksuite/blocks`.

[The side effect free trick in esbuild config](https://github.com/toeverything/AFFiNE/pull/6415)  does not clean up these - not sure why. 

It has been already 6 month since we introduced the migration code in DB.
Instead of finding out the real root cause, I think may be better to remove the migration code completely so that no blocksuite code will be in the import paths in helper.js.


pengx17:
* **#6969** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6969?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

darkskygit (commented):
lgtm

darkskygit (approved):
lgtm

Brooooooklyn (approved):

PR #6972, opened 2024-05-17
refactor(core): replace history to ViewService.history

JimmFly:
upstream: #6966

### DIFF
--- a/packages/frontend/core/src/components/blocksuite/block-suite-editor/blocksuite-editor-container.tsx
+++ b/packages/frontend/core/src/components/blocksuite/block-suite-editor/blocksuite-editor-container.tsx
@@ -1,4 +1,5 @@
-import type { BlockElement } from '@blocksuite/block-std';
+import { ViewService } from '@affine/core/modules/workbench/services/view';
+import type { BaseSelection, BlockElement } from '@blocksuite/block-std';
 import type { Disposable } from '@blocksuite/global/utils';
 import type {
   AffineEditorContainer,
@@ -7,7 +8,7 @@ import type {
 } from '@blocksuite/presets';
 import type { Doc } from '@blocksuite/store';
 import { Slot } from '@blocksuite/store';
-import type { DocMode } from '@toeverything/infra';
+import { type DocMode, useServiceOptional } from '@toeverything/infra';
 import clsx from 'clsx';
 import type React from 'react';
 import type { RefObject } from 'react';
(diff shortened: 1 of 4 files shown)


JimmFly:
* **#6972** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/6972?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/components/blocksuite/block-suite-editor/blocksuite-editor-container.tsx
```diff
@@ -249,22 +234,71 @@ export const BlocksuiteEditorContainer = forwardRef<
...
+    blockElement,
+    affineEditorContainerProxy,
+    mode,
+    scrolled,
+    currentPath,
+    locationHash,
+  ]);
+
+  useEffect(() => {
+    if (currentPath === `/all` || (locationHash && !scrolled)) {
```
EYHN: When will currentPath be '/all'

JimmFly: It doesn’t seem to happen now. Before this, in the spilt view, the currentPath would occasionally become all, causing it to be replaced back to the all doc page after clicking on the page.

By the way, when you open a doc in the client, document.location.pathname is always all

EYHN (approved):

packages/frontend/core/src/components/blocksuite/block-suite-editor/blocksuite-editor-container.tsx
```diff
@@ -249,22 +234,71 @@ export const BlocksuiteEditorContainer = forwardRef<
...
-      disposable?.dispose();
+      group.dispose();
     };
-  }, [blockElement, affineEditorContainerProxy, mode, scrolled]);
+  }, [
+    affineEditorContainerProxy,
+    currentPath,
+    currentView,
+    locationHash,
+    scrolled,
```
pengx17: These properties look weird to me because changes to them will try to rebind the listener in affineEditorContainerProxy's updateComplete, that adding to disposable group is in async function but removing group is synced. I think there could be chances of dangling listeners if you write this way.

JimmFly: Is this modification appropriate?

JimmFly: I have attempted to make some revisions. Does this align with your thoughts?

PR #7061, opened 2024-05-24
fix(server): OIDC bug

IamTaoChen:
1. ZodError: parse response.json();

```bash
[Nest] 2163  - 05/24/2024, 10:45:06 PM     LOG [GraphQLModule] Mapped {/graphql, POST} route +160ms
file:///workspaces/AFFiNE/packages/backend/server/node_modules/zod/lib/index.mjs:587
                const error = new ZodError(ctx.common.issues);
                              ^

ZodError: [
  {
    "code": "invalid_type",
    "expected": "object",
    "received": "promise",
    "path": [],
    "message": "Expected object, received promise"
  }
]
    at get error [as error] (file:///workspaces/AFFiNE/packages/backend/server/node_modules/zod/lib/index.mjs:587:31)
    at ZodObject.parse (file:///workspaces/AFFiNE/packages/backend/server/node_modules/zod/lib/index.mjs:692:22)
    at OIDCClient.fetch (file:///workspaces/AFFiNE/packages/backend/server/src/plugins/oauth/providers/oidc.ts:52:25)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async OIDCClient.create (file:///workspaces/AFFiNE/packages/backend/server/src/plugins/oauth/providers/oidc.ts:59:28)
    at async OIDCProvider.onModuleInit (file:///workspaces/AFFiNE/packages/backend/server/src/plugins/oauth/providers/oidc.ts:148:27)
    at async Promise.all (index 2)
    at async callModuleInitHook (/workspaces/AFFiNE/packages/backend/server/node_modules/@nestjs/core/hooks/on-module-init.hook.js:43:5)
    at async NestApplication.callInitHook (/workspaces/AFFiNE/packages/backend/server/node_modules/@nestjs/core/nest-application-context.js:223:13)
    at async NestApplication.init (/workspaces/AFFiNE/packages/backend/server/node_modules/@nestjs/core/nest-application.js:100:9) {
  issues: [
    {
      code: 'invalid_type',
      expected: 'object',
      received: 'promise',
      path: [],
      message: 'Expected object, received promise'
    }
  ],
  addIssue: [Function (anonymous)],
  addIssues: [Function (anonymous)],
  errors: [
    {
      code: 'invalid_type',
      expected: 'object',
      received: 'promise',
      path: [],
      message: 'Expected object, received promise'
    }
  ]
}

Node.js v18.20.3
[nodemon] app crashed - waiting for file changes before starting...
```
2. Updated OIDCUserInfoSchema:
* Replaced the `id` field with the `sub` field because the OIDC user info does not contain an `id` field. Instead, it uses `sub` as the unique identifier.

According to the OpenID Connect (OIDC) specification, when requesting the openid scope, the Identity Provider (IdP) returns an ID Token and (optionally) a response from the UserInfo endpoint. The common fields in the OIDC user info include:

- `sub`: The unique identifier for the user at the identity provider (Subject Identifier). This is guaranteed to be returned under the OIDC specification and is used to uniquely identify the user.
- `preferred_username`: The user's preferred username.
- `email`: The user's email address.
- `name`: The user's full name.


EYHN (approved):

packages/backend/server/src/config/affine.env.ts
```diff
@@ -15,8 +15,13 @@ AFFiNE.ENV_MAP = {
   OAUTH_GOOGLE_CLIENT_SECRET: 'plugins.oauth.providers.google.clientSecret',
   OAUTH_GITHUB_CLIENT_ID: 'plugins.oauth.providers.github.clientId',
   OAUTH_GITHUB_CLIENT_SECRET: 'plugins.oauth.providers.github.clientSecret',
+  OAUTH_OIDC_ISSUER: 'plugins.oauth.providers.oidc.issuer',
   OAUTH_OIDC_CLIENT_ID: 'plugins.oauth.providers.oidc.clientId',
   OAUTH_OIDC_CLIENT_SECRET: 'plugins.oauth.providers.oidc.clientSecret',
+  OAUTH_OIDC_SCOPE: 'plugins.oauth.providers.oidc.args.scope',
+  OAUTH_OIDC_CLAIM_MAP_USERNAME: 'plugins.oauth.providers.oidc.args.claim_id',
+  OAUTH_OIDC_CLAIM_MAP_EMAIL: 'plugins.oauth.providers.oidc.args.claim_email',
+  OAUTH_OIDC_CLAIM_MAP_NAME: 'plugins.oauth.providers.oidc.args.claim_name',
```
darkskygit, line 24: using other fields as id/email/name seems to be a false requirement
when we use scope 'openid profile email' to request, the oidc provider should provide the above fields as return fields

packages/backend/server/src/config/affine.env.ts
```diff
@@ -15,8 +15,13 @@ AFFiNE.ENV_MAP = {
   OAUTH_GOOGLE_CLIENT_SECRET: 'plugins.oauth.providers.google.clientSecret',
   OAUTH_GITHUB_CLIENT_ID: 'plugins.oauth.providers.github.clientId',
   OAUTH_GITHUB_CLIENT_SECRET: 'plugins.oauth.providers.github.clientSecret',
+  OAUTH_OIDC_ISSUER: 'plugins.oauth.providers.oidc.issuer',
   OAUTH_OIDC_CLIENT_ID: 'plugins.oauth.providers.oidc.clientId',
   OAUTH_OIDC_CLIENT_SECRET: 'plugins.oauth.providers.oidc.clientSecret',
+  OAUTH_OIDC_SCOPE: 'plugins.oauth.providers.oidc.args.scope',
```
darkskygit, line 21: in what scenarios do we need to define different scope fields?
In affine login, we only need to get the username, email address and id as the unique identifier of the account

packages/backend/server/src/config/affine.env.ts
IamTaoChen: `id` is not the field of OIDC. so here we can set `OAUTH_OIDC_CLAIM_MAP_USERNAME=sub` or `OAUTH_OIDC_CLAIM_MAP_USERNAME=preferred_username`(default)

packages/backend/server/src/config/affine.env.ts
IamTaoChen: the default value, `openid profile email`, is OK. but I set this for the future, maybe we want more scope, e.g. `groups`.

packages/backend/server/src/config/affine.env.ts
IamTaoChen: According to the OpenID Connect (OIDC) specification, when requesting the openid scope, the Identity Provider (IdP) returns an ID Token and (optionally) a response from the UserInfo endpoint. The common fields in the OIDC user info include:

- `sub`: The unique identifier for the user at the identity provider (Subject Identifier). This is guaranteed to be returned under the OIDC specification and is used to uniquely identify the user.
- `preferred_username`: The user's preferred username.
- `email`: The user's email address.
- `name`: The user's full name.

darkskygit: what I mean is that there shouldn't be a situation where some oidc provider are missing necessary fields, so no additional mapping is needed
id/email/name can all be found in the claims field according to the standard definition: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

IamTaoChen: the `email` doesn't need a map. but `name` and `id` need, in my opinion.

e.g. 
### name
as I know, the `name` from keycloak is "first name"+ " " + "last name", most situation is ok. but you may now Chinese name is "last name"+ " " + "first name", so I define a new field as name for Chinese name, this is the reason I config this. but If we don't sync name from OIDC and we can change it in the website, yes, we can delete it.

### id
I'm thinking how we recognize the user, by email address(I think this is the current mode) or the `id`. 
* Here may be a problem, if I change my email address in my IDP, what happens? 
* But if we use `id` to distinguish users, if we don't change IDP, we can solve above problem.  

At the end, I agree that we can delete `MAP_EMAIL` and `MAP_NAME`. But for `USERNAME`(`id`) , I'm not sure.

darkskygit: > I'm thinking how we recognize the user, by email address(I think this is the current mode) or the id.

we dont use email to identify a unique user but use guid, so that can allow user change thire email

but what I am concerned about is whether we have such a scenario:

for a certain oidc provider, the id/email/name field is empty or does not exist, or the semantics of the corresponding field is incorrect, so we need to specify claim_id/email/name to declare that another field is used as id/email/name

If this does not happen, we can just use the fields in userinfo directly like this:

``` js
  async userinfo(token: string) {
    // fetch something
    const userinfo = ...

    return {
        id: userinfo.sub,
        email: userinfo.email,
        name: userinfo.preferred_username || userinfo.name
    };
  }
```

IamTaoChen: I believe all OIDC providers should offer the standard fields: `sub`, `email`, and `name`. However, we should include this but not expose it to users, in case some special OIDC providers do not provide these fields in the future. By the way, should we make the default ID `sub` or `preferred_username`?

darkskygit: i believe `sub` is the only choose of unique id, `preferred_username` is based on user's name, its not a stable identify of user account

IamTaoChen: But for selfhost, `preferred_username ` is more useful. 

Here is an example,

I use AD to save my users' information. and Keycloak as my IdP. But If I change my IdP to another one, the users' sub would be changed. but the `preferred_username` is from my AD and it would change.

PR #7072, opened 2024-05-26
feat: bootstrap support

darkskygit:
fix needing to repeatedly add upgrade prompt migrations
fix AFF-1075


darkskygit:
* **#7072** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7072?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

forehalo:
we will soon have an admin panel where we can manage prompts

darkskygit:
> we will soon have an admin panel where we can manage prompts

i think temporary and emergency prompt fixes can use the admin panel, however, all updates still need to be synchronized to the repository
because the content of the prompt is related to the frontend interaction, old versions of prompt may not support future versions of the front end

forehalo:
> > we will soon have an admin panel where we can manage prompts
> 
> i think temporary and emergency prompt fixes can use the admin panel, however, all updates still need to be synchronized to the repository
> because the content of the prompt is related to the frontend interaction, old versions of prompt may not support future versions of the front end

all prompts are read from db when used, so we will just need to sync the prompts to the latest in case of brand new deployments. Prompts CRUD in admin panel equals data migrations

darkskygit:
> > > we will soon have an admin panel where we can manage prompts
> > 
> > 
> > i think temporary and emergency prompt fixes can use the admin panel, however, all updates still need to be synchronized to the repository
> > because the content of the prompt is related to the frontend interaction, old versions of prompt may not support future versions of the front end
> 
> all prompts are read from db when used, so we will just need to sync the prompts to the latest in case of brand new deployments. Prompts CRUD in admin panel equals data migrations

lgtm

PR #7123, opened 2024-05-31
feat: add workflow resolver

darkskygit:
fix AFF-1166

PR #7119, opened 2024-05-31
feat: no branches workflow support

darkskygit:
fix AFF-1165 AFF-1164

PR #7159, opened 2024-06-06
feat: workflow executor

darkskygit:
fix AFF-1221 AFF-1232


darkskygit:
* **#7311** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7311?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7231** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7231?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7159** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7159?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#7123** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7123?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7119** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7119?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`

<h2></h2>

packages/backend/server/src/plugins/copilot/workflow/node.ts
```diff
@@ -0,0 +1,146 @@
...
+    return this.edges.length;
+  }
+
+  private async evaluateCondition(
+    condition: string,
+    params: WorkflowNodeState
+  ): Promise<string | undefined> {
+    try {
+      const chooseNode = this.worker.run({
+        iife: `(${condition})(nodeIds, params)`,
```
forehalo: why do we need to send condition to a worker to execute? the conditions are programmatically executable and predefined, all args are known at the mean time

darkskygit: we will allow users to modify prompts in the admin dashboard in future, then the workflow graph should also allow console modifications
in this case, the condition will not only be predefined, but may also be unknown risk inputs
so we need to use workers to isolate untrusted code

forehalo: It must be a long story allowing users modifications, and I prefer to use worker when it's really needed

darkskygit: switched to evaluate via new Function in unmodified state

packages/backend/server/src/plugins/copilot/workflow/node.ts
```diff
@@ -1,21 +1,70 @@
...
+    private readonly data: NodeData
+  ) {
+    if (data.nodeType === WorkflowNodeType.Basic) {
+      this.executor = getWorkflowExecutor(data.type);
+    } else if (data.nodeType === WorkflowNodeType.Decision) {
+      // prepare decision condition, reused in each run
+      const iife = `(${data.condition})(nodeIds, params)`;
+      // only eval the condition in worker if graph has been modified
+      if (graph.modified) {
+        const worker = new Piscina({
```
forehalo, line 37: still using worker?

darkskygit: only use worker when modified=true
currently, no workflow sets this flag, and it is reserved and will allow admin to modify it at runtime in the future

packages/backend/server/src/plugins/copilot/workflow/node.ts
```diff
@@ -1,21 +1,70 @@
...
+          execArgv: [],
+        });
+        this.condition = (params: WorkflowNodeState) =>
+          worker.run({
+            iife,
+            nodeIds: this.edges.map(node => node.id),
+            params,
+          });
+      } else {
+        const func = new Function(
```
forehalo: make data.condition type `fn | string`? that would be more safe and efficient when not modified cause I don't trust `fn.toString()` much

forehalo (approved):

PR #7166, opened 2024-06-06
feat: doc search infra

EYHN:
detail in `packages/common/infra/src/sync/indexer/README.md`


EYHN:
* **#7166** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7166?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/common/infra/src/sync/indexer/impl/indexeddb/__tests__/bm25.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { expect, test } from 'vitest';
+
+import { bm25 } from '../bm25';
+
+test('bm25', () => {
+  expect(bm25(1, 1, 10, 10, 15)).toMatchSnapshot();
```
forehalo: simple snapshot, make it inline?

packages/common/infra/src/sync/indexer/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+> **Array type**
+> All types can contain one or more values, so each field can store an array.
+
+2. Pick a backend
+
+Currently, there are two backends available.
+
+- `MemoryIndexer`: in-memory indexer, useful for testing.
+- `IndexedDBIndex`: persistent indexer using IndexedDB.
```
forehalo, line 32: no sqlite?

packages/common/infra/src/sync/indexer/impl/indexeddb/__tests__/tokenizer.spec.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import { SimpleTokenizer } from '../tokenizer';
+
+test('tokenizer', () => {
+  const tokens = new SimpleTokenizer().tokenize('hello  world,\n AFFiNE');
+
+  expect(tokens).toEqual([
+    { term: 'hello', start: 0, end: 5 },
+    { term: 'world', start: 7, end: 12 },
+    { term: 'affine', start: 15, end: 21 },
```
forehalo: lose the ability of case-sensitive search?

packages/common/infra/src/sync/indexer/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+Write data to the indexer. you need to start a write transaction by `await index.write()` first and then complete the batch write through `await writer.commit()`.
+
+> **Transactional**
+> Typically, the indexer does not provide transactional guarantees; reliable locking logic needs to be implemented at a higher level.
+
+```typescript
+const indexer = new IndexedDBIndex(schema);
+
+const writer = await index.write();
```
forehalo, line 50: is there a higher level wrapper to hide the detail of typed indexer?

packages/common/infra/src/sync/indexer/README.md
EYHN: Not yet, desktop also uses indexeddb

packages/common/infra/src/sync/indexer/impl/indexeddb/__tests__/tokenizer.spec.ts
EYHN: The tokenizer is used for full-text search. which generally needs to match as many results as possible, so is case insensitive.

packages/common/infra/src/sync/indexer/README.md
EYHN: No, is already a high-level abstraction

packages/common/infra/src/sync/indexer/README.md
```diff
@@ -0,0 +1,147 @@
...
+## Using
+
+1. Define schema
+
+First, we need to define the shape of the data. Currently, there are the following data types.
+
+- 'Integer'
+- 'Boolean'
+- 'FullText': for full-text search, it will be tokenized and stemmed.
+- 'String': for exact match search, e.g. tags, ids.
```
CatsJuice, line 14: ~~Maybe better to define a type for available type strings, and provide a method to define a schema(such as `defineSchema()`) to get better type hints and type safety?~~

Oh, I found it's already defined, but would it be better to define the Schema with generics?

CatsJuice: Maybe like:

```ts
type IndexType = 'Boolean' | 'String' | 'FullText' | 'Integer';
type IndexSchema<Fields extends string> = Record<Fields, IndexType>;

const defineSchema = <F extends string>(schema: IndexSchema<F>) =>
  schema;

const schema = defineSchema({
  title: 'FullText',
  tag: 'String',
  size: 'Integer',
});
```

packages/common/infra/src/sync/indexer/impl/indexeddb/match.ts
```diff
@@ -0,0 +1,126 @@
+export class Match {
+  scores = new Map<number, number>();
+  highlighters = new Map<
```
pengx17, line 6: it seems to me that highlighters are not relavent to `Match`. highlighters are for the consumers to render the result, right?

packages/common/infra/src/sync/indexer/impl/indexeddb/inverted-index.ts
```diff
@@ -0,0 +1,303 @@
...
+    }
+  }
+}
+
+export class FullTextInvertedIndex implements InvertedIndex {
+  constructor(readonly fieldKey: string) {}
+
+  async match(trx: DataStructROTransaction, term: string): Promise<Match> {
+    const queryTokens = new GeneralTokenizer().tokenize(term);
+    console.log(queryTokens);
```
pengx17: remove log

packages/common/infra/src/sync/indexer/__tests__/black-box.spec.ts
```diff
@@ -0,0 +1,415 @@
...
+  tag: 'String',
+  size: 'Integer',
+} as const;
+
+let index: Index<typeof schema> = null!;
+
+describe.each([
+  { name: 'memory', backend: MemoryIndex },
+  { name: 'idb', backend: IndexedDBIndex },
+])('impl tests($name)', ({ backend }) => {
```
pengx17: might give better naming like `indexer tests($name)`

packages/common/infra/src/sync/indexer/README.md
```diff
@@ -0,0 +1,147 @@
...
+        {
+            type: 'match',
+            field: 'title',
+            match: 'hello world',
+        },
+        {
+            type: 'match',
+            field: 'tag',
+            match: 'doc',
+        },
```
CatsJuice, line 87: Does the `FullText` type support both exact match and fuzzy search? 

It seems that the syntax for fuzzy search and exact match is exactly the same. Is it the case that for `FullText` it is tokenized search, and for `String` it is exact match?

packages/common/infra/src/sync/indexer/impl/indexeddb/__tests__/tokenizer.spec.ts
forehalo: what if user want to do case sensitive search

EYHN: I choose to give up this feature.
If you want to support both case-sensitive and case-insensitive simultaneously, you can only store two sets of indexes "AFFiNE" and "affine," which will require twice the storage space.

In a database, if you store the index for "AFFiNE," you won't be able to search for it using "affine."

In Elasticsearch, if case-sensitive search is needed, the indexing method must also be redefined, and it cannot switch to case-insensitive at runtime.

packages/common/infra/src/sync/indexer/impl/indexeddb/match.ts
EYHN: match is where the intermediate state of the query. If the name is not appropriate, do you have any suggestions?

EYHN: the highlighter here is ultimately ref to the highlight method in FullTextInvertedIndex. It may be more appropriate to replace it with the ranges array? `[field, start, end][]`

pengx17: > It may be more appropriate to replace it with the ranges array? [field, start, end][]

Agreed

packages/common/infra/src/sync/indexer/README.md
EYHN: fulltext always fuzzy search, string always exact search, and there is currently no way to define whether you need fuzzy or exact search.

They all use `match`, and the final behavior is defined by the internal implementation.

This is due to the limitations of the underlying implementation. For example, `elasticsearch` cannot choose which search method to use at runtime, and must decide before index.

packages/common/infra/src/sync/indexer/impl/indexeddb/tokenizer.ts
```diff
@@ -0,0 +1,162 @@
...
+  tokenize(text: string): Token[];
+}
+
+export interface Token {
+  term: string;
+  start: number;
+  end: number;
+}
+
+export class SimpleTokenizer implements Tokenizer {
```
pengx17, line 13: Intl.Segmenter might help with tokenization

EYHN: I'll try it later

pengx17 (approved):

packages/common/infra/src/sync/indexer/impl/indexeddb/index.ts
```diff
@@ -0,0 +1,124 @@
...
+    return (await this.getAll([id]))[0] ?? null;
+  }
+
+  async getAll(ids: string[]): Promise<Document<S>[]> {
+    const trx = await this.data.readonly();
+    return this.data.getAll(trx, ids);
+  }
+
+  async write(): Promise<IndexWriter<S>> {
+    return new IndexedDBIndexWriter(this.data, await this.data.readwrite());
```
darkskygit, line 43: add a option to create [relaxed durability](https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction/durability#relaxed) transaction? in next pr we will create write transaction frequently, default durability strategy will require browser frush fs every time when we make a new index for doc

EYHN: In my testing, write performance is not an issue

forehalo (commented):
I marked this pr draft to avoid miss merging it into canary.

Will focus on this pr after 0.15 released

packages/common/infra/src/sync/indexer/impl/indexeddb/tokenizer.ts
forehalo: add a todo

PR #7212, opened 2024-06-13
feat(infra): job system

EYHN:


EYHN:
* **#7212** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7212?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/common/infra/src/sync/job/impl/indexeddb/index.ts
```diff
@@ -0,0 +1,223 @@
...
+    return job.startTime !== null && job.startTime + this.TIMEOUT < Date.now();
+  }
+
+  private isAcceptable(job: JobRecord) {
+    return job.startTime === null || this.isTimeout(job);
+  }
+}
+
+function defer() {
+  const deferred = {} as {
```
pengx17: Can be rewritten using `Promise.withResolvers` instead

darkskygit (approved):

packages/common/infra/src/sync/job/impl/indexeddb/index.ts
```diff
@@ -0,0 +1,209 @@
...
+    if (jobs.length === 0) {
+      const batchKeys = trx.objectStore('jobs').index('batchKey').iterate();
+
+      let currentBatchKey: string = null as any;
+      let currentBatchJobs = [];
+      let skipCurrentBatch = false;
+
+      for await (const item of batchKeys) {
+        if (item.value.batchKey !== currentBatchKey) {
+          if (skipCurrentBatch === false && currentBatchJobs.length > 0) {
```
forehalo: `!skipCurrentBatch`, why does linter not complain about this

packages/common/infra/src/sync/job/impl/indexeddb/index.ts
```diff
@@ -0,0 +1,209 @@
...
+        }
+      }
+
+      if (skipCurrentBatch === false && currentBatchJobs.length > 0) {
+        jobs.push(...currentBatchJobs);
+      }
+    }
+
+    for (const { id, job } of jobs) {
+      const startTime = Date.now();
```
forehalo, line 93: start counting timeout immediately after job fetched is not a good idea, is not accurate for batched jobs

packages/common/infra/src/sync/job/impl/indexeddb/index.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    return jobs.map(({ id, job }) => ({
+      id: id.toString(),
+      batchKey: job.batchKey,
+      payload: job.payload,
+    }));
+  }
+
+  async waitForAccept(signal: AbortSignal): Promise<Job<J>[]> {
+    const broadcast = new BroadcastChannel(
```
forehalo, line 109: why not reuse the one in class instance?

packages/common/infra/src/sync/job/impl/indexeddb/index.ts
```diff
@@ -0,0 +1,209 @@
...
+      'idb-job-queue:' + this.databaseName
+    );
+
+    try {
+      let deferred = Promise.withResolvers<void>();
+
+      deferred.resolve();
```
forehalo: why resolved directly?

packages/common/infra/src/sync/job/runner.ts
```diff
@@ -0,0 +1,60 @@
...
+    private readonly queue: JobQueue<J>,
+    private readonly worker: (
+      jobs: Job<J>[],
+      signal: AbortSignal
+    ) => Promise<void>,
+    private readonly interval: () => Promise<void> = async () => {}
+  ) {}
+
+  start() {
+    this.stop();
```
forehalo, line 21: puzzling name, better to be something like `listenOnStopSignal`

packages/common/infra/src/sync/job/impl/indexeddb/index.ts
```diff
@@ -0,0 +1,209 @@
...
+  };
+}
+
+interface JobRecord {
+  batchKey: string;
+  startTime: number | null;
+  payload: any;
+}
+
+export class IndexedDBJobQueue<J> implements JobQueue<J> {
```
pengx17, line 23: how does it work in cases of multiple active browser tabs?

packages/common/infra/src/sync/job/README.md
```diff
@@ -0,0 +1,47 @@
+# job
+
+Job system abstraction for AFFiNE. Currently, only `IndexedDBJobQueue` is implemented; more backends will be implemented in the future.
+
+Run background jobs in browser & distributed environment. `runners` can consume tasks simultaneously without additional communication.
```
pengx17, line 5: Can you elaborate more on the use cases?

scripts/setup/polyfill.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable */
+// @ts-nocheck
+
+Promise.withResolvers ||
```
pengx17: may use `Promise.withResolvers ??= ` syntax instead

packages/common/infra/src/sync/job/impl/indexeddb/index.ts
```diff
@@ -0,0 +1,209 @@
...
+        batchKey: job.batchKey,
+        payload: job.payload,
+        startTime: null,
+      });
+    }
+
+    trx.commit();
+
+    // send broadcast to notify new jobs
+    this.broadcast.postMessage('new-jobs');
```
pengx17, line 47: I think maybe the indexer could be moved into service workers so that we can simulate a client-server pattern, so that:
1. we do not need to have to do tab synchronization with broadcast channels manually since transactions could be handled with message queues in the worker
2. it seems to me fuse is very resource consuming. Moveing out of the main thread might be a good idea.

EYHN: 1. I understand that if we can have a resident shared process in the browser, we can simplify the current process.
   
   However, at this stage, I don't want to introduce a service worker because it involves an independent lifecycle and version management, which would be very complex.


2. If it's just a regular worker instead of a service worker, our code supports doing so. If there are observable performance issues, I will do so.
   
   The `IndexedDBIndex` we currently use does not use fuse (that's `MemoryIndex`). The work of `IndexedDBIndex` is very light; based on my observation, the most intensive task is the `applyUpdate` operation when pulling data from doc storage. Executing this in a worker is indeed a good idea.

packages/common/infra/src/sync/job/impl/indexeddb/index.ts
EYHN: BroadcastChannel will not trigger message events for messages itself sends. This is to be able to receive events sent by itself.

packages/common/infra/src/sync/job/runner.ts
EYHN: which name?

packages/common/infra/src/sync/job/impl/indexeddb/index.ts
EYHN: What is the best way to do it?

packages/common/infra/src/sync/job/impl/indexeddb/index.ts
pengx17: > However, at this stage, I don't want to introduce a service worker because it involves an independent lifecycle and version management, which would be very complex.

This is debatable though. We could have a simple poc to evaluate the complexity on this solution.

> If there are observable performance issues, I will do so.

I suspect fuse.js should have similar performance compared to flexsearch that we are using right now in blocksuite. For the new impl the query is more advanced and might take even longer time for large dataset like our internal affine workspace.

EYHN: Are you referring to fuse.js, which is the method used in quicksearch to search for `commands` and `collections`?

We have no way to do that, because commands are dynamic and associated with the page state, we create a new fusejs instance every time.

packages/common/infra/src/sync/job/runner.ts
```diff
@@ -0,0 +1,60 @@
...
+    this.abort = null;
+  }
+
+  async loop(signal: AbortSignal) {
+    while (throwIfAborted(signal)) {
+      const jobs = await this.queue.waitForAccept(signal);
+
+      if (jobs !== null) {
+        try {
+          await this.worker(jobs, signal);
```
pengx17, line 42: may drop job with the same payload

EYHN: The worker can define such behaviors

PR #7213, opened 2024-06-13
feat(core): docs search service

EYHN:


EYHN:
* **#7213** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7213?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

CatsJuice (commented):
Is it possible to put the logic of the indexing in the worker?

EYHN:
> Is it possible to put the logic of the indexing in the worker?
> 
> 

Yes, I plan to do that later.

pengx17 (approved):

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
```diff
@@ -0,0 +1,299 @@
...
+import { blockIndexSchema, docIndexSchema } from '../schema';
+
+const logger = new DebugLogger('crawler');
+
+interface IndexerJobPayload {
+  docId: string;
+}
+
+export class DocsIndexer extends Entity {
+  private readonly jobQueue = new IndexedDBJobQueue<IndexerJobPayload>(
```
forehalo: this is exactly what I mean for higher abstraction for indexer, without refer to specific backend implementation

forehalo: if we have sqlite based indexer for electron, this would be hard to adapt

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  async crawlingDocData(docId: string) {
+    const rootDocBuffer =
+      await this.workspaceEngine.doc.storage.loadDocFromLocal(this.workspaceId);
+
+    if (!rootDocBuffer) {
+      return;
+    }
+
+    const yRootDoc = new YDoc();
```
forehalo, line 127: ydoc instance again?

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
```diff
@@ -0,0 +1,299 @@
...
+    applyUpdate(yRootDoc, rootDocBuffer);
+
+    const docStoragePossibleIds = Array.from(yRootDoc.getSubdocs())
+      .map(doc => doc.guid)
+      .filter(id => id.endsWith(docId));
+
+    let docBuffer;
+
+    for (const id of docStoragePossibleIds) {
```
forehalo, line 136: what's the meaning of this loop

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
```diff
@@ -0,0 +1,299 @@
+import { DebugLogger } from '@affine/debug';
```
forehalo, line 1: the calculating logic looks heavy. should we put it in worker, ?

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
EYHN: they have same interface `JobQueue`, If there is a replacement implementation, we can use DI to inject

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
EYHN: The doc id stored in doc storage may be in various formats, such as `space:{docid}`, `{workspaceid}:space:{docid}` , this loop list all possible id

EYHN: i forget break the loop

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
forehalo: then depends on interface instead of implementation

packages/common/infra/src/sync/doc/index.ts
```diff
@@ -23,6 +23,7 @@ export {
 } from './storage';
 
 export class DocEngine {
+  clientId: string;
```
Saul-Mirone: maybe readonly?

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
```diff
@@ -0,0 +1,299 @@
...
+
+      for (const block of blocks.values()) {
+        const flavour = block.get('sys:flavour')?.toString();
+        const blockId = block.get('sys:id')?.toString();
+
+        if (!flavour || !blockId) {
+          continue;
+        }
+
+        if (flavour === 'affine:page') {
```
Saul-Mirone, line 203: I think eventually every block can define the crawler for itself.

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
```diff
@@ -0,0 +1,299 @@
...
+            skip: 0,
+          },
+        }
+      )
+    ).nodes.map(n => n.id);
+
+    const needDelete = difference(allIndexedDocs, availableDocs);
+    const needAdd = difference(availableDocs, allIndexedDocs);
+
+    console.log(availableDocs, allIndexedDocs, needAdd, needDelete);
```
Saul-Mirone: Was this console.log left by design?

packages/frontend/core/src/modules/docs-search/services/docs-search.ts
```diff
@@ -0,0 +1,127 @@
...
+  WorkspaceEngineBeforeStart,
+} from '@toeverything/infra';
+
+import { DocsIndexer } from '../entities/docs-indexer';
+
+@OnEvent(WorkspaceEngineBeforeStart, s => s.handleWorkspaceEngineBeforeStart)
+export class DocsSearchService extends Service {
+  private readonly indexer = this.framework.createEntity(DocsIndexer);
+
+  constructor() {
```
Saul-Mirone: This constructor seems useless.

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
EYHN: Need to be define in blocksuite

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
darkskygit: prepare subdoc ids in `crawlingRootDocData` if we make full workspace indexing can save this root doc initial logic

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
```diff
@@ -0,0 +1,444 @@
...
+          batchKey: this.workspaceId,
+          payload: { docId: this.workspaceId },
+        },
+      ])
+      .catch(err => {
+        console.error('Error enqueueing job', err);
+      });
+  }
+
+  async crawlingDocData(docId: string) {
```
pengx17, line 119: This function impl is too complex. Could you please break it down?

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
```diff
@@ -0,0 +1,444 @@
...
+
+  private readonly workspaceId = this.workspaceService.workspace.id;
+
+  constructor(private readonly workspaceService: WorkspaceService) {
+    super();
+  }
+
+  setupListener() {
+    this.workspaceEngine.doc.storage.eventBus.on(event => {
+      if (event.clientId === this.workspaceEngine.doc.clientId) {
```
pengx17, line 69: how about remote updates?

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
```diff
@@ -0,0 +1,444 @@
...
+  async crawlingDocData(docId: string) {
+    const rootDocBuffer =
+      await this.workspaceEngine.doc.storage.loadDocFromLocal(this.workspaceId);
+
+    if (!rootDocBuffer) {
+      return;
+    }
+
+    const yRootDoc = new YDoc();
+    applyUpdate(yRootDoc, rootDocBuffer);
```
pengx17, line 128: applyUpdate is pretty slow. any ideas on improve it?

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
EYHN: Changed in subsequent PRs

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
EYHN: I'll do this when we have other implementations, but for now I've added a type here.

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
EYHN: This also includes the remote remote, where the clientid represents the client that receives the remote update. When there are multiple tabs, the tab that receives the update is responsible for pushing it into the job queue.

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
EYHN: I try to make each task as side-effect-free as possible.

EYHN: When we have a job queue and run it in a worker, I don't care too much about the efficiency of task execution.

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
EYHN: i have moved this to worker

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
EYHN: What your mean? We cannot get the data without creating an instance.

packages/frontend/core/src/modules/docs-search/entities/docs-indexer.ts
EYHN: One way to simplify this is implement the logic here by using blocksuite block model, which we will do later.

PR #7214, opened 2024-06-13
refactor(core): new quick search service

EYHN:


EYHN:
* **#7214** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7214?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#7267](https://github.com/toeverything/AFFiNE/pull/7267) <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7267?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7296](https://github.com/toeverything/AFFiNE/pull/7296) <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7296?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈
* `canary`

<h2></h2>

packages/frontend/core/src/modules/i18n/services/i18n.ts
```diff
@@ -0,0 +1,33 @@
+import { getI18n } from '@affine/i18n';
+import { Service } from '@toeverything/infra';
+
+import type { I18nFuncs, I18nString } from '../types/i18n';
+
+class RawI18nService extends Service {
```
pengx17: 1. shall we move out i18n refactor out of this pr?
2. it seems to me i18n does not need to be scoped into DI, since i18n is already a global singleton

packages/frontend/core/src/modules/quicksearch/services/cmdk.ts
```diff
@@ -0,0 +1,82 @@
...
+    super();
+  }
+
+  toggle() {
+    if (this.quickSearchService.quickSearch.show$.value) {
+      this.quickSearchService.quickSearch.hide();
+    } else {
+      this.quickSearchService.quickSearch.show(
+        [
+          this.framework.createEntity(RecentDocsQuickSearchSession),
```
pengx17, line 27: I think these sessions should be put as groups of config in quickSearch service, and using a strategy pattern with key identifiers to select the appropriate session group for `show`.

packages/frontend/core/src/modules/quicksearch/impls/commands.ts
```diff
@@ -0,0 +1,210 @@
...
+    const docMode =
+      get(this.contextService.globalContext.docMode.$) ?? undefined;
+    const commands = getAllCommand({ docMode });
+
+    const fuse = new Fuse(commands, {
+      keys: [{ name: 'label.title', weight: 2 }, 'label.subTitle'],
+      includeMatches: true,
+      includeScore: true,
+      threshold: 0.4,
+    });
```
pengx17, line 155: Is it better to wrap Fuse with a proper abstraction instead of accessing it directly?

pengx17: As a benefit, we can easily switch the underlying indexer engine and evaluate/compare the performance among them.

packages/frontend/core/src/modules/quicksearch/services/cmdk.ts
EYHN: The reason session provided by the caller is that the source of these sessions is very flexible.

For example, in insert link search, any third-party link can be matched. For this kind of non-reusable and scenario-specific, I can use a temporary function to implement it.

https://github.com/toeverything/AFFiNE/pull/7214/files/2c0c6bc958d7885dc4bd6d6854911da3cd10ab38#diff-54ad81614a441711f595238131eb727b2c0e7bd0aa8ecfb4c83c121ba2305493R376

Implementing it in QuickSearchService may make it unclear due to lack of context

packages/frontend/core/src/modules/i18n/services/i18n.ts
EYHN: I want to use i18n outside of react, and I wrote this service, I think it would be ok to use the global instance.

I can write some utility functions in @affine/i18n

pengx17: yeah I think the react part of i18n is already a wrapper around the i18n global instance. It is indeed useful to be used outside of react, including to be used somehow in lit

packages/frontend/core/src/modules/quicksearch/impls/docs.ts
```diff
@@ -0,0 +1,158 @@
...
+
+        return result;
+      }).pipe(
+        mergeMap((items: QuickSearchItem<'docs', DocsPayload>[]) => {
+          this.items$.next(items);
+          return EMPTY;
+        }),
+        onStart(() => {
+          this.items$.next([]);
+          // loading
```
pengx17: it seems isLoading$ should be used here

packages/frontend/core/src/modules/quicksearch/impls/docs.ts
```diff
@@ -0,0 +1,158 @@
...
+                timestamp: docRecord.meta$.value.updatedDate,
+                payload: {
+                  docId: maybeLink.docId,
+                },
+              },
+            ];
+          }
+        }
+
+        const docs = await this.docSearchService.search(query);
```
pengx17: do you think if it is a good idea to emit the link result immediately in line 73?

packages/frontend/core/src/modules/quicksearch/impls/commands.ts
EYHN: This is only used to search for command and collection. Different search algorithms should be implemented according to different data. I will keep using fusejs here.

forehalo (approved):

PR #7241, opened 2024-06-18
feat(component): helper function observeResize to observe size change via global ResizeObserver

CatsJuice:
```ts
import { observeResize } from "@affine/component";


useEffect(() => {
  const dispose = observeResize(element entry => {
    console.log(entry.contentRect);
  });

  return () => dispose();
}, []);
```

### DIFF
--- a/packages/frontend/component/src/index.ts
+++ b/packages/frontend/component/src/index.ts
@@ -24,3 +24,4 @@ export * from './ui/switch';
 export * from './ui/table';
 export * from './ui/toast';
 export * from './ui/tooltip';
+export * from './utils';
--- a/packages/frontend/component/src/ui/date-picker/week-date-picker.tsx
+++ b/packages/frontend/component/src/ui/date-picker/week-date-picker.tsx
@@ -13,6 +13,7 @@ import {
   useState,
 } from 'react';
 
+import { observeResize } from '../../utils';
 import { IconButton } from '../button';
 import * as styles from './week-date-picker.css';
 
@@ -116,8 +117,7 @@ export const WeekDatePicker = memo(function WeekDatePicker({
     const el = weekRef.current;
     if (!el) return;
 
-    const resizeObserver = new ResizeObserver(entries => {
-      const rect = entries[0].contentRect;
+    return observeResize(el, ({ contentRect: rect }) => {
       const width = rect.width;
       if (!width) return;
 
@@ -127,11 +127,6 @@ export const WeekDatePicker = memo(function WeekDatePicker({
       setViewPortSize(Math.max(1, Math.min(viewPortCount, 7)));
       setDense(width < 300);
     });
-    resizeObserver.observe(el);
-
-    return () => {
-      resizeObserver.disconnect();
-    };
   }, []);
 
   // when value changes, reset cursor
--- a/packages/frontend/component/src/utils/index.ts
+++ b/packages/frontend/component/src/utils/index.ts
@@ -1,0 +1,1 @@
+export * from './observe-resize';
--- a/packages/frontend/component/src/utils/observe-resize.ts
+++ b/packages/frontend/component/src/utils/observe-resize.ts
@@ -1,0 +1,77 @@
+import type { ResizeObserverEntry } from '@juggle/resize-observer';
+
+type ObserveResize = {
+  callback: (entity: ResizeObserverEntry) => void;
+  dispose: () => void;
+};
+
+let _resizeObserver: ResizeObserver | null = null;
+const elementsMap = new WeakMap<Element, Array<ObserveResize>>();
+
+// for debugging
+(window as any)._resizeObserverElementsMap = elementsMap;
+
+/**
+ * @internal get or initialize the ResizeObserver instance
+ */
+const getResizeObserver = () =>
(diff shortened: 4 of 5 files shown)


CatsJuice:
* **#7263** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7263?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7262** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7262?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7241** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7241?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/component/src/hooks/use-resize-observer.ts
```diff
@@ -0,0 +1,75 @@
...
+const elementsMap = new WeakMap<Element, UseResizeObserverCallback>();
+
+// for debugging
+(window as any)._resizeObserverElementsMap = elementsMap;
+
+/**
+ * @internal get or initialize the ResizeObserver instance
+ */
+const getResizeObserver = () =>
+  _resizeObserver ??
```
pengx17: may use the new `??=` syntax here

pengx17 (commented):
Thanks! TIL it is better to use single instance of ResizeObserver
https://github.com/WICG/resize-observer/issues/59

packages/frontend/component/src/hooks/use-resize-observer.ts
```diff
@@ -0,0 +1,75 @@
...
+  const prevElRef = useRef<Element | null>(null);
+
+  const observe = useCallback(
+    (element: Element, callback: UseResizeObserverCallback) => {
+      unobserve(prevElRef.current);
+
+      const observer = getResizeObserver();
+      prevElRef.current = element;
+      observer.observe(element);
+      elementsMap.set(element, callback);
```
pengx17: Is it possible to have multiple callbacks for the same element?

packages/frontend/component/src/hooks/use-resize-observer.ts
```diff
@@ -0,0 +1,74 @@
...
+
+      return () => unobserve(element);
+    },
+    []
+  );
+
+  useEffect(() => {
+    return () => {
+      unobserve(prevElRef.current);
+    };
```
pengx17: If the same element being observed twice, unobserve at one place will unobserve all.

packages/frontend/component/src/hooks/use-resize-observer.ts
CatsJuice: Yes, it is not currently supported. I think it's not a common scenario, or it could be handled externally. What do you think? @pengx17  @EYHN

packages/frontend/component/src/hooks/use-resize-observer.ts
CatsJuice: If same element being observed for the second time, the first should be unobserved first. All of this is based on each element having only one handler.

packages/frontend/component/src/hooks/use-resize-observer.ts
```diff
@@ -0,0 +1,74 @@
...
+(window as any)._resizeObserverElementsMap = elementsMap;
+
+/**
+ * @internal get or initialize the ResizeObserver instance
+ */
+const getResizeObserver = () =>
+  (_resizeObserver ??= new ResizeObserver(entries => {
+    entries.forEach(entry => {
+      const resizeCallback = elementsMap.get(entry.target);
+      resizeCallback?.(entry);
```
pengx17: I think we need to filter out unrelated target here

packages/frontend/component/src/hooks/use-resize-observer.ts
pengx17: It is technically possible, also `observe` can be called in the same component multiple times on different element and this misusage could hard to be avoided:

```tsx
const Foo = () => {
  const observe = useResizeObserver();
  observe(ref0.current);
  observe(ref1.current);
  return <div0  ref={ref0}/><div1 ref={ref1}/>
}
```

pengx17: I guess the reason of why you brought in `prevElRef` & `elementsMap` is to make sure element are unobserved in the end, but I think it we can make it the user's reponsibility to do cleanup. Then we do not need to worry about the issues I pointed out here.

```
const observe = useCallback(
  (element: Element, callback: UseResizeObserverCallback) => {
    const observer = getResizeObserver();
    observer.observe(element);
    return () => {
      observer.unobserve(element);
    }
  },
  []
);

```

CatsJuice: Got it, and if so without cleanup, a normal function rather than hook is suitable here.

CatsJuice: Extracted into a regular method, and now supports binding multiple listeners to the same element. Please take another look when you have time.

pengx17 (approved):

EYHN (approved):

PR #7251, opened 2024-06-18
refactor(i18n): i18n utils tools

EYHN:


EYHN:
* **#7251** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7251?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

JimmFly (approved):

packages/frontend/i18n/src/utils/time.ts
```diff
@@ -0,0 +1,281 @@
...
+
+  if (relativeOption) {
+    // show relative
+
+    const formatter = new Intl.RelativeTimeFormat(i18n.language, {
+      style: 'narrow',
+      numeric: relativeOption.yesterdayAndTomorrow ? 'auto' : 'always',
+    });
+
+    const timeUnitProcessor = {
```
pengx17, line 113: I think dayjs already provides relative time with locale support https://day.js.org/docs/en/plugin/relative-time
Do we need to do it on our own?

packages/frontend/i18n/src/i18n.ts
```diff
@@ -0,0 +1,54 @@
...
+    } else {
+      return (self as any)[key as string] as any;
+    }
+  },
+});
+
+/**
+ * I18n['com.affine.xxx']({ arg1: 'hello' }) -> '中文 hello'
+ */
+export const I18n = I18nProxy as I18nFuncs & typeof I18nMethod;
```
pengx17, line 53: I am wondering function if localized strings with I18n will get updated when changing language in React

packages/frontend/i18n/src/utils/time.ts
EYHN: dayjs needs to load additional language files to implement i18n, using browser intl api can reduce the bundle size

packages/frontend/i18n/src/i18n.ts
EYHN: The i18n here itself will change the language as react does, but the places where the i18n is used need to be re-executed.

pengx17: Then I guess using react hooks or <Trans /> component is still the recommended way of doing i18n in affine code, at least in react component?

EYHN: Yes

pengx17 (approved):

PR #7313, opened 2024-06-24
refactor: image preview component issues in center peek

pengx17:
fix AF-948

- remove jotai
- every editor got its own image block modal
- image block uses center peek modal for animation and style


pengx17:
* **#7313** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7313?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/components/image-preview/index.tsx
```diff
@@ -16,54 +17,131 @@ import {
...
-  previewblocksAtom,
-} from './index.jotai';
+import * as styles from './index.css';
 
 const filterImageBlock = (block: BlockModel): block is ImageBlockModel => {
   return block.flavour === 'affine:image';
 };
 
+function resolveMimeType(buffer: Uint8Array): string {
+  if (
```
forehalo, line 44: any better solution?

packages/frontend/core/src/components/image-preview/index.tsx
```diff
@@ -16,54 +17,131 @@ import {
...
+    return 'image/jpeg';
+  } else {
+    // unknown, fallback to png
+    console.error('unknown image type');
+    return 'image/png';
+  }
+}
+
+function copyImageToClipboard(image: HTMLImageElement) {
+  const canvas = document.createElement('canvas');
```
forehalo: need to set dpr or the image will blur

packages/frontend/core/src/components/image-preview/index.tsx
```diff
@@ -16,54 +17,131 @@ import {
...
+  disabled,
+  ...props
+}: PropsWithChildren<{
+  disabled?: boolean;
+  icon?: ReactElement;
+  tooltip: string;
+  onClick: () => void;
+  className?: string;
+}>) => {
+  const element = icon ? (
```
forehalo, line 140: this is stupid, why icon only mode is not supported by Button component by default

packages/frontend/core/src/components/image-preview/index.tsx
```diff
@@ -16,54 +17,131 @@ import {
...
+      })
+      .finally(() => {
+        canvas.remove();
+      });
+  }, 'image/png');
+}
+
+function saveBufferToFile(buffer: Uint8Array, filename: string) {
+  const blob = new Blob([buffer], { type: resolveMimeType(buffer) });
+  const url = URL.createObjectURL(blob);
```
CatsJuice: Never revoked?

packages/frontend/core/src/components/image-preview/index.tsx
pengx17: if I recall correctly @EYHN did something similar somewhere in affine as well in the past. Also this function is brought from the existing code. I extracted for clearance.

pengx17 (commented):
Will refactor the implemation to center peek based.

packages/frontend/core/src/components/image-preview/index.tsx
pengx17: Right now `IconButton` & `Button` are two different implementations. `<Button />` has icon prop as well but it serves different purpose.
I wrap it here for ease of use only for image preview toolbar.

packages/frontend/core/src/components/image-preview/index.tsx
```diff
@@ -472,20 +451,31 @@ export const ImagePreviewErrorBoundary = (
...
+  const isOpen = show && !!blockId;
 
-  const handleKeyUp = useCallback(
-    (event: KeyboardEvent) => {
-      if (event.key === 'Escape') {
-        event.preventDefault();
-        event.stopPropagation();
-        if (isOpen) {
-          setIsOpen(false);
+  useEffect(() => {
```
forehalo: it would be better to use layout effect

packages/frontend/core/src/components/image-preview/index.tsx
```diff
@@ -533,49 +526,52 @@ export const ImagePreviewModal = (
...
-    document.body
+          ) : null}
+          <button
+            data-testid="image-preview-close-button"
+            onClick={() => {
+              setShow(false);
+            }}
+            className={styles.imagePreviewModalCloseButtonStyle}
+          >
+            <svg
```
forehalo: extract as a icon component?

packages/frontend/core/src/components/image-preview/index.tsx
CatsJuice: Indeed, let's create an issue and deal with it later. [AF-960](https://linear.app/affine-design/issue/AF-960/deprecate-iconbutton-and-use-button-instead)

packages/frontend/core/src/modules/peek-view/view/doc-peek-view.tsx
```diff
@@ -154,6 +158,13 @@ export function DocPeekPreview({
               defaultSelectedBlockId={blockId}
               page={doc.blockSuiteDoc}
             />
+            {editor?.host ? (
+              <ImagePreviewModal
+                pageId={doc.id}
+                docCollection={doc.blockSuiteDoc.collection}
+                host={editor?.host}
```
forehalo: `?.` is not needed

packages/frontend/core/src/pages/workspace/detail-page/detail-page.tsx
```diff
@@ -299,7 +299,13 @@ const DetailPageImpl = memo(function DetailPageImpl() {
         }
       />
 
-      <ImagePreviewModal pageId={doc.id} docCollection={docCollection} />
+      {editor?.host ? (
+        <ImagePreviewModal
+          pageId={doc.id}
+          docCollection={docCollection}
+          host={editor?.host}
```
forehalo: `?.` is not needed

packages/frontend/core/src/components/image-preview/index.tsx
pengx17: changed to new impl

packages/frontend/core/src/components/image-preview/index.tsx
EYHN: maybe import `file-type`

pengx17: https://www.npmjs.com/package/@sgtpooki/file-type this one? It seems we already have it as a indirect dependency

EYHN:
> Will refactor the implemation to center peek based.

So this is still not the final version?

I think the issue is that this image preview still relies on us manually listening to events from the img tag, rather than the image block calling us

EYHN (approved):

pengx17:
> > Will refactor the implemation to center peek based.
> 
> So this is still not the final version?
> 
> I think the issue is that this image preview still relies on us manually listening to events from the img tag, rather than the image block calling us

Its not the final version but I will create a new pr to follow the peek view api for image preview.

packages/frontend/core/src/components/image-preview/index.tsx
forehalo: forget about it, it too big for doing so simple job

forehalo (approved):

packages/frontend/core/src/modules/peek-view/view/doc-peek-view.css.ts
```diff
@@ -7,3 +8,12 @@ export const editor = style({
   },
   minHeight: '100%',
 });
+
+export const affineDocViewport = style({
+  display: 'flex',
+  flexDirection: 'column',
+  userSelect: 'none',
```
L-Sun, line 15: What is the purpose of using `userSelect: 'none'`?

pengx17: sorry, I cannot recall why it is added

L-Sun: Never mind, I deleted this line so that the text can be selected in read-only mode

PR #7326, opened 2024-06-25
fix: use sgtpooki/file-type for resolving image type

pengx17:


pengx17:
* **#7326** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7326?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

forehalo (commented):
this library is ridiculous, can we really trust it?

pengx17:
> this library is ridiculous, can we really trust it?

There could be other image types like webp that we may also need to introduce soon. Besides blocksuite already has its as the dependency in https://github.com/toeverything/blocksuite/blob/45bdebd333aef568fd4c0e7bd5aebf9e6a8662f0/packages/blocks/src/attachment-block/utils.ts#L185.

forehalo:
@pengx17 I dont know the reason but why don't we use `file-type` directly? browser support is included after I checked

EYHN (approved):

EYHN:
> @pengx17 I dont know the reason but why don't we use `file-type` directly? browser support is included after I checked

More context:

https://github.com/toeverything/blocksuite/pull/6803
https://github.com/sindresorhus/file-type/issues/578

forehalo:
I prefer easily patching file-type with yarn by our own

EYHN:
> I prefer easily patching file-type with yarn by our own

I also tried patch, but the reference to `node:buffer` comes from his upstream package `token-types`

PR #7335, opened 2024-06-26
fix(core): fix error overflow

EYHN:


EYHN:
* **#7335** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7335?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/modules/workbench/view/route-container.css.ts
```diff
@@ -4,7 +4,7 @@ import { style } from '@vanilla-extract/css';
 export const root = style({
   display: 'flex',
   height: '100%',
-  overflow: 'hidden',
+  overflow: 'clip',
```
forehalo, line 7: what's the difference visually

EYHN: no difference visually, but hidden can still be scrolled by the program, while clip cannot be scrolled by the program.
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#clip

forehalo (approved):

PR #7337, opened 2024-06-26
chore(core): removed obsolete TODO

JimmFly:
Cleaned up some TODOs that no longer need to be done

### DIFF
--- a/packages/frontend/core/src/pages/workspace/detail-page/detail-page.css.ts
+++ b/packages/frontend/core/src/pages/workspace/detail-page/detail-page.css.ts
@@ -22,7 +22,6 @@ export const affineDocViewport = style({
   flexDirection: 'column',
   userSelect: 'none',
   containerName: 'viewport',
-  // TODO(@JimmFly): find out what this does in bs
   containerType: 'inline-size',
   background: cssVar('backgroundPrimaryColor'),
   '@media': {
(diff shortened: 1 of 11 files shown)


JimmFly:
* **#7337** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7337?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/pages/workspace/detail-page/detail-page.css.ts
```diff
@@ -22,7 +22,6 @@ export const affineDocViewport = style({
   flexDirection: 'column',
   userSelect: 'none',
   containerName: 'viewport',
-  // TODO(@JimmFly): find out what this does in bs
```
forehalo, line 25: so, found out?

CatsJuice (approved):

packages/frontend/core/src/pages/workspace/detail-page/detail-page.css.ts
JimmFly: This style itself is copied from [block-suite-editor/styles.css.ts](https://github.com/toeverything/AFFiNE/commit/f832ada98cf8fcf09339bd3a2c08cb95c95f5016#diff-30ec7d1597a28e385440da4a95099eb4ae0dc7a828c2b0bd246cfbb41f642872), and the style inside is completely different.
Regarding this class name, there are many places in blocksuite to query it. This is a required class name.

pengx17: It is to determine the main doc padding width on the left & right.

forehalo (approved):

PR #7339, opened 2024-06-26
refactor(core): remove the automatic URL change feature

JimmFly:
close [AF-954](https://linear.app/affine-design/issue/AF-954/center-peek-中-editor-selection-变化不应修改-location-hash)

Since the URL changes constantly with the selection, causing the editor to render frequently and also affecting the use of center peek, we decided to remove the function of automatically changing the URL.

### DIFF
--- a/packages/frontend/core/src/hooks/affine/use-share-url.ts
+++ b/packages/frontend/core/src/hooks/affine/use-share-url.ts
@@ -1,10 +1,11 @@
 import { notify } from '@affine/component';
 import { getAffineCloudBaseUrl } from '@affine/core/modules/cloud/services/fetch';
-import { WorkbenchService } from '@affine/core/modules/workbench';
 import { mixpanel } from '@affine/core/utils';
 import { useI18n } from '@affine/i18n';
-import { useLiveData, useService } from '@toeverything/infra';
-import { useCallback, useMemo } from 'react';
+import type { Disposable } from '@blocksuite/global/utils';
+import { useCallback, useEffect, useMemo, useState } from 'react';
+
+import { useActiveBlocksuiteEditor } from '../use-block-suite-editor';
 
 type UrlType = 'share' | 'workspace';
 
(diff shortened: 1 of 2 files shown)


JimmFly:
* **#7339** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7339?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

forehalo (commented):
any context?

packages/frontend/core/src/hooks/affine/use-share-url.ts
```diff
@@ -21,9 +22,8 @@ const useGenerateUrl = ({ workspaceId, pageId, urlType }: UseSharingUrl) => {
   // to generate a public url like https://app.affine.app/share/123/456
```
pengx17, line 27: <i class='graphite__hidden'>[Re: line 18]<br/><br/></i>

This hook is used for the sharing panel and it seems feature-wise we do not need to sync the blockid into hash there.

Some implementation issues:
Consdering hash is the same with blockId, which should also be part of the `UseSharingUrl` props. Also, the given active editor may not necessarily the same with the given pageId.
Thus the hook may be refactored to a plain function.<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7339?utm_source=unchanged-line-comment">Graphite</a>.</span>

JimmFly:
> Since the URL changes constantly with the selection, causing the editor to render frequently and also affecting the use of center peek, we decided to remove the function of automatically changing the URL.

JimmFly:
> any context?

Since the URL changes constantly with the selection, causing the editor to render frequently and also affecting the use of center peek, we decided to remove the function of automatically changing the URL. @forehalo

forehalo:
> > any context?
> 
> Since the URL changes constantly with the selection, causing the editor to render frequently and also affecting the use of center peek, we decided to remove the function of automatically changing the URL. @forehalo 

cool

packages/frontend/core/src/hooks/affine/use-share-url.ts
JimmFly: I made some changes according to your suggestions. What do you think about this?

pengx17 (approved):

packages/frontend/core/src/hooks/affine/use-share-url.ts
```diff
@@ -73,6 +79,30 @@ export const useSharingUrl = ({
...
+    if (currentBlockSelection) {
+      setBlockId(`#${currentBlockSelection.blockId}`);
+    }
+
+    disposable = selectManager.slots.changed.on(selections => {
+      if (selections[0] && selections[0].type === 'block') {
+        setBlockId(`#${selections[0].blockId}`);
+      } else if (blockId.length > 0) {
+        setBlockId('');
+      }
```
EYHN:
```suggestion
      setBlockId(
          (prev) => {
                if (selections[0] && selections[0].type === 'block') {
                  return `#${selections[0].blockId}`;
                 } else if (prev.length > 0) {
                  return '';
                 } else {
                   return prev;
                 }
          }
      )
```

packages/frontend/core/src/hooks/affine/use-share-url.ts
```diff
@@ -73,6 +79,30 @@ export const useSharingUrl = ({
...
+    });
+    return () => {
+      disposable?.dispose();
+    };
+  }, [blockId.length, editor?.host?.selection, urlType]);
```
EYHN:
```suggestion
  }, [editor?.host?.selection, urlType]);
```

EYHN (approved):

PR #7355, opened 2024-06-27
refactor(core): workbench

EYHN:
Merge the right sidebar logic into the workbench. this can simplify our logic.

Previously we had 3 modules

* workbench
* right-sidebar (Control sidebar open&close)
* multi-tab-sidebar (Control tabs)

Now everything is managed in Workbench.

# Behavioral changes

The sidebar button is always visible and can be opened at any time.
If there is no content to display,  will be `No Selection`

Elements in the sidebar can now be defined as`unmountOnInactive=false`. Inactive sidebars are marked with `display: none` but not unmount, so the `ChatPanel` can always remain in the DOM and user input will be retained even if the sidebar is closed.


EYHN:
* **#7355** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7355?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

pengx17 (commented):
can you update the description on the purpose of changes?

packages/frontend/core/src/modules/workbench/entities/view.ts
```diff
@@ -2,13 +2,25 @@ import { Entity, LiveData } from '@toeverything/infra';
...
 import { Observable } from 'rxjs';
 
-import { createIsland } from '../../../utils/island';
 import { createNavigableHistory } from '../../../utils/navigable-history';
 import type { ViewScope } from '../scopes/view';
 
 export class View extends Entity {
   id = this.scope.props.id;
 
+  sidebarTabs$ = new LiveData<string[]>([]);
```
CatsJuice: 🤔 Why not keep `Sidebar` entity but manage in view directly?

EYHN: I deleted the sidebar, and the sidebar is now managed by the view

CatsJuice: Yes, I meant that is it more intuitive to maintain a `Sidebar` entity for the logic of the sidebar, as  `view.sidebar`?

CatsJuice (approved):

EYHN:
> can you update the description on the purpose of changes?

Simplified the logic of the sidebar, which was a bit over-designed before.

And now that the sidebar is included in the workbench, their states can also be saved and persisted together.

PR #7365, opened 2024-06-27
fix: escape key handling compatibility issue with blocksuite

pengx17:
fix PD-1347

See
https://github.com/radix-ui/primitives/blob/main/packages/react/use-escape-keydown/src/useEscapeKeydown.tsx#L19-L20

This behavior in radix-ui is not possible to be prevented since escape keydown is bound to document with capturing, unless dialogs in blocksuite is also implemented with radix-ui.


pengx17:
* **#7365** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7365?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

CatsJuice (approved):

PR #7374, opened 2024-06-28
fix: click continue with ai not work

regischen:


regischen:
* **#7374** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7374?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

 Join @regischen and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

EYHN (approved):

CatsJuice (approved):

pengx17 (commented):
The bug should be caused by temporaryParams here.
We'd better not let affine to directly access chat panels' internal value in run time. A proper solution might be exposed through the ChatPanel's interface.

regischen:
> The bug should be caused by temporaryParams here.
> We'd better not let affine to directly access chat panels' internal value in run time. A proper solution might be exposed through the ChatPanel's interface.
> 
> 

The bug is because the handler is in chat-cards component, but it is not rendered when there are histories.

fundon:
This is a workaround. 

There should be a better solution, can we use localstorage for temporary storage?

PR #7459, opened 2024-07-09
fix: switch to `file-type` v19.1.0

lawvs:
The official `file-type` package has replaced the usage of node:buffer with Uint8Array. This change allows it to run safely in the browser now.

Related to https://github.com/sindresorhus/file-type/issues/578

### DIFF
--- a/packages/frontend/core/package.json
+++ b/packages/frontend/core/package.json
@@ -57,6 +57,7 @@
     "cmdk": "^1.0.0",
     "css-spring": "^4.1.0",
     "dayjs": "^1.11.10",
+    "file-type": "^19.1.0",
     "foxact": "^0.2.33",
     "fractional-indexing": "^3.2.0",
     "fuse.js": "^7.0.0",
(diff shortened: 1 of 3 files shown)


pengx17 (commented):
❤️ Can you please fix lint issue?

EYHN (approved):

packages/frontend/core/package.json
```diff
@@ -57,6 +57,7 @@
     "cmdk": "^1.0.0",
```
EYHN, line 57: <i class='graphite__hidden'>[Re: line 50]<br/><br/></i>

remove this ?<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7459?utm_source=unchanged-line-comment">Graphite</a>.</span>

lawvs: I found the `cmdk` is used in the core package. Can it be safely removed?

https://github.com/toeverything/AFFiNE/blob/40e381e2723fdedc626f6c4c27c44d4eceeee496/packages/frontend/core/src/modules/quicksearch/views/cmdk.tsx#L4

forehalo: cleanup work should be done in separated pr

PR #7464, opened 2024-07-10
feat: adapt workflow for ppt & minimap

darkskygit:


packages/frontend/core/src/components/blocksuite/block-suite-editor/ai/setup-provider.tsx
```diff
@@ -289,10 +291,47 @@ Could you make a new website based on these notes and send back just the html fi
...
+        return null;
+      }
+    };
+    const postfix = (text: string): string => {
+      const slides = text
+        .split('\n')
+        .map(parseJson)
+        .filter((v): v is Slide => !!v);
+      return slides
+        .map(slide => {
```
pengx17, line 314: how about converting in backend?

pengx17 (approved):

packages/frontend/core/src/components/blocksuite/block-suite-editor/ai/setup-provider.tsx
darkskygit: backend's workflow not impl a custom code node to allow this behaviour, i mark a todo here, will impl it after i made the yocto binding

PR #7465, opened 2024-07-10
fix(core): values in edit not saved when closing info modal

JimmFly:
Before the modification, when the modal is closed, if a value is being edited, the input's on blur event cannot be triggered to automatically save.

https://github.com/toeverything/AFFiNE/assets/102217452/60235cec-0022-4c4d-b213-28f2331a0c5b

### DIFF
--- a/packages/frontend/core/src/components/affine/page-properties/info-modal/info-modal.tsx
+++ b/packages/frontend/core/src/components/affine/page-properties/info-modal/info-modal.tsx
@@ -12,7 +12,17 @@ import {
   useService,
   type Workspace,
 } from '@toeverything/infra';
-import { Suspense, useCallback, useContext, useMemo, useRef } from 'react';
+import { assignInlineVars } from '@vanilla-extract/dynamic';
+import clsx from 'clsx';
+import {
+  Suspense,
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+} from 'react';
+import { useTransition } from 'react-transition-state';
 
 import { BlocksuiteHeaderTitle } from '../../../blocksuite/block-suite-header/title';
 import { managerContext } from '../common';
@@ -27,6 +37,8 @@ import * as styles from './info-modal.css';
 import { TagsRow } from './tags-row';
 import { TimeRow } from './time-row';
 
+const animationTimeout = 120;
+
 export const InfoModal = ({
   open,
   onOpenChange,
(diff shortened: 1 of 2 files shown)


JimmFly:
* **#7474** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7474?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7465** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7465?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/components/affine/page-properties/info-modal/info-modal.tsx
```diff
@@ -63,14 +75,19 @@ export const InfoModal = ({
...
-      onOpenChange={onOpenChange}
+      onOpenChange={handleOpenChange}
       withoutCloseButton
     >
       <Scrollable.Root>
         <Scrollable.Viewport
           className={styles.viewport}
           data-testid="info-modal"
         >
+          <input
```
pengx17: sorry I don't understand this fix. Can you please explain?

JimmFly: Before the modification, when the modal is closed, if a value is being edited, the input's on blur event cannot be triggered to automatically save.
The way I handle this issue is to create a hidden input and automatically focus it before closing the modal, thus triggering on blur event for the other input.

packages/frontend/core/src/components/affine/page-properties/info-modal/info-modal.tsx
```diff
@@ -39,7 +39,19 @@ export const InfoModal = ({
   workspace: Workspace;
 }) => {
   const titleInputHandleRef = useRef<InlineEditHandle>(null);
+  const hiddenInputRef = useRef<HTMLInputElement>(null);
+
   const manager = usePagePropertiesManager(page);
+
+  const handleOpenChange = useCallback(
+    (open: boolean) => {
+      if (!open) {
```
CatsJuice: Maybe just execute `(document.activeElement as HTMLElement)?.blur()` here?

packages/frontend/core/src/components/affine/page-properties/info-modal/info-modal.tsx
```diff
@@ -63,14 +75,19 @@ export const InfoModal = ({
...
     >
       <Scrollable.Root>
         <Scrollable.Viewport
           className={styles.viewport}
           data-testid="info-modal"
         >
+          <input
+            type="text"
+            ref={hiddenInputRef}
+            className={styles.hiddenInput}
```
CatsJuice: [`hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden)

packages/frontend/core/src/components/affine/page-properties/info-modal/info-modal.tsx
pengx17: I see. Not sure if this fix is too hacky. Maybe adding an exiting animation could also work?

CatsJuice: If animation is necessary, I think it's better to implement in global `<Modal />` and make it configurable.

EYHN (approved):

PR #7466, opened 2024-07-10
fix(core): can not preview image in shared page

L-Sun:
FIx BS-798

### DIFF
--- a/packages/frontend/core/src/components/blocksuite/block-suite-editor/lit-adaper.tsx
+++ b/packages/frontend/core/src/components/blocksuite/block-suite-editor/lit-adaper.tsx
@@ -97,9 +97,7 @@ const usePatchSpecs = (page: Doc, shared: boolean, mode: DocMode) => {
       patchReferenceRenderer(patched, reactToLit, referenceRenderer),
       confirmModal
     );
-    if (!page.readonly) {
-      patched = patchPeekViewService(patched, peekViewService);
-    }
+    patched = patchPeekViewService(patched, peekViewService);
     if (!page.readonly) {
       patched = patchQuickSearchService(patched, framework);
     }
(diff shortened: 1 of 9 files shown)


L-Sun:
* **#7466** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7466?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

 Join @L-Sun and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/frontend/core/src/components/blocksuite/block-suite-editor/lit-adaper.tsx
```diff
@@ -97,9 +97,7 @@ const usePatchSpecs = (page: Doc, shared: boolean, mode: DocMode) => {
       patchReferenceRenderer(patched, reactToLit, referenceRenderer),
       confirmModal
     );
-    if (!page.readonly) {
```
CatsJuice, line 100: This might lead to all peek-able blocks being peeked, which probably isn't expected. Perhaps CenterPeek should provide some capability to support configuration? @pengx17

L-Sun: Currently, there are four blocks can be peekable:
- `image`
- `surface-ref`  (insert Frame or Group into page)
- `embed-linked-doc`
- `embed-syncd-doc`

The `embed-linked-doc` and `embed-syncd-doc` will show a **empty banner** in shared doc, since their render logic is override by https://github.com/toeverything/AFFiNE/blob/6c82148b5e266a0807c5e7724d84fedf0d29f74e/packages/frontend/core/src/components/blocksuite/block-suite-editor/specs/custom/spec-patchers.tsx#L514-L540

CatsJuice: Seems card view link can still be peeked, and show a loading page.

L-Sun: However, I think it is better to provide an interface to enable or disable the peekable ability based on some context. Like 
```ts
@Peekable({
    showWhen: (ctx) : boolean => {/* ... */}
})
```

L-Sun: > Seems card view link can still be peeked, and show a loading page.

@Catsjuice I have disabled both embed cards peekable in read-only mode in PR https://github.com/toeverything/blocksuite/pull/7596, which has been bump to AFFiNE.

CatsJuice (approved):

PR #7474, opened 2024-07-11
feat(component): add animations to modal

JimmFly:
Add opening and closing animations to modal.

The usage of conditional rendering as shown below is not recommended:
```
open ? (
      <Modal
        open={open}
        ...
      />
    ) : null,
```

When the modal is closed, it gets removed from the DOM instantly without running any exit animations that might be defined in the Modal component.

### DIFF
--- a/packages/frontend/component/src/ui/modal/modal.tsx
+++ b/packages/frontend/component/src/ui/modal/modal.tsx
@@ -10,7 +10,8 @@ import * as VisuallyHidden from '@radix-ui/react-visually-hidden';
 import { assignInlineVars } from '@vanilla-extract/dynamic';
 import clsx from 'clsx';
 import type { CSSProperties } from 'react';
-import { forwardRef, useCallback } from 'react';
+import { forwardRef, useCallback, useEffect } from 'react';
+import { type TransitionState, useTransition } from 'react-transition-state';
 
 import type { IconButtonProps } from '../button';
 import { IconButton } from '../button';
@@ -28,7 +29,8 @@ export interface ModalProps extends DialogProps {
    * @default false
    */
   persistent?: boolean;
-
+  // animation for modal open/close
+  animationTimeout?: number;
   portalOptions?: DialogPortalProps;
   contentOptions?: DialogContentProps;
   overlayOptions?: DialogOverlayProps;
@@ -57,8 +59,10 @@ export const Modal = forwardRef<HTMLDivElement, ModalProps>(
       withoutCloseButton = false,
       modal,
       persistent,
-
+      animationTimeout = 120,
       portalOptions,
+      open: customOpen,
+      onOpenChange: customOnOpenChange,
       contentOptions: {
         style: contentStyle,
         className: contentClassName,
@@ -68,6 +72,7 @@ export const Modal = forwardRef<HTMLDivElement, ModalProps>(
       } = {},
       overlayOptions: {
         className: overlayClassName,
+        style: overlayStyle,
         ...otherOverlayOptions
       } = {},
       closeButtonOptions = {},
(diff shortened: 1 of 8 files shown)


JimmFly:
* **#7474** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7474?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#7465** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7465?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`

<h2></h2>

packages/frontend/component/src/ui/modal/modal.tsx
```diff
@@ -76,11 +81,44 @@ export const Modal = forwardRef<HTMLDivElement, ModalProps>(
...
+    });
+    useEffect(() => {
+      toggle(customOpen);
+    }, [customOpen]);
+
+    const handleOpenChange = useCallback(
+      (open: boolean) => {
+        if (!open) {
+          toggle(open);
+          const timeout = setTimeout(() => {
```
CatsJuice: It's a little strange to handle timeout ourself, will it be better to handle by `react-transition-state`? Maybe something like `onStateChange`. What do you think about @pengx17 ?

pengx17: I think the reason why adding timeout here is to make sure the content wrapped in the modal will still show during exiting when closing is triggered by modal itself like clicking outside. However the issue still happens when the upstream changes changes `open` to false.

The fix here is not complete. Maybe you can just skip this patch - or using web transition api instead.

CatsJuice: Not sure if it's okay this way:

```tsx
const [{}, toggle] = useTransition({
  timeout: animationTimeout,
  onStateChange: useCallback(({ current }) => {
    if (current.status === 'exited') customOnOpenChange?.(false)
    if (current.status === 'entered') customOnOpenChange?.(true)
  }, []),
})

<Dialog.Root
  onOpenChange={toggle}
/>
```

pengx17: yeah, better add an animation started/finished callbacks instead of customOnOpenChange

CatsJuice: Both `onOpenChange` & `onOpenChanged`, but may need to migrate all used.

pengx17: @JimmFly IMO trying to encapsulate content's view transition state (during exiting) inside of this modal dialog seems too tricky without view transition api, since there is no better way to do it when upstream changes open to false.

packages/frontend/core/src/components/page-list/view/edit-collection/hooks.tsx
```diff
@@ -25,6 +30,7 @@ export const useSelectPage = ({
         height="80%"
         overlayOptions={{ style: { backgroundColor: 'transparent' } }}
         contentOptions={{
+          'aria-describedby': undefined,
```
CatsJuice: Is this prop necessary for all `DialogContent`? If so, I think it's better to provide default value globally because we may forget to add this.

JimmFly: QQ_1721108596413
This property was added to resolve the warning in the image, and I have added it directly globally.

CatsJuice (approved):

packages/frontend/core/src/components/page-list/view/use-edit-collection.tsx
```diff
@@ -50,19 +54,23 @@ export const useEditCollectionName = ({
...
+  }, []);
 
   return {
-    node: data ? (
+    node: (
       <CreateCollectionModal
         showTips={showTips}
         title={title}
-        init={data.name}
+        init={data?.name ?? 'initial value'}
```
EYHN: 'initial value' by design?

EYHN (approved):

PR #7481, opened 2024-07-12
feat: support save chat to block

graydawnc:


packages/frontend/core/src/blocksuite/presets/ai/chat-panel/actions/actions-handle.ts
```diff
@@ -14,25 +16,154 @@ import {
...
+async function constructChatBlockMessages(
+  chatMessages: ChatMessage[],
+  latestMessageId: string
+) {
+  const latestIndex = chatMessages.findIndex(
+    item => item.id === latestMessageId
+  );
+  const historyMessages =
+    latestIndex === -1 ? [] : chatMessages.slice(0, latestIndex + 1);
+  if (historyMessages.length < 2 || historyMessages.length % 2 !== 0) {
```
forehalo: what's the meaning of `historyMessages.length % 2 !== 0`

packages/frontend/core/src/blocksuite/presets/ai/chat-panel/chat-panel-input.ts
```diff
@@ -466,6 +473,24 @@ export class ChatPanelInput extends WithDisposable(LitElement) {
...
+
+        if (!this.chatContextValue.chatSessionId) {
+          this.updateContext({
+            chatSessionId: AIProvider.LAST_ACTION_SESSIONID,
+          });
+        }
+
+        const { items } = this.chatContextValue;
+        const last = items[items.length - 1] as ChatMessage;
+        if (!last.id) {
```
EYHN, line 485: Strange logic, is this to deal with the inability to get the message id and session id from the stream? Can we improve the back-end? In the stream before returning the message content, return the message ID.

darkskygit: server can provide a event that included the final message id when the stream is completed
events have been used to provide the execution status of the workflow, but the frontend currently has no mechanism to read events, which requires some refactor
there should be no refactoring scheduled for this period? cc @pengx17

pengx17: I don't quite get it. Are you suggesting that the source event stream should emit more event types including some metadata/status about the current stream, and then in frontend the stream abstraction (i.e., a new version of toTextStream) produce event objects instead of plain text?

darkskygit: what I mean is that if want to save requests, it is feasible to provide the message id in the event source. The interfaces of fal and openai are also doing the same thing, and workflow loading may need to know the execution status of the backend workflow

however, for the scenario of this pr, assemble and maintain the message status through the return value on the frontend will make the code more complex, so I recommend **making one more request to the backend** to get the latest message content

pengx17: I see.

forehalo (approved):

PR #7539, opened 2024-07-18
feat: improve workflow performance

darkskygit:


packages/frontend/core/src/components/blocksuite/block-suite-editor/ai/request.ts
```diff
@@ -44,6 +44,34 @@ export function createChatSession({
   });
 }
 
+async function resizeImage(blob: Blob | File): Promise<Blob | null> {
+  try {
+    const img = new Image();
+    img.src = URL.createObjectURL(blob);
```
pengx17: The url created by createObjectURL should be revoked later

pengx17 (approved):

PR #7549, opened 2024-07-19
chore: fix assets prefix

darkskygit:


tools/cli/src/webpack/s3-plugin.ts
```diff
@@ -34,7 +34,7 @@ export class WebpackS3Plugin implements WebpackPluginInstance {
         const putObjectCommandOptions: PutObjectCommandInput = {
           Body: assetSource,
           Bucket: R2_BUCKET,
-          Key: asset,
+          Key: join(process.env.BUILD_TYPE || 'canary', asset),
```
forehalo, line 37: whoops, I forgot the public path suffix won't be included in asset name.
but this looks not stable, we should update webpack output config instead

PR #7566, opened 2024-07-22
feat: add blocksuite tag in Sentry

akumatus:
Should we add a independent tag or just use advanced search？

### DIFF
--- a/packages/frontend/core/src/bootstrap/global-catch.ts
+++ b/packages/frontend/core/src/bootstrap/global-catch.ts
@@ -1,0 +1,11 @@
+import * as Sentry from '@sentry/react';
+
+window.addEventListener('error', event => {
+  const { error } = event;
+  Sentry.withScope(scope => {
+    if (error.stack.includes('blocksuite')) {
+      scope.setTag('category', 'blocksuite');
+    }
+    Sentry.captureException(error);
+  });
+});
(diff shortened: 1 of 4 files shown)


akumatus:
* **#7566** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7566?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/components/affine/affine-error-boundary/index.tsx
```diff
@@ -25,8 +26,21 @@
     console.error('Uncaught error:', error, componentStack);
   }, []);
 
+  const beforeCapture = useCallback(
+    (scope: Scope, error: unknown, componentStack: string | undefined) => {
+      if (componentStack?.includes('blocksuite')) {
```
pengx17, line 31: Will this work for a bundled app?

packages/frontend/core/src/bootstrap/global-catch.ts
```diff
@@ -0,0 +1,11 @@
+import * as Sentry from '@sentry/react';
+
```
Saul-Mirone, line 2: Maybe also https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event

PR #7567, opened 2024-07-22
chore: update name in one shot

darkskygit:


forehalo (approved):

packages/backend/server/src/data/migrations/1721299086340-refresh-unnamed-user.ts
```diff
@@ -1,25 +1,13 @@
-import { PrismaClient, User } from '@prisma/client';
```
Keg12, line 1: ##_@keg12$!
_

PR #7601, opened 2024-07-25
feat: support chatting in center peek

graydawnc:


pengx17 (commented):
Any advice on how to review this PR?

graydawnc:
> Any advice on how to review this PR?

Maybe can review if there is any logic that seems strange, or the component implementation is unreasonable?

packages/frontend/core/src/blocksuite/presets/ai/chat-panel/index.ts
```diff
@@ -122,6 +122,7 @@ export class ChatPanel extends WithDisposable(ShadowlessElement) {
         this._chatSessionId = history.sessionId;
         this.chatContextValue.chatSessionId = history.sessionId;
         items.push(...history.messages);
+        AIProvider.LAST_ROOT_SESSION_ID = history.sessionId;
```
EYHN, line 125: why?

EYHN (approved):

packages/frontend/core/src/blocksuite/presets/ai/chat-panel/index.ts
graydawnc: Each doc will have one and only one root chat session, which is the session corresponding to the conversation in the right sidebar. It is needed when conducting conversations or forking sessions. This session id needs to be recorded when loading the sidebar for the first time or switching doc.

PR #7606, opened 2024-07-25
refactor(core): adjust modal animation

EYHN:
<div class='graphite__hidden'>
          <div>🎥 Video uploaded on Graphite:</div>
            <a href="https://app.graphite.dev/media/video/g3jz87HxbjOJpXV3FPT7/529d6c3f-4b23-43ac-84cc-171713d3dc72.mp4">
              
            </a>
          </div>
<video src="https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/g3jz87HxbjOJpXV3FPT7/529d6c3f-4b23-43ac-84cc-171713d3dc72.mp4">CleanShot 2024-07-25 at 20.04.01.mp4</video>

When a modal is closed, sometimes its components are completely unmounted from the component tree, making it difficult to animate. This pr defining a custom element as the container of ReactDOM.portal, rewriting the `removeChild` function, and use `startViewTransition` when ReactDOM calls it to implement the animation.

# Save Input

Some inputs use blur event to save data, but when they are unmounted, blur event will not be triggered at all. This pr changes blur event to native addEventListener, which will be called after the DOM element is unmounted, so as to save data in time.

### DIFF
--- a/packages/frontend/component/src/ui/input/input.tsx
+++ b/packages/frontend/component/src/ui/input/input.tsx
@@ -2,21 +2,26 @@ import clsx from 'clsx';
 import type {
   ChangeEvent,
   CSSProperties,
-  FocusEventHandler,
   ForwardedRef,
   InputHTMLAttributes,
   KeyboardEvent,
   KeyboardEventHandler,
   ReactNode,
 } from 'react';
-import { forwardRef, useCallback, useLayoutEffect, useRef } from 'react';
+import {
+  forwardRef,
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useRef,
+} from 'react';
 
 import { input, inputWrapper } from './style.css';
 
 export type InputProps = {
   disabled?: boolean;
   onChange?: (value: string) => void;
-  onBlur?: FocusEventHandler<HTMLInputElement>;
+  onBlur?: (ev: FocusEvent & { currentTarget: HTMLInputElement }) => void;
   onKeyDown?: KeyboardEventHandler<HTMLInputElement>;
   autoSelect?: boolean;
   noBorder?: boolean;
@@ -27,7 +32,7 @@ export type InputProps = {
   type?: HTMLInputElement['type'];
   inputStyle?: CSSProperties;
   onEnter?: () => void;
-} & Omit<InputHTMLAttributes<HTMLInputElement>, 'onChange' | 'size'>;
+} & Omit<InputHTMLAttributes<HTMLInputElement>, 'onChange' | 'size' | 'onBlur'>;
 
 export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
   {
@@ -43,6 +48,7 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
     endFix,
     onEnter,
     onKeyDown,
+    onBlur,
     autoFocus,
     autoSelect,
     ...otherProps
@@ -58,6 +64,17 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
       }
     }
   }, [autoFocus, autoSelect, upstreamRef]);
+
+  // use native blur event to get event after unmount
+  // don't use useLayoutEffect here, because the cleanup function will be called before unmount
+  useEffect(() => {
+    if (!onBlur) return;
+    inputRef.current?.addEventListener('blur', onBlur as any);
+    return () => {
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+      inputRef.current?.removeEventListener('blur', onBlur as any);
+    };
+  }, [onBlur]);
 
   return (
     <div
--- a/packages/frontend/component/src/ui/modal/modal.tsx
+++ b/packages/frontend/component/src/ui/modal/modal.tsx
@@ -10,8 +10,7 @@ import * as VisuallyHidden from '@radix-ui/react-visually-hidden';
 import { assignInlineVars } from '@vanilla-extract/dynamic';
 import clsx from 'clsx';
 import type { CSSProperties } from 'react';
-import { forwardRef, useCallback, useEffect } from 'react';
-import { type TransitionState, useTransition } from 'react-transition-state';
+import { forwardRef, useCallback } from 'react';
 
 import type { IconButtonProps } from '../button';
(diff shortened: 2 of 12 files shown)


EYHN:
* **#7606** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7606?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/component/src/ui/modal/modal.tsx
```diff
@@ -77,112 +92,135 @@ export const Modal = forwardRef<HTMLDivElement, ModalProps>(
...
+      // clearing the previous animation can avoid ghosting
+      lastModalAnimationCleanup.current?.();
+      return () => {
+        if (overlayRef.current && contentRef.current) {
+          // eslint-disable-next-line react-hooks/exhaustive-deps
+          const clonedOverlay = overlayRef.current.cloneNode(
+            true
+          ) as HTMLDivElement;
+          // eslint-disable-next-line react-hooks/exhaustive-deps
+          const clonedContent = contentRef.current.cloneNode(
```
pengx17: I am a little concerned about cloning node when it contains web components in the tree. It could introduce some unexpected side effects when node is mounted/dismounted, and could even crash the app if they are not correctly configured.

EYHN: I changed the solution to no longer need to copy the dom

packages/frontend/component/src/ui/modal/modal.tsx
```diff
@@ -48,141 +45,178 @@ const getVar = (style: number | string = '', defaultValue = '') => {
...
-          // add more status if needed
-          if (current.status === 'exited') customOnOpenChange?.(false);
-          if (current.status === 'entered') customOnOpenChange?.(true);
-        },
-        [customOnOpenChange]
-      ),
+/**
+ * This component is a hack to support `startViewTransition` in the modal.
+ */
+class ModalTransitionContainer extends HTMLElement {
```
pengx17, line 51: I think we can use Lit here

packages/frontend/core/src/pages/index.tsx
```diff
@@ -152,13 +152,15 @@ export const Component = () => {
...
           top: '50%',
+          zIndex: 1,
         }}
       >
         <Menu
           rootOptions={{
             open: true,
           }}
           items={<UserWithWorkspaceList />}
+          noPortal
```
pengx17, line 162: Why noPortal here?

packages/frontend/component/src/ui/modal/styles.css.ts
```diff
@@ -64,6 +47,13 @@ export const modalContentWrapper = style({
   alignItems: 'center',
   justifyContent: 'center',
   zIndex: cssVar('zIndexModal'),
+  animation: `${contentShow} 150ms cubic-bezier(0.42, 0, 0.58, 1)`,
+  animationFillMode: 'forwards',
+  viewTransitionName: 'modal-content',
```
pengx17: viewTransitionName is globally unique. we may need to generate an hash for it

packages/frontend/component/src/ui/input/input.tsx
```diff
@@ -2,22 +2,28 @@ import clsx from 'clsx';
...
+  useRef,
+} from 'react';
 
 import { input, inputWrapper } from './style.css';
 
 export type InputProps = {
   disabled?: boolean;
   onChange?: (value: string) => void;
-  onBlur?: FocusEventHandler<HTMLInputElement>;
+  onBlur?: (ev: FocusEvent & { currentTarget: HTMLInputElement }) => void;
```
forehalo, line 24: `currentTarget` is added as parameter but not passed when called

packages/frontend/core/src/components/affine/page-properties/property-row-value-renderer.tsx
```diff
@@ -87,14 +87,25 @@ export const TextValue = ({ property }: PropertyRowValueProps) => {
   const handleClick = useCallback((e: React.MouseEvent) => {
     e.stopPropagation();
   }, []);
+  const ref = useRef<HTMLTextAreaElement>(null);
   const handleBlur = useCallback(
-    (e: React.ChangeEvent<HTMLTextAreaElement>) => {
+    (e: FocusEvent) => {
+      console.log(123);
```
forehalo: console.log(456);

packages/frontend/core/src/pages/index.tsx
EYHN: This portal will cover the dialog, it is zindex higher than dialog.

packages/frontend/component/src/ui/input/input.tsx
EYHN: I just want to modify the FocusEvent here, Actually event.currentElement is always this `<input />`.

packages/frontend/component/src/ui/modal/modal.tsx
EYHN: I think this is enough since we don't need to render anything. Are there any benefits to using lit?

pengx17: maybe not. I think lit's syntax may make the code more readable when defining wc.

packages/frontend/component/src/ui/modal/modal.tsx
```diff
@@ -48,141 +45,174 @@ const getVar = (style: number | string = '', defaultValue = '') => {
...
+      cancelAnimationFrame(this.animationFrame);
+    }
+
+    this.animationFrame = requestAnimationFrame(() => {
+      if (typeof document.startViewTransition === 'function') {
+        const nodes = this.pendingTransitionNodes;
+        document.startViewTransition(() => {
+          nodes.forEach(child => {
+            // eslint-disable-next-line unicorn/prefer-dom-node-remove
+            super.removeChild(child);
```
pengx17, line 84: This makes DOM mutation async. Is it safe to do so?

EYHN: should be no problem

pengx17 (approved):
[cool](https://app.graphite.dev//settings/meme-library?org=toeverything)

PR #7624, opened 2024-07-29
fix(component): lit-react disappear when modal in close animation

EYHN:


EYHN:
* **#7624** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7624?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/component/src/lit-react/create-component.ts
```diff
@@ -314,13 +314,10 @@ export const createComponent = <
 
     React.useLayoutEffect(() => {
       const container = containerRef.current;
-      if (!container) {
+      if (!container || container.contains(element)) {
```
pengx17, line 317: The wrapped lit component will never be unmounted?

CatsJuice: I'm also quite worried, it feels dangerous

EYHN: React will unmount the parent element.

PR #7652, opened 2024-07-30
feat(core): improve mixpanel

EYHN:
move @affine/core/utils/mixpanel -> @affine/core/mixpanel

now you can debug mixpanel on browser devtool


EYHN:
* **#7659** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7659?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7652** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7652?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/types/dnd.ts
```diff
@@ -58,6 +58,11 @@ export interface AffineDNDData extends DNDData {
         }
       | {
           at: 'explorer:tags:docs';
+        }
+      | {
+          at: 'tabbar:tab:view';
```
pengx17: remove

pengx17 (commented):
> now you can debug mixpanel on browser devtool
I thought we already have this supported before this change

CatsJuice (approved):

packages/frontend/core/src/mixpanel/index.ts
```diff
@@ -1,7 +1,88 @@
...
+      });
+    },
+    track<
+      T extends string,
+      P extends (T extends keyof MixpanelEvents
+        ? MixpanelEvents[T]
+        : Record<string, unknown>) &
+        GeneralMixpanelEvent,
+    >(event_name: T, properties?: P) {
+      logger.debug('track', event_name, properties);
```
pengx17, line 43: not sure if it is a good idea to enable it on production

EYHN: why not

EYHN: If you worry about performance, there shouldn't be much performance hit if you don't have the browser devtools open.

Maybe we should disable debug output for production by default in @affine/debug so there are no issues, but let's just leave it here for now.

pengx17: IIRC I only enabled logging in the previous version because there was no way of debugging  mixpanel locally previously.

pengx17 (approved):

PR #7660, opened 2024-07-30
fix: ai chat block added later should be with higher index

graydawnc:
[BS-1000](https://linear.app/affine-design/issue/BS-1000/更好地处理新增内容的层级)

### DIFF
--- a/packages/frontend/core/src/blocksuite/presets/ai/_common/chat-actions-handle.ts
+++ b/packages/frontend/core/src/blocksuite/presets/ai/_common/chat-actions-handle.ts
@@ -131,7 +131,7 @@ function getViewportCenter(
 
 // Add AI chat block and focus on it
 function addAIChatBlock(
-  doc: Doc,
+  host: EditorHost,
   messages: ChatMessage[],
   sessionId: string,
   viewportCenter: { x: number; y: number }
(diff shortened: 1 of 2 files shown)


graydawnc:
* **#7660** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7660?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

 Join @donteatfriedrice and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/frontend/core/src/blocksuite/presets/ai/_common/chat-actions-handle.ts
```diff
@@ -158,6 +161,7 @@ function addAIChatBlock(
     {
       xywh: bound.serialize(),
       messages: JSON.stringify(messages),
+      index: layer.generateIndex('affine:embed-ai-chat'),
```
doouding, line 164: If you need to add block at the very top, you could change the parameter to 'shape'. Otherwise, it will be on the second layer instead of the top layer.

doouding (approved):

PR #7672, opened 2024-07-31
fix(electron): electron cmd+r issue

pengx17:
fix AF-1136

PR #7671, opened 2024-07-31
fix(electron): window theme issue after exiting presentation

pengx17:

PR #7666, opened 2024-07-30
fix: using width atom for syncing app headers position

pengx17:
may use global state to replace these sidebar state atoms

fix AF-1109

PR #7675, opened 2024-07-31
refactor(electron): tab title/icon update logic

pengx17:
fix AF-1122
fix AF-1136

### DIFF
--- a/packages/frontend/core/src/modules/workbench/services/desktop-state-synchronizer.ts
+++ b/packages/frontend/core/src/modules/workbench/services/desktop-state-synchronizer.ts
@@ -1,35 +1,13 @@
-import {
-  apis,
-  appInfo,
-  events,
-  type WorkbenchViewMeta,
-} from '@affine/electron-api';
-import { I18n, type I18nKeys, i18nTime } from '@affine/i18n';
-import type { DocsService } from '@toeverything/infra';
-import { Service } from '@toeverything/infra';
-import { combineLatest, filter, map, of, switchMap } from 'rxjs';
+import { apis, appInfo, events } from '@affine/electron-api';
+import { LiveData, Service } from '@toeverything/infra';
 
-import { resolveRouteLinkMeta } from '../../navigation';
-import type { RouteModulePath } from '../../navigation/utils';
-import type { WorkspacePropertiesAdapter } from '../../properties';
 import type { WorkbenchService } from '../../workbench';
-
-const routeModuleToI18n = {
-  all: 'All pages',
-  collection: 'Collections',
-  tag: 'Tags',
-  trash: 'Trash',
-} satisfies Record<RouteModulePath, I18nKeys>;
 
 /**
  * Synchronize workbench state with state stored in main process
  */
 export class DesktopStateSynchronizer extends Service {
-  constructor(
-    private readonly workbenchService: WorkbenchService,
-    private readonly workspaceProperties: WorkspacePropertiesAdapter,
-    private readonly docsService: DocsService
-  ) {
+  constructor(private readonly workbenchService: WorkbenchService) {
     super();
     this.startSync();
   }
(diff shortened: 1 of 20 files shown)


pengx17:
* **#7684** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7684?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7675** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7675?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#7671** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7671?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7672](https://github.com/toeverything/AFFiNE/pull/7672) <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7672?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#7666** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7666?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7665** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7665?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`

<h2></h2>

packages/frontend/core/src/modules/workbench/services/desktop-state-synchronizer.ts
```diff
@@ -170,58 +132,9 @@ export class DesktopStateSynchronizer extends Service {
...
-    return basename + location.pathname + location.search + location.hash;
-  }
-
-  // fill tab view meta with title & moduleName
-  private fillTabViewMeta(
-    view: WorkbenchViewMeta
-  ): WorkbenchViewMeta & { docId?: string } {
-    if (!view.path) {
-      return view;
+  updateViewMeta = async (view: WorkbenchViewMeta) => {
```
EYHN: I think we can just store `title` & `icon` in `View` as a state?

EYHN: in browser we can also use them to create dynamic `<title />`

packages/frontend/core/src/modules/app-tabs-header/views/tab-view-meta.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * Used for updating the meta of a split view tab on the desktop client.
+ */
+export const ViewTabMeta = (meta: SplitViewTabMeta) => {
+  const view = useServiceOptional(ViewService)?.view;
+  const desktopStateSynchronizer = useServiceOptional(DesktopStateSynchronizer);
+  const t = useI18n();
+
+  useEffect(() => {
+    if (view && desktopStateSynchronizer) {
+      desktopStateSynchronizer
```
EYHN: view.setTitle('xxx')

packages/frontend/core/src/modules/app-tabs-header/views/tab-view-meta.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export interface SplitViewTabMeta {
+  title?: string | null;
+  icon: TabIconName;
+}
+
+/**
+ * Used for updating the meta of a split view tab on the desktop client.
+ */
+export const ViewTabMeta = (meta: SplitViewTabMeta) => {
```
EYHN: maybe split this to `ViewTitle` and `ViewIcon`,  I feel that the concept contained in the name ViewTabMeta is a bit complicated.

packages/frontend/core/src/modules/workbench/services/desktop-state-synchronizer.ts
```diff
@@ -80,10 +64,8 @@ export class DesktopStateSynchronizer extends Service {
 
     // sync workbench state with main process
     // also fill tab view meta with title & moduleName
-    this.workspaceProperties.workspace.engine.rootDocState$
+    workbench.views$
```
EYHN: This can be simplified:

```
LiveData.compute((get) => {
  return get(workbench.views$).map((view) => ({
    title: get(view.title$),
    location: get(view.location$),
    ....
  }))
}).subscribe(
  (data) => updateWorkbenchMeta(data)
)
```

EYHN (approved):

packages/frontend/core/src/modules/workbench/services/desktop-state-synchronizer.ts
pengx17: I think we can think about this in the <Title /> component level but not in the service.
In the browser we do not need to persist title/icons in state anyway. We just need to update the core/title.

packages/frontend/core/src/modules/app-tabs-header/views/tab-view-meta.tsx
pengx17: I think this may cause redundant electron api calls so it is merged into one. Also I feel like there is no obvious benefit to split it into two.

packages/frontend/core/src/modules/workbench/services/desktop-state-synchronizer.ts
EYHN: I think that storing the `title` and `icon` in the view is more unified so that they can be synchronized in the same way as `location`.

PR #7681, opened 2024-07-31
feat(core): add globalcontext info to mixpanel track

EYHN:


EYHN:
* **#7681** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7681?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

JimmFly (approved):

packages/frontend/core/src/modules/telemetry/services/telemetry.ts
```diff
@@ -25,14 +18,19 @@ import { WorkbenchService } from '../../workbench';
...
+    private readonly auth: AuthService,
+    private readonly globalContextService: GlobalContextService
+  ) {
     super();
   }
 
   onApplicationStart() {
     const account = this.auth.session.account$.value;
     this.updateIdentity(account);
+    this.registerMiddlewares();
```
pengx17, line 33: I tried to use `ApplicationStarted` event for syncing tabs, however it seems this requires the service to be instantiated before this event has been emitted.

packages/frontend/core/src/pages/workspace/all-page/all-page.tsx
```diff
@@ -25,6 +34,19 @@ export const AllPage = () => {
...
+  const isActiveView = useIsActiveView();
+
+  useEffect(() => {
+    if (isActiveView) {
+      globalContext.isAllDocs.set(true);
+
+      return () => {
+        globalContext.isAllDocs.set(false);
+      };
+    }
```
pengx17, line 46: may abstract this pattern into a new hook

packages/frontend/core/src/modules/telemetry/services/telemetry.ts
EYHN: This is not the case as long as the service is in the root scope

pengx17: I guess it the service must be eagarly created

packages/frontend/core/src/pages/workspace/all-page/all-page.tsx
EYHN: I'll think about how to abstract

pengx17 (approved):

PR #7774, opened 2024-08-07
fix(infra): better search result

EYHN:


EYHN:
* **#7774** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7774?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

pengx17 (approved):

packages/frontend/core/src/modules/docs-search/worker/types.ts
```diff
@@ -34,7 +34,7 @@ export type WorkerInput =
     }
   | {
       type: 'doc';
-      docId: string;
+      storageDocId: string;
```
pengx17, line 37: can you explain changes to `storageDocId`? to increase worker's inefficiency?

EYHN: This is due to historical reasons. The id of the doc stored may not be consistent with the docid. The current solution is to traverse the space map in the rootDoc ​ and query the docid from the storageDocId in the worker.

EYHN: This is to fix the issue that some old docs not automatically update the index

pengx17: I see, thanks!

PR #7795, opened 2024-08-08
feat(i18n): add Arabic

tasnim0tantawi:


JimmFly (approved):
Thank you so much for your valuable contribution in adding Arabic language support to AFFiNE! 🎉

JimmFly:
By the way, if you encounter text overflow and are unable to view the full text, please submit an issue or create a PR to fix it. The same applies to any other issues you may come across, your feedback is invaluable.

PR #7842, opened 2024-08-12
feat(core): use emoji as folder icon

EYHN:

### DIFF
--- a/packages/frontend/core/src/modules/explorer/views/tree/node.tsx
+++ b/packages/frontend/core/src/modules/explorer/views/tree/node.tsx
@@ -14,6 +14,7 @@ import { RenameModal } from '@affine/component/rename-modal';
 import { appSidebarWidthAtom } from '@affine/core/components/app-sidebar/index.jotai';
 import { WorkbenchLink } from '@affine/core/modules/workbench';
 import type { AffineDNDData } from '@affine/core/types/dnd';
+import { extractEmojiIcon } from '@affine/core/utils';
 import { useI18n } from '@affine/i18n';
 import {
   ArrowDownSmallIcon,
@@ -59,7 +60,7 @@ export type ExplorerTreeNodeIcon = React.ComponentType<{
 export const ExplorerTreeNode = ({
   children,
   icon: Icon,
-  name,
+  name: rawName,
   onClick,
   to,
   active,
@@ -68,6 +69,7 @@ export const ExplorerTreeNode = ({
   onRename,
   disabled,
   collapsed,
+  extractEmojiAsIcon,
   setCollapsed,
   canDrop,
   reorderable = true,
@@ -87,6 +89,7 @@ export const ExplorerTreeNode = ({
   active?: boolean;
   reorderable?: boolean;
   defaultRenaming?: boolean;
+  extractEmojiAsIcon?: boolean;
   collapsed: boolean;
   setCollapsed: (collapsed: boolean) => void;
   renameable?: boolean;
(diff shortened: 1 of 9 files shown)


EYHN:
* **#7842** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7842?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#7856** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7856?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`

<h2></h2>

JimmFly (approved):

packages/frontend/core/src/modules/explorer/views/tree/node.tsx
```diff
@@ -299,34 +323,36 @@ export const ExplorerTreeNode = ({
...
+          className={styles.collapsedIconContainer}
+        >
+          <ArrowDownSmallIcon
+            className={styles.collapsedIcon}
+            data-collapsed={collapsed !== false}
           />
         </div>
-      )}
+        {emoji
+          ? emoji
```
CatsJuice: `emoji` should also have a wrapper otherwise may have different width and not aligned with others

packages/frontend/core/src/modules/explorer/views/tree/node.tsx
```diff
@@ -117,6 +120,27 @@ export const ExplorerTreeNode = ({
...
   const [lastInGroup, setLastInGroup] = useState(false);
   const rootRef = useRef<HTMLDivElement>(null);
+  const { emoji, name } = useMemo(() => {
+    if (!extractEmojiAsIcon || !rawName) {
+      return {
+        emoji: null,
+        name: rawName,
+      };
+    }
+    const isStartsWithEmoji = /^(\p{Emoji_Presentation})/u.test(rawName);
```
pengx17: may extract this to a utility function? I guess it might be reused in places like in tab header
image

pengx17 (approved):

pengx17 (approved):

packages/frontend/core/src/modules/explorer/views/tree/node.tsx
```diff
@@ -117,6 +120,27 @@ export const ExplorerTreeNode = ({
...
+    if (isStartsWithEmoji) {
+      // emoji like "👨🏻‍❤️‍💋‍👨🏻" are combined. Graphemer can handle these.
+      const emojiEnd = Graphemer.nextBreak(rawName, 0);
```
pengx17: Should be able to be replaced by Intl.Segmenter

EYHN: We have used Grapemer in the indexer already. `Intl.Segmenter` may have different results on different browsers. I prefer Grapemer.

pengx17: how about Graphemer's perf? i think the native solution may perform better 
https://arc.net/l/quote/ftyvfooe

pengx17: Some interesting discussion https://github.com/eslint/eslint/issues/17835

EYHN: Or you can use this library https://github.com/cometkim/unicode-segmenter.
One issue with Intl.Segmenter is that it doesn't work properly in the CI environment of nodejs because there is no intl data in CI.

pengx17: how does Graphemer deal with intl data in this case?

packages/frontend/core/src/utils/__tests__/extract-emoji-icon.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { extractEmojiIcon } from '../extract-emoji-icon';
+
+test('extract-emoji-icon', () => {
+  expect(extractEmojiIcon('👨🏻‍❤️‍💋‍👨🏻123')).toEqual({
+    emoji: '👨🏻‍❤️‍💋‍👨🏻',
+    rest: '123',
+  });
+
+  expect(extractEmojiIcon('❤️123')).toEqual({
```
pengx17, line 11: it seems ❤️ displays as emoji on windows

EYHN: Its depending on the font family, we adopt a conservative strategy, only exclusive emoji characters

pengx17: yeah i mean maybe we could use other symbols that reduces confusion here

PR #7858, opened 2024-08-13
fix(admin): user form not dynamically updating as expected

JimmFly:
- Fixed the issue that a certain feature must be enabled when creating a user
- Fixed the issue that the modified content was not reset to the default content when exiting the user form without saving after modification
- Fixed the issue that the content did not switch as expected when switching user forms of different users

https://github.com/user-attachments/assets/02567021-9342-4ed1-be77-3bdcbb3d86ab

### DIFF
--- a/packages/frontend/admin/src/modules/accounts/components/user-form.tsx
+++ b/packages/frontend/admin/src/modules/accounts/components/user-form.tsx
@@ -31,9 +31,16 @@ function UserForm({
 }: UserFormProps) {
   const serverConfig = useServerConfig();
 
-  const [changes, setChanges] = useState<Partial<UserInput>>({
-    features: defaultValue?.features ?? [],
-  });
+  const defaultUser: Partial<UserInput> = useMemo(
+    () => ({
+      name: defaultValue?.name ?? '',
+      email: defaultValue?.email ?? '',
+      features: defaultValue?.features ?? [],
+    }),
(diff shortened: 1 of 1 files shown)


JimmFly:
* **#7858** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7858?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#7854** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7854?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`

<h2></h2>

packages/frontend/admin/src/modules/accounts/components/user-form.tsx
```diff
@@ -121,11 +137,7 @@ function UserForm({
             <div key={feature}>
               <ToggleItem
                 name={feature}
-                checked={(
-                  changes.features ??
-                  defaultValue?.features ??
-                  []
-                ).includes(feature)}
+                checked={changes.features?.includes(feature) ?? false}
```
forehalo, line 140: use `defaultChecked`?

JimmFly: Now it works fine. If I use the `defaultChecked` attribute, it also needs to handle the logic of clearing after exiting.

forehalo (approved):

PR #7888, opened 2024-08-15
feat: update ingress health check

darkskygit:


darkskygit:
* **#7888** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7888?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

forehalo (approved):

.github/helm/affine/charts/graphql/templates/backend-config.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: cloud.google.com/v1
```
Brooooooklyn: This config is not managed by helm, you should create it manually

.github/helm/affine/values.yaml
```diff
@@ -36,14 +36,14 @@ graphql:
     type: ClusterIP
     port: 3000
     annotations:
-      cloud.google.com/backend-config: '{"default": "affine-backendconfig"}'
+      cloud.google.com/backend-config: '{"default": "affine-graphql-backendconfig"}'
```
Brooooooklyn: I suggest naming it `affine-api-backendconfig` rather than `affine-graphql-backendconfig` because it's a cross-application configuration.

darkskygit: sync and gql server using different port now, so cannot share same config

PR #7897, opened 2024-08-16
feat(core): share in workspace link

EYHN:
ShareDocsService -> ShareDocsListService
ShareService -> ShareInfoService
(*new) ShareReaderService

`/share/:workspaceId/:docId` -> redirect to -> `/workspace/:workspaceId/:docId`

workspace loading process

1. find workspace in workspace list
2. (if not found) revalidate workspace list
3. (if still not found) try load share page
4. (if share page found) => share page
5. (if share page not found) => 404
6. (if workspace found) => workspace page


EYHN:
* **#7897** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7897?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

JimmFly (approved):

packages/frontend/core/src/modules/share-doc/stores/share-reader.ts
```diff
@@ -0,0 +1,42 @@
...
+      );
+      const publishMode = docResponse.headers.get(
+        'publish-mode'
+      ) as DocMode | null;
+      const docBinary = await docResponse.arrayBuffer();
+
+      const workspaceResponse = await this.fetchService.fetch(
+        `/api/workspaces/${workspaceId}/docs/${workspaceId}`
+      );
+      const workspaceBinary = await workspaceResponse.arrayBuffer();
```
pengx17, line 24: I think this is a security breach that leaks metadata of the whole workspace even when the user has no access to the workspace.
Is that necessary to load it for shared page?

pengx17 (approved):

packages/frontend/core/src/pages/workspace/index.tsx
```diff
@@ -37,24 +38,104 @@ declare global {
...
+  const listLoading = useLiveData(workspacesService.list.isRevalidating$);
+  const workspaces = useLiveData(workspacesService.list.workspaces$);
   const meta = useMemo(() => {
     return workspaces.find(({ id }) => id === params.workspaceId);
   }, [workspaces, params.workspaceId]);
 
-  const workspace = useWorkspace(meta);
-  const globalContext = useService(GlobalContextService).globalContext;
+  // if listLoading is false, we can show 404 page, otherwise we should show loading page.
+  useEffect(() => {
```
pengx17, line 79: I think this section could be moved into route loader

packages/frontend/core/src/pages/workspace/index.tsx
```diff
@@ -37,24 +38,104 @@ declare global {
...
+    return;
+  }, [listLoading, meta, workspaceNotFound, workspacesService]);
+
+  if (workspaceNotFound) {
+    if (
+      detailDocRoute /*  */ &&
+      environment.isBrowser /* only browser has share page */
+    ) {
+      return (
+        <SharePage
```
pengx17, line 106: Not sure but I think the rendering should be handled in the child route, not in the parent route for clarity.

EYHN: that is when the workspace does not exist, the sub-route will not be entered at all

packages/frontend/core/src/modules/share-doc/stores/share-reader.ts
EYHN: I'm not sure if blocksuite can render with just subdoc.

packages/frontend/core/src/pages/workspace/index.tsx
EYHN: the routes of the workspace page and the share page are exactly the same,

but the workspace has a sidebar and a topbar that the share page does not have.

So for share page we cannot render the workspace route.

packages/frontend/core/src/pages/workspace/index.tsx
EYHN: I don't like using router-loader, I think it's hard to control.

And with this we can also automatically recover from the 404 page after the workspace list changes, workspace appear again (maybe the network is restored).

packages/frontend/core/src/pages/workspace/index.tsx
pengx17: I see 👌

packages/frontend/core/src/pages/workspace/index.tsx
pengx17: How about making the steps 1~6 into a state machine in a service? It is hard to maintain the logic into individual useEffects.

packages/frontend/core/src/modules/share-doc/stores/share-reader.ts
EYHN: track in AF-1259

packages/frontend/core/src/pages/workspace/index.tsx
EYHN: I encountered some difficulties when I tried it, and it doesn't seem easy to write it as a state machine.

But you are right, since we will need a mobile version of the UI in the future, we also need to share the logic here. Maybe we need to extract them to one service.

EYHN: We can discuss a suitable solution later.

PR #7898, opened 2024-08-16
chore(electron): re-enable sentry for electron

pengx17:
related
https://github.com/toeverything/AFFiNE/pull/7001

### DIFF
--- a/packages/frontend/electron/src/main/index.ts
+++ b/packages/frontend/electron/src/main/index.ts
@@ -2,6 +2,8 @@ import './security-restrictions';
 
 import path from 'node:path';
 
+import * as Sentry from '@sentry/electron/main';
+import { IPCMode } from '@sentry/electron/main';
 import { app } from 'electron';
 
 import { createApplicationMenu } from './application-menu/create';
@@ -12,6 +14,7 @@ import { registerEvents } from './events';
 import { registerHandlers } from './handlers';
 import { logger } from './logger';
 import { registerProtocol } from './protocol';
+import { isOnline } from './ui';
 import { registerUpdater } from './updater';
 import { launch } from './windows-manager/launcher';
 import { launchStage } from './windows-manager/stage';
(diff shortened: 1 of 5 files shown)


pengx17:
* **#7898** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7898?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#7894** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7894?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7887** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7887?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`

<h2></h2>

packages/frontend/electron/src/main/index.ts
```diff
@@ -75,3 +78,18 @@ app
...
+  // https://docs.sentry.io/platforms/javascript/guides/electron/
+  Sentry.init({
+    dsn: process.env.SENTRY_DSN,
+    environment: process.env.BUILD_TYPE ?? 'development',
+    ipcMode: IPCMode.Protocol,
+    transportOptions: {
+      maxAgeDays: 30,
+      maxQueueSize: 100,
+      shouldStore: () => !isOnline,
+      shouldSend: () => isOnline,
```
EYHN, line 92: should this return a promise?  Not sure how this works

pengx17: it says "If this function returns false, `shouldStore` will be called to determine if the envelope should be stored.". It seems when an envelope is about to send but `shouldSend` returns false, it will store the event into the queue. I am not sure when the queue will be flushed though.

```ts
import * as Sentry from "@sentry/electron/main";

Sentry.init({
  dsn: "https://examplePublicKey@o0.ingest.sentry.io/0",
  transportOptions: {
    /* The maximum number of days to keep an event in the queue. */
    maxQueueAgeDays: 30,
    /* The maximum number of events to keep in the queue. */
    maxQueueCount: 30,
    /**
     * Called before we attempt to send an envelope to Sentry.
     *
     * If this function returns false, `shouldStore` will be called to determine if the envelope should be stored.
     *
     * Default: () => true
     *
     * @param envelope The envelope that will be sent.
     * @returns Whether we should attempt to send the envelope
     */
    shouldSend: (envelope: Envelope) => boolean | Promise<boolean>;
    /**
     * Called before an event is stored.
     *
     * Return false to drop the envelope rather than store it.
     *
     * Default: () => true
     *
     * @param envelope The envelope that failed to send.
     * @param error The error that occurred.
     * @param retryDelay The current retry delay in milliseconds.
     * @returns Whether we should store the envelope in the queue
     */
    shouldStore: (envelope: Envelope, error: Error, retryDelay: number) => boolean | Promise<boolean>;
  },
});
```

Brooooooklyn (approved):

PR #7956, opened 2024-08-23
feat(core): editor setting service

EYHN:
define editor setting schema in `packages/frontend/core/src/modules/editor-settting/schema.ts`

e.g.

```ts
const BSEditorSettingSchema = z.object({
  connector: z.object({
    stroke: z
      .union([
        z.string(),
        z.object({
          dark: z.string(),
          light: z.string(),
        }),
      ])
      .default('#000000'), // default is necessary
  }),
});
```

schema can be defined in a nested way. EditorSetting api is in flat way:

editorSetting api:

```ts
editorSetting.settings$ === {
  'connector.stroke': '#000000'
}
editorSetting.set('connector.stroke', '#000')
```

and use `expandFlattenObject` function can restore the flattened structure to a nested structure. nested structure is required by blocksuite

```ts
editorSetting.settings$.map(expandFlattenObject) === {
  connector: {
    stroke: '#000000'
  }
}
```

### DIFF
--- a/packages/frontend/core/src/modules/editor-settting/__test__/editor-setting.spec.ts
+++ b/packages/frontend/core/src/modules/editor-settting/__test__/editor-setting.spec.ts
@@ -1,0 +1,72 @@
+import { Framework, GlobalState, MemoryMemento } from '@toeverything/infra';
+import { expect, test } from 'vitest';
+
+import { unflattenObject } from '../../../utils/unflatten-object';
+import { EditorSetting } from '../entities/editor-setting';
+import { GlobalStateEditorSettingProvider } from '../impls/global-state';
+import { EditorSettingProvider } from '../provider/editor-setting-provider';
+import { EditorSettingService } from '../services/editor-setting';
+
+test('editor setting service', () => {
+  const framework = new Framework();
+
+  framework
+    .service(EditorSettingService)
+    .entity(EditorSetting, [EditorSettingProvider])
+    .impl(EditorSettingProvider, GlobalStateEditorSettingProvider, [
+      GlobalState,
+    ])
+    .impl(GlobalState, MemoryMemento);
+
+  const provider = framework.provider();
+
+  const editorSettingService = provider.get(EditorSettingService);
+
+  // default value
+  expect(editorSettingService.editorSetting.settings$.value).toMatchObject({
+    fontFamily: 'Sans',
+    'connector.stroke': '#000000',
+  });
+
+  editorSettingService.editorSetting.set('fontFamily', 'Serif');
+  expect(editorSettingService.editorSetting.settings$.value).toMatchObject({
+    fontFamily: 'Serif',
+  });
+
+  // nested object, should be serialized
+  editorSettingService.editorSetting.set('connector.stroke', {
+    dark: '#000000',
+    light: '#ffffff',
+  });
+  expect(
+    (
+      editorSettingService.editorSetting
+        .provider as GlobalStateEditorSettingProvider
+    ).get('connector.stroke')
+  ).toBe('{"dark":"#000000","light":"#ffffff"}');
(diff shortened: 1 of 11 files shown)


EYHN:
* **#7956** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/7956?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/utils/flatten-object.ts
```diff
@@ -0,0 +1,35 @@
+export function flattenObject(ob: any, target: any = {}, prefix = '') {
```
pengx17: how about array?

packages/frontend/core/src/utils/flatten-object.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (typeof ob[key] === 'object' && ob[key] !== null) {
+      flattenObject(ob[key], target, prefix + key + '.');
+    } else {
+      target[prefix + key] = ob[key];
+    }
+  }
+  return target;
+}
+
+export function expandFlattenObject(ob: any) {
```
pengx17: the name looks weird. how about `unflattenObject`?

packages/frontend/core/src/modules/editor-settting/impls/global-state.ts
```diff
@@ -0,0 +1,35 @@
...
+ * just for testing, vary poor performance
+ */
+export class GlobalStateEditorSettingProvider
+  extends Service
+  implements EditorSettingProvider
+{
+  constructor(public readonly globalState: GlobalState) {
+    super();
+  }
+  set(key: string, value: string): void {
```
pengx17, line 19: it seems there is no way to remove a k/v, though it might not be an issue for now

EYHN: Yes, there is no need to delete the settings yet, we can easily add it in the future

packages/frontend/core/src/modules/editor-settting/__test__/editor-setting.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    editorSettingService.editorSetting.settings$.value
+  );
+  expect(expended).toMatchObject({
+    fontFamily: 'Sans',
+    connector: {
+      stroke: {
+        dark: '#000000',
+        light: '#ffffff',
+      },
+    },
```
akumatus, line 70: Is there a way to get `BSEditorSettingSchema` only?

EYHN: No, I don’t think there is any need to distinguish between blocksuite and others separately.

We can manually maintain them to ensure there are no conflicts

Editorsetting should not change frequently and there no performance issues.

packages/frontend/core/src/utils/flatten-object.ts
EYHN: I think string is more intuitive here

pengx17: I mean what if value of `ob[key]` is an array

EYHN: I deleted this function because we don't actually use it. We actually use
https://github.com/toeverything/AFFiNE/blob/8982886440069c44b1dbf24832eb3b249739b490/packages/frontend/core/src/modules/editor-settting/schema.ts#L41

It will stop flatten when it encounters any non-zod.object element.

pengx17 (approved):

JimmFly (approved):

akumatus (approved):

PR #7977, opened 2024-08-26
fix: update package.json

vrzit:


vrzit (commented):
"dev": "yarn workspace @affine/cli dev --static"
I can be good solution in local test.

PR #8039, opened 2024-09-01
feat(mobile): sign in page

pengx17:
fix AF-1237

### DIFF
--- a/packages/frontend/mobile/src/pages/sign-in.tsx
+++ b/packages/frontend/mobile/src/pages/sign-in.tsx
@@ -1,7 +1,37 @@
-// Default route fallback for mobile
-import { SignIn } from '@affine/core/pages/auth/sign-in';
+import {
+  RouteLogic,
+  useNavigateHelper,
+} from '@affine/core/hooks/use-navigate-helper';
+import { AuthService } from '@affine/core/modules/cloud';
+import { useLiveData, useService } from '@toeverything/infra';
+import { useEffect } from 'react';
+import { useNavigate, useSearchParams } from 'react-router-dom';
+
+import { MobileSignIn } from '../views/sign-in/mobile-sign-in';
 
 export const Component = () => {
-  // placeholder impl
-  return <SignIn />;
+  const session = useService(AuthService).session;
+  const status = useLiveData(session.status$);
+  const isRevalidating = useLiveData(session.isRevalidating$);
+  const navigate = useNavigate();
+  const { jumpToIndex } = useNavigateHelper();
+  const [searchParams] = useSearchParams();
+  const isLoggedIn = status === 'authenticated' && !isRevalidating;
+
+  useEffect(() => {
+    if (isLoggedIn) {
+      const redirectUri = searchParams.get('redirect_uri');
+      if (redirectUri) {
+        navigate(redirectUri, {
+          replace: true,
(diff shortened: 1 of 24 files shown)


pengx17:
* **#8039** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8039?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#8048** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8048?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`

<h2></h2>

packages/frontend/mobile/src/pages/sign-in.tsx
```diff
@@ -1,7 +1,73 @@
...
+} from '@affine/core/hooks/use-navigate-helper';
+import { AuthService } from '@affine/core/modules/cloud';
+import { useLiveData, useService } from '@toeverything/infra';
+import { useAtom } from 'jotai';
+import { useCallback, useEffect } from 'react';
+import { useNavigate, useSearchParams } from 'react-router-dom';
+
+import { MobileSignInLayout } from '../components/sign-in/layout';
 
 export const Component = () => {
```
CatsJuice, line 12: We have previously agreed that sign-in should use a full-screen modal (implemented in the same way as settings page).

pengx17: I think we still need a standalone sign-in page for mobile since we may only allow logged in user to use the mobile version.
I will update the login modal accordingly.

packages/frontend/mobile/src/components/sign-in/background.tsx
```diff
@@ -0,0 +1,203 @@
+import { useTheme } from 'next-themes';
+
+import * as styles from './background.css';
+
+const artsLight = `<svg viewBox="0 0 393 268" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Brooooooklyn: put it into a separate file and name it with `arts-light.inline.svg`, then add a rule in our [webpack config](./tools/cli/src/webpack/config.ts):

```ts
{
    test: /\.inline\.svg$/,
    type: 'asset/inline',
 },
```

And then u can import the file and use it in css.

packages/frontend/mobile/src/router.tsx
```diff
@@ -13,6 +16,25 @@ import { Component as Home } from './pages/workspace/home';
...
 import { Component as Tag } from './pages/workspace/tag';
 import { Component as TagDetail } from './pages/workspace/tag/detail';
+import { AllWorkspaceModals } from './provider/model-provider';
+
+function RootRouter() {
+  const navigate = useNavigate();
+  const [ready, setReady] = useState(false);
+  useEffect(() => {
+    // a hack to make sure router is ready
+    setReady(true);
```
forehalo, line 26: any detail?

pengx17: Not sure ... this is copied from '@affine/core/router';

forehalo: approve but need to be refactored

forehalo (approved):

CatsJuice (approved):

PR #8042, opened 2024-09-02
chore: bump bs with new extension api

Saul-Mirone:


Saul-Mirone:
* **#8042** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8042?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

 Join @Saul-Mirone and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/common/env/package.json
```diff
@@ -3,8 +3,8 @@
   "private": true,
   "type": "module",
   "devDependencies": {
-    "@blocksuite/global": "0.17.0-canary-202408191538-1511d04",
-    "@blocksuite/store": "0.17.0-canary-202408191538-1511d04",
+    "@blocksuite/global": "0.0.0-canary-20240902070217",
+    "@blocksuite/store": "0.0.0-canary-20240902070217",
```
fundon, line 7: Need an official version

packages/frontend/core/src/commands/affine-creation.tsx
```diff
@@ -31,7 +32,7 @@ export function registerAffineCreationCommands({
       run() {
         track.$.cmdk.creation.createDoc({ mode: 'page' });
 
-        pageHelper.createPage('page');
+        pageHelper.createPage('page' as DocMode);
```
akumatus, line 35: Better to use `DocMode.Page` instead of `as DocMode` here?

packages/frontend/core/src/components/blocksuite/block-suite-editor/specs/common.ts
```diff
@@ -42,6 +39,6 @@ export const CommonBlockSpecs: BlockSpec[] = [
   AICodeBlockSpec,
   AIImageBlockSpec,
   AIParagraphBlockSpec,
-  AIChatBlockSpec,
-  EdgelessAIChatBlockSpec,
-];
+  // AIChatBlockSpec,
+  // EdgelessAIChatBlockSpec,
```
akumatus: Delete unused AI Spec

packages/frontend/core/src/components/blocksuite/block-suite-editor/specs/custom/attachment-block.ts
```diff
@@ -14,7 +19,15 @@ class CustomAttachmentBlockService extends AttachmentBlockService {
...
+export const CustomAttachmentBlockSpec: ExtensionType[] = [
   ...AttachmentBlockSpec,
-  service: CustomAttachmentBlockService,
-};
+  {
+    setup: di => {
+      di.override(
+        BlockServiceIdentifier('affine:attachment'),
+        CustomAttachmentBlockService,
+        [StdIdentifier, BlockFlavourIdentifier('affine:attachment')]
```
akumatus, line 29: Can we improve and simplify this override API like `di.override('affine:attachment').to(CustomAttachmentBlockService)` in the feature?

Saul-Mirone: We will remove block services

packages/common/env/package.json
Saul-Mirone: I prefer to not add version in snapshot version. And we should stop using snapshot versions in affine.

packages/frontend/core/src/commands/affine-creation.tsx
Saul-Mirone: It breaks the unit test somehow. So I have to use type import here.

packages/frontend/core/src/components/blocksuite/block-suite-editor/specs/common.ts
Saul-Mirone: Done

Flrande (approved):

PR #8054, opened 2024-09-02
fix(core): handle text overflow styles for search box

SOUMITRO-SAHA:
### Description:

This PR implements dynamic text handling for the search box to ensure that text is truncated with an ellipsis when it exceeds the box's size. This enhancement improves the display of search box labels across different languages by ensuring that text is consistently presented, even when it is longer than the available space.

### Changes Made:

- Added CSS to dynamically apply the ellipsis property to the search box text.
- Ensured that the truncation behaviour adjusts based on the size of the search box, making it suitable for various languages.

### Benefits:

- Provides a more consistent and readable user interface.
- Enhances support for multi-language text, preventing overflow and maintaining a clean design.

image

image


packages/frontend/core/src/components/app-sidebar/quick-search-input/index.css.ts
```diff
@@ -28,3 +28,9 @@ export const shortcutHint = style({
   color: cssVarV2('text/tertiary'),
   fontSize: cssVar('fontBase'),
 });
+export const quickSearchBarEllipsisStyle = style({
+  maxWidth: '200px',
```
JimmFly:
```suggestion
```
Remove this

JimmFly: and add `overflow: 'hidden',` to `root` class
```
export const root = style({
...
  overflow: 'hidden',
});
```

SOUMITRO-SAHA: You’re right about the maxWidth, but after setting `overflow: 'hidden'` on the root, it now looks like this. Is it okay?

image

SOUMITRO-SAHA: You were absolutely right about the maxWidth. I realise now that it wasn’t necessary at all. I’m not sure what I was thinking at the time. Great code review, much appreciated @JimmFly

JimmFly (approved):

PR #8082, opened 2024-09-04
refactor(core): better block link

EYHN:


EYHN:
* **#8082** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8082?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/hooks/use-stable.ts
```diff
@@ -0,0 +1,9 @@
+import { useRef } from 'react';
+
+export const useReadOnce = <T>(fn: () => T): T => {
```
pengx17: I think this is the same with `const [v] = useState(fn); return v;`;

packages/frontend/core/src/hooks/use-stable.ts
```diff
@@ -0,0 +1,9 @@
...
+  const ref = useRef<T | null>(null);
+  if (ref.current === null) {
```
pengx17: what if fn() returns null

packages/frontend/core/src/components/affine/reference-link/index.tsx
```diff
@@ -128,21 +119,11 @@ export function AffinePageReference({
 
   const ref = useRef<HTMLAnchorElement>(null);
 
+  const [refreshKey, setRefreshKey] = useState<string>(() => nanoid());
```
forehalo, line 122: make a standalone hook?
```
const refresh = useForceRefresh()
```

packages/frontend/core/src/hooks/use-stable.ts
forehalo: excuse me? `const a = useReadOnce(fn)` equals `a = fn()`

forehalo: the same, `const a = useMemo(() => v.a, [v])`  equals `a = v.a`

forehalo: [confused](https://app.graphite.dev//settings/meme-library?org=toeverything)

pengx17: I think the use case of this hook is to make sure fn is only invoked once only even its ref is changed

packages/frontend/core/src/modules/editor/entities/editor.ts
```diff
@@ -36,6 +39,149 @@ export class Editor extends Entity<{
...
+
+    return () => {
+      unsubscribeEditorParams.unsubscribe();
+      unsubscribeViewParams.unsubscribe();
+    };
+  }
+
+  bindEditorContainer(
+    editorContainer: AffineEditorContainer,
+    docTitle: HTMLElement
```
pengx17: Use `DocTitle`?

packages/frontend/core/src/hooks/use-stable.ts
forehalo: the fn will always accept zero dependencies, which means it's pure return statement like `() => service.xxx`. it costs nothing

forehalo: whatever

packages/frontend/core/src/components/affine/reference-link/index.tsx
EYHN: <div class='graphite__hidden'>
          <div>🎥 Video uploaded on Graphite:</div>
            <a href="https://app.graphite.dev/media/video/g3jz87HxbjOJpXV3FPT7/503335bd-0766-48c9-bae5-4a1108dab272.mp4">
              
            </a>
          </div>
<video src="https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/g3jz87HxbjOJpXV3FPT7/503335bd-0766-48c9-bae5-4a1108dab272.mp4">CleanShot 2024-09-04 at 13.37.52.mp4</video>

This is to add a key to the url so that each click can trigger the focus action even if the blockid does not change.

packages/frontend/core/src/hooks/use-stable.ts
EYHN: never used... deleted

fundon (approved):

PR #8170, opened 2024-09-09
feat: improve ai query performance

darkskygit:


packages/backend/server/schema.prisma
```diff
@@ -457,6 +458,7 @@ model AiSession {
   prompt   AiPrompt           @relation(fields: [promptName], references: [name], onDelete: Cascade)
   messages AiSessionMessage[]
 
+  @@index([userId, workspaceId])
```
forehalo, line 462: index separately?

darkskygit: in most queries, we need to query both userId and workspaceId
there is only one place where you need to query userId alone, I add a separate index for userId

forehalo (approved):

PR #8176, opened 2024-09-09
fix(component): windows client does not have app controls on some pages

JimmFly:
close AF-1266

### DIFF
--- a/packages/frontend/component/src/components/affine-other-page-layout/index.css.ts
+++ b/packages/frontend/component/src/components/affine-other-page-layout/index.css.ts
@@ -27,6 +27,12 @@ export const topNav = style({
     },
   },
 });
+export const draggableHeader = style({
+  height: '52px',
+  width: '100%',
+  position: 'fixed',
+  ['WebkitAppRegion' as string]: 'drag',
+});
 export const topNavLinks = style({
   display: 'flex',
   columnGap: 4,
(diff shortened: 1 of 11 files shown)


JimmFly:
* **#8176** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8176?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

forehalo (commented):
we should find a better solution. it's not a good idea to add it for every page again and again

packages/frontend/component/src/components/auth-components/change-email-page.tsx
```diff
@@ -42,6 +44,7 @@ export const ChangeEmailPage = ({
     <AuthPageContainer
       title={t['com.affine.auth.change.email.page.title']()}
       subtitle={t['com.affine.auth.change.email.page.subtitle']()}
+      windowsAppControls={windowsAppControls}
```
EYHN: Will this affect the signup modal?

JimmFly: I don't think it will affect other pages. Basically, places where the `AffineOtherPageLayout` component can be used cannot be directly jumped in the client. It seems that you have to open the browser.

JimmFly: For example, this page
CleanShot 2024-09-10 at 11 05 15@2x

JimmFly: By the way, I adjusted the `404` page and `PageNotFound` components. I think it is more reasonable to call `AffineOtherPageLayout` only when jumping to the 404 page.

JimmFly:
> we should find a better solution. it's not a good idea to add it for every page again and again

I moved it to the electron app component, which should work as you expect. @forehalo

EYHN (approved):

PR #8335, opened 2024-09-20
fix(core): correctly ensure properties in PagePropertiesMetaManager

pengx17:
when using PagePropertiesMetaManager solely (eg., adding new properties in workspace settings), the page properties may not have been initialized.
added ensureRootProperties in PagePropertiesMetaManager only when mutating the properties explicitly

### DIFF
--- a/packages/frontend/core/src/components/affine/page-properties/page-properties-manager.ts
+++ b/packages/frontend/core/src/components/affine/page-properties/page-properties-manager.ts
@@ -88,6 +88,7 @@ export class PagePropertiesMetaManager {
     type: PagePropertyType;
     icon?: string;
   }) {
+    this.adapter.ensureRootProperties();
     const id = nanoid();
     const { type, icon } = schema;
     const newOrder =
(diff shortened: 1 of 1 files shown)


pengx17:
* **#8335** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8335?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

forehalo (commented):
nonono, it will cause date lost

pengx17:
> nonono, it will cause date lost

It only ensure these values when the user explicitly mutate the properties. I think it should be safe for the moment. Also the existing page info properties are considered deprecated right?

forehalo:
I prefer deleting all related `ensureXXX` code.  any properties should not be protectional initialized unless when docs or workspaces were created

pengx17:
ensureXXX was designed for old workspaces & docs that have not been set with these properties. Any other possible solutions?

PR #8347, opened 2024-09-23
fix(electron): cookie redirecting issue

pengx17:
fix AF-1416

### DIFF
--- a/packages/frontend/apps/electron/src/main/protocol.ts
+++ b/packages/frontend/apps/electron/src/main/protocol.ts
@@ -110,7 +110,8 @@ export function registerProtocol() {
     const protocol = url.protocol;
     const origin = url.origin;
 
-    const sameOrigin = origin === CLOUD_BASE_URL || protocol === 'file:';
+    const sameSite =
+      url.host === new URL(CLOUD_BASE_URL).host || protocol === 'file:';
 
     // offline whitelist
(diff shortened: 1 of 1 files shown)


pengx17:
* **#8347** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8347?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

EYHN (approved):

packages/frontend/apps/electron/src/main/protocol.ts
```diff
@@ -110,7 +110,8 @@ export function registerProtocol() {
     const protocol = url.protocol;
     const origin = url.origin;
 
-    const sameOrigin = origin === CLOUD_BASE_URL || protocol === 'file:';
+    const sameSite =
+      url.host === new URL(CLOUD_BASE_URL).host || protocol === 'file:';
```
forehalo, line 114: is it possible to migrate from listening 'file://' protocol to 'server_origin' and do the same file resolving like the file:// way, by which we won't need to manage cookies manually

pengx17: We can try this solution this iteration. The fix in this pr need to be released in next patch.

PR #8351, opened 2024-09-23
fix(core): handle composition event for Input component

JimmFly:
close AF-1065

### DIFF
--- a/packages/frontend/component/src/ui/input/input.tsx
+++ b/packages/frontend/component/src/ui/input/input.tsx
@@ -1,16 +1,14 @@
 import clsx from 'clsx';
 import type {
-  ChangeEvent,
   CSSProperties,
   ForwardedRef,
   InputHTMLAttributes,
-  KeyboardEvent,
   KeyboardEventHandler,
   ReactNode,
 } from 'react';
-import { forwardRef, useCallback, useEffect } from 'react';
+import { forwardRef } from 'react';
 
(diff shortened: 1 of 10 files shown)


JimmFly:
* **#8351** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8351?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/component/src/ui/input/input.tsx
```diff
@@ -26,6 +27,7 @@ export type InputProps = {
   endFix?: ReactNode;
   type?: HTMLInputElement['type'];
   inputStyle?: CSSProperties;
+  native?: boolean;
```
CatsJuice: Write some comments for this prop, I'm confused about why we need this?

JimmFly: Because I don't want to handle the Composition event for each` <input />`tag separately, and some input boxes cannot use the common style of our `<Input />` component in design, and require an `<Input native/>` component that can be adjusted more easily.

CatsJuice: Maybe we can extract logic from `<Input>` as a new component like `<RawInput />`, and use `<RawInput />` inside `<Input>`?

EYHN (approved):

PR #8369, opened 2024-09-24
feat(server): support onetime payment subscription

forehalo:


forehalo:
* **#8462** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8462?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8369** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8369?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/backend/server/src/plugins/payment/service.ts
```diff
@@ -46,20 +47,20 @@ const OnStripeEvent = (
...
-): [SubscriptionPlan, SubscriptionRecurring, SubscriptionPriceVariant?] {
+): [SubscriptionPlan, SubscriptionRecurring, SubscriptionVariant?] {
   const [plan, recurring, variant] = key.split('_');
 
   return [
     plan as SubscriptionPlan,
     recurring as SubscriptionRecurring,
-    variant as SubscriptionPriceVariant | undefined,
+    variant as SubscriptionVariant | undefined,
   ];
```
darkskygit, line 64: can we add a zod verify for this

packages/backend/server/src/plugins/payment/service.ts
```diff
@@ -607,6 +633,72 @@ export class SubscriptionService {
...
+        throw new InternalServerError(
+          'Unexpected onetime subscription with no end date'
+        );
+      }
+
+      const period =
+        // expired, reset the period
+        savedSubscription.end <= new Date()
+          ? {
+              start: new Date(),
```
darkskygit, line 666: should we extend the expiration date of the current subscription?

forehalo: what do you mean?

packages/backend/server/src/plugins/payment/service.ts
forehalo: it's kinda meaningless because the prices are consistent configured in stripe and the app will fail if the prices are setting wrongly. we can't recover from any unexpected inputs(from stripe), the only thing we can do is updating stripe configurations. 
user facing types are already protected by GraphQL

packages/backend/server/src/plugins/payment/service.ts
darkskygit: i misunderstood the condition, there's nothing wrong with the logic

darkskygit (approved):

PR #8387, opened 2024-09-25
feat(server): new login flow

darkskygit:
fix AF-1406

### DIFF
--- a/packages/backend/server/src/plugins/oauth/providers/google.ts
+++ b/packages/backend/server/src/plugins/oauth/providers/google.ts
@@ -33,7 +33,7 @@ export class GoogleOAuthProvider extends AutoRegisteredOAuthProvider {
   getAuthUrl(state: string) {
     return `https://accounts.google.com/o/oauth2/v2/auth?${this.url.stringify({
       client_id: this.config.clientId,
-      redirect_uri: this.url.link('/oauth/callback'),
+      redirect_uri: this.url.link('/api/oauth/redirect'),
       response_type: 'code',
       scope: 'openid email profile',
       prompt: 'select_account',
@@ -50,7 +50,7 @@ export class GoogleOAuthProvider extends AutoRegisteredOAuthProvider {
         code,
         client_id: this.config.clientId,
         client_secret: this.config.clientSecret,
-        redirect_uri: this.url.link('/oauth/callback'),
+        redirect_uri: this.url.link('/api/oauth/redirect'),
         grant_type: 'authorization_code',
       }),
       headers: {
--- a/packages/backend/server/src/plugins/oauth/providers/google.ts
+++ b/packages/backend/server/src/plugins/oauth/providers/google.ts
@@ -33,7 +33,7 @@ export class GoogleOAuthProvider extends AutoRegisteredOAuthProvider {
   getAuthUrl(state: string) {
     return `https://accounts.google.com/o/oauth2/v2/auth?${this.url.stringify({
       client_id: this.config.clientId,
-      redirect_uri: this.url.link('/api/oauth/redirect'),
+      redirect_uri: this.url.link('/oauth/callback'),
       response_type: 'code',
       scope: 'openid email profile',
       prompt: 'select_account',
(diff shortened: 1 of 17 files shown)


darkskygit:
* **#8387** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8387?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8387?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/toeverything?ref=merge-queue-instructions-in-stack-comment&prId=8387).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/backend/server/src/plugins/oauth/providers/google.ts
```diff
@@ -33,7 +33,7 @@ export class GoogleOAuthProvider extends AutoRegisteredOAuthProvider {
   getAuthUrl(state: string) {
     return `https://accounts.google.com/o/oauth2/v2/auth?${this.url.stringify({
       client_id: this.config.clientId,
-      redirect_uri: this.url.link('/oauth/callback'),
+      redirect_uri: this.url.link('/api/oauth/redirect'),
```
forehalo: changing redirect_uri requires update the config in providers, better keep it

packages/frontend/core/src/components/affine/auth/oauth.tsx
```diff
@@ -46,19 +45,25 @@ export function OAuth() {
...
   const { icon } = OAuthProviderMap[provider];
 
   const onClick = useCallback(() => {
-    let oauthUrl =
-      (BUILD_CONFIG.isElectron ? BUILD_CONFIG.serverUrlPrefix : '') +
-      `/oauth/login?provider=${provider}`;
-
-    if (BUILD_CONFIG.isElectron) {
-      oauthUrl += `&client=${appInfo?.schema}`;
+    async function preflight() {
```
forehalo, line 72: the logic can't be async otherwise safari will block popup

packages/backend/server/src/plugins/oauth/providers/google.ts
darkskygit: already update the provider config

packages/frontend/core/src/components/affine/auth/oauth.tsx
darkskygit: fixed

packages/backend/server/src/plugins/oauth/providers/google.ts
forehalo: there are selfhost users make the oauth work now. it's not easy to notify them updating the config as well. keep the old routes.

and if it's user facing route, it should not start with `/api`

darkskygit: we have not handle `/oauth` route in ingress, and i not prefer redirect page at frontend
server side redirect can avoid most attacks supported by browser extensions, and will not cause the login process to be terminated due to front-end errors
and the current front-end routing needs to load most of the JS before it can execute the redirection logic, which is too slow

darkskygit: image
according to statistics, there are less than 20 event using unofficial OAuth login sources
i believe the current unofficial oauth user is very small

packages/frontend/core/src/modules/cloud/services/auth.ts
```diff
@@ -190,6 +210,26 @@ export class AuthService extends Service {
     }
   }
 
+  async signInOauthToken(token: string, provider: string) {
+    try {
+      const res = await this.fetchService.fetch('/api/oauth/callback', {
+        method: 'POST',
+        body: JSON.stringify({ token, state: this.state }),
```
forehalo: I start doubting whether it's really necessary
https://github.com/desktop/desktop/blob/8bc4ec835573972cad14c13fe1a11d5dae22ce61/app/src/lib/stores/sign-in-store.ts#L550

forehalo: the desktop authentication works more like device authentication
https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps#device-flow

darkskygit: https://github.com/desktop/desktop/blob/development/app/src/lib/api.ts#L2185

in-app secret also exists, the difference is that our secret is requested before logging in, but github's is preset

forehalo:
I did't get time review on this PR. will back on it soon

fengmk2 (commented):
move to #10648

PR #8456, opened 2024-10-08
refactor(i18n): lazy load languages

forehalo:
closes AF-1397


forehalo:
* **#8456** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8456?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/modules/i18n/context.ts
```diff
@@ -0,0 +1,14 @@
+import { useService } from '@toeverything/infra';
+import { type PropsWithChildren, useEffect } from 'react';
+
+import { I18nService } from './services/i18n';
+
+export function I18nProvider({ children }: PropsWithChildren) {
+  const i18n = useService(I18nService).i18n;
+
+  useEffect(() => {
+    i18n.init();
```
EYHN: Should we wait for i18n to be ready before rendering children? there might be flickering from en to other languages?

forehalo: en better than nothing

packages/frontend/i18n/src/index.ts
```diff
@@ -35,88 +33,87 @@ declare module 'react-i18next' {
...
+        read: (lng: Language, _ns: string, callback) => {
+          const resource = SUPPORTED_LANGUAGES[lng].resource;
+          if (typeof resource === 'function') {
+            resource()
+              .then(data => {
+                logger.info(`Loaded i18n ${lng} resource`);
+                callback(null, data);
+              })
+              .catch(err => {
+                logger.error(`Failed to load i18n ${lng} resource`, err);
```
pengx17: fallback to `defaultLng` when load error?

forehalo: fallback setting at L67, 'en' is always loaded

pengx17: i think `fallbackLng` at `L67` is for individual missing trans keys for that locale.

forehalo: the resolver will always traverse all fallbacks to find a key. when error happens, i18next will fill that resources with empty object

```
langs = ['zh-Hans', 'en']

foreach lang in langs
  if lang.resolve(key) 
    break
```

pengx17 (approved):

PR #8088, opened 2024-09-04
feat: init renderer server

Brooooooklyn:

PR #8460, opened 2024-10-09
fix(server): blank screen on mobile

### LINKED ISSUE #8338: ​Blank screen on mobile when sharing
### What happened?

On mobile browsers, when attempting to open shared notes, a blank screen is displayed.

**Steps to Reproduce:**
1. Open Chrome browser on a mobile device.
2. Go to [this link](https://docs.polyakov.marketing/workspace/8f799969-8391-40eb-a85e-722cd6146c06/2wXXc3IsiRkYXavczZcUA?mode=page).  https://docs.polyakov.marketing/workspace/8f799969-8391-40eb-a85e-722cd6146c06/2wXXc3IsiRkYXavczZcUA?mode=page
3. Observe the blank screen.

**Additional Information:**
- The issue only occurs in notes created in AFFiNE version 0.17. Previous stable versions work correctly.

logs in console:
CleanShot 2024-09-20 at 18 26 57@2x

large screenshot from iPhone
AFFiNE

on desktops:
CleanShot 2024-09-20 at 18 17 53@2x

### Distribution version

Web (app.affine.pro)

Chrome

### Are you self-hosting?

- [X] Yes

### Relevant log output

_No response_

### Anything else?

_No response_

bytemain:
close: https://github.com/toeverything/AFFiNE/issues/8338

this bug is caused by https://github.com/toeverything/AFFiNE/pull/8088#discussion_r1792087140

which did not serve the mobile static resource path

Fixed：
image

image


packages/backend/server/src/core/selfhost/index.ts
```diff
@@ -74,13 +70,25 @@ export class SelfhostModule implements OnModuleInit {
...
+    // mobile => 'static/mobile/selfhost'
     app.use(
       basePath + '/admin',
       serveStatic(join(staticPath, 'admin', 'selfhost'), {
         redirect: false,
         index: false,
       })
     );
+    app.use(
+      basePath + '/mobile',
```
forehalo: there is no /mobile route

bytemain: it was introduced by this pr, DocRenderer will render this path if user agent is mobile

packages/backend/server/tests/app/selfhost.e2e.ts
```diff
@@ -101,6 +122,67 @@ test('should always return static asset files', async t => {
...
+
+  t.true(res.text.includes('AFFiNE Mobile'));
+
+  res = await request(t.context.app.getHttpServer())
+    .get(
+      '/workspace/3aa2e665-7b0d-41c6-9979-db17c8d93836/Sabgfj_trVBY6_iM_6uaO'
+    )
+    .expect(200);
+
+  t.false(res.text.includes('AFFiNE Mobile'));
```
forehalo: app tests should only contain basic healthy tests
I will add doc renderer tests separately later

tools/cli/src/webpack/config.ts
```diff
@@ -76,6 +76,10 @@ export const getPublicPath = (buildFlags: BuildFlags) => {
     return publicPath;
   }
 
+  if (buildFlags.distribution === 'mobile') {
+    return '/mobile/';
```
forehalo: this will make affine itself fail. it should be joint with CDN prefix

packages/backend/server/tests/app/selfhost.e2e.ts
bytemain: All right, what can I do for this? And this test case is working on my laptop, but failed in github actions, I will see it when I have time

tools/cli/src/webpack/config.ts
bytemain: Reasonable, I will change here.

packages/backend/server/tests/app/selfhost.e2e.ts
forehalo: I will append a commit to fix

packages/backend/server/src/core/doc-renderer/controller.ts
```diff
@@ -214,11 +222,24 @@ export class DocRendererController {
...
+      const publicPath = this.config.isSelfhosted
+        ? this.config.server.host + assets.selfhostPublicPath
+        : assets.publicPath;
+
+      if (this.config.isSelfhosted) {
+        // append `config.server.host` to public path
+        assets.selfhostPublicPath = publicPath;
+      }
+
+      assets.js = assets.js.map(path => publicPath + path);
```
bytemain, line 236: if `publicPath` is `/`, and the path is `/js/main.js`, the result is `//js/main.js`, this is not expected, seems should remove the prefix `/` of path

forehalo (approved):

PR #8461, opened 2024-10-09
fix(core): unexpected double-click behavior when selecting multiple options on all doc pages

JimmFly:
close AF-968

### DIFF
--- a/packages/frontend/core/src/components/page-list/list.tsx
+++ b/packages/frontend/core/src/components/page-list/list.tsx
@@ -61,7 +61,9 @@ const useItemSelectionStateEffect = () => {
         if (
           target.tagName === 'BUTTON' ||
           target.tagName === 'INPUT' ||
-          (e.target as HTMLElement).closest('button, input, [role="toolbar"]')
+          (e.target as HTMLElement).closest(
+            'button, input, [role="toolbar"], [role="list-item"]'
+          )
         ) {
           return;
         }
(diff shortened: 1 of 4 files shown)


JimmFly:
* **#8461** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8461?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/components/page-list/list.tsx
```diff
@@ -61,7 +61,9 @@ const useItemSelectionStateEffect = () => {
         if (
           target.tagName === 'BUTTON' ||
           target.tagName === 'INPUT' ||
-          (e.target as HTMLElement).closest('button, input, [role="toolbar"]')
+          (e.target as HTMLElement).closest(
+            'button, input, [role="toolbar"], [data-all-doc-list="true"]'
```
pengx17: I think maybe you can reuse `[data-testid=page-list-item]`

pengx17 (approved):

packages/frontend/core/src/components/page-list/list.tsx
JimmFly: I remember that data-test-id seemed to be removed in the production environment.

JimmFly: I think I remembered it wrongly.

pengx17: Good point. Or perhaps adding button role to the item list and then filter the event based on roles

EYHN (approved):

PR #8477, opened 2024-10-11
fix(core): prevent floating sidebar from disappearing unexpectedly

JimmFly:
close AF-1475

Fixed an issue where you couldn't create collection or use the rename function in the floating sidebar.

https://github.com/user-attachments/assets/41c2b6a8-8fc9-4f8b-ab51-bd7ce2a58738

### DIFF
--- a/packages/frontend/core/src/modules/app-sidebar/views/index.tsx
+++ b/packages/frontend/core/src/modules/app-sidebar/views/index.tsx
@@ -124,13 +124,28 @@ export function AppSidebar({ children }: PropsWithChildren) {
     appSidebarService.setOpen(false);
   }, [appSidebarService]);
 
-  const onMouseEnter = useCallback(() => {
-    appSidebarService.setHovering(true);
-  }, [appSidebarService]);
+  useEffect(() => {
+    if (sidebarState !== 'floating' || resizing) {
+      return;
+    }
+    const onMouseMove = (e: MouseEvent) => {
+      const menuElement = document.querySelector(
+        'body > [data-radix-popper-content-wrapper] > [data-radix-menu-content]'
+      );
 
-  const onMouseLeave = useCallback(() => {
-    appSidebarService.setHovering(false);
-  }, [appSidebarService]);
+      if (menuElement) {
+        return;
+      }
+
+      if (e.clientX > width + 20) {
+        appSidebarService.setHovering(false);
+      }
+    };
+    document.addEventListener('mousemove', onMouseMove);
+    return () => {
+      document.removeEventListener('mousemove', onMouseMove);
+    };
+  }, [appSidebarService, resizing, sidebarState, width]);
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


JimmFly:
* **#8477** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8477?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

pengx17 (commented):
Thoughts: Normally for react portals that are in the sub tree of appsidebar, there `mouseleave` handler should not trigger moving from the sidebar to nested portals, e.g. the operation menus. However some of our implementations of *modals* are either using `useMount` to mount the modal outside of the normal react tree, or being rendered somewhere globally.

Determining whether the mouseleave event should be handled without a defined React tree is tricky. In my opinion, you should add condition checks in `onMouseLeave`, rather than in the useEffect hook.

A working example
```ts
  const onMouseLeave = useCallback(
    (e: React.MouseEvent) => {
      const target = e.relatedTarget as HTMLElement;
      if (
        target?.closest(
          '[role="dialog"], [role="menu"], modal-transition-container'
        )
      ) {
        return;
      }
      appSidebarService.setHovering(false);
    },
    [appSidebarService]
  );
```

JimmFly:
> Thoughts: Normally for react portals that are in the sub tree of appsidebar, there `mouseleave` handler should not trigger moving from the sidebar to nested portals, e.g. the operation menus. However some of our implementations of _modals_ are either using `useMount` to mount the modal outside of the normal react tree, or being rendered somewhere globally.
> 
> Determining whether the mouseleave event should be handled without a defined React tree is tricky. In my opinion, you should add condition checks in `onMouseLeave`, rather than in the useEffect hook.


It seems that it doesn't work when I want to use the rename function. @pengx17 

CleanShot 2024-10-11 at 14 56 39@2x

pengx17:
add the following check should work `!(target instanceof HTMLElement) || resizing ||`
i think this is because radix tries to blur the element and as a result the window object got focused

pengx17:
I also noticed that resizing sidebar becomes slow. Can you take a look?

JimmFly:
> I also noticed that resizing sidebar becomes slow. Can you take a look?

Does it happen on pages other than edgeless page?

pengx17:
happens everywhere

pengx17 (approved):

packages/frontend/core/src/modules/app-sidebar/views/index.tsx
```diff
@@ -128,9 +128,33 @@ export function AppSidebar({ children }: PropsWithChildren) {
...
+      ) {
+        return;
+      }
+
+      // handle uncaught popper
+      const popper = document.querySelector(
+        '[data-radix-popper-content-wrapper]'
+      );
+      if (popper) {
+        return;
```
EYHN: It seems that in this case the sidebar never closes, even if the dialog has been closed.

EYHN: I think we can try, when the sidebar is floating, we listen to the mousemove event on the window, and close it when the mouse moves a certain distance out of the sidebar range

PR #8484, opened 2024-10-12
Added responsiveness to the settings modal for mobile devices

WhiteSnek:
### Description

This PR makes the settings modal responsive for mobile devices. It adjusts the layout, font sizes, and spacing to improve usability on smaller screens.

### Changes

- Added responsive styles to ensure the modal works well on mobile devices.
- Updated layout and element sizes for better touch interaction and readability.

### Before:

Screenshot 2024-10-12 at 21-25-02 AFFiNE

### After:

Screenshot 2024-10-12 at 21-51-26 AFFiNE


packages/frontend/core/src/components/affine/setting-modal/setting-sidebar/index.tsx
```diff
@@ -145,14 +147,20 @@ export const SettingSidebar = ({
     [onTabChange]
   );
 
+  const isMobile = window.innerWidth <= 768;
```
EYHN: How about put it in css, and using media query + `display: none` ?

mrdjohnson:
Random guy here 👋 

- the value for the display language appears to be missing in your "after", maybe its pushed too far to the left?
- not a designer but if possible it would be nice if the modal was full screened (or at least smaller margins) for mobile devices no?

packages/frontend/core/src/components/affine/setting-modal/setting-sidebar/index.tsx
WhiteSnek: Apologies for the delayed response. Thank you for the suggestion! I'll update the implementation accordingly.

WhiteSnek:
> Random guy here 👋
> 
>     * the value for the display language appears to be missing in your "after", maybe its pushed too far to the left?
> 
>     * not a designer but if possible it would be nice if the modal was full screened (or at least smaller margins) for mobile devices no?

you're right. i'll fix it

WhiteSnek:
@EYHN @mrdjohnson i've implemented the changes
Screenshot 2024-12-01 154522

PR #8487, opened 2024-10-13
feat(admin): allow config unlimited copilot

bytemain:
If you configure copilot in your self host instance, it still has a limit: 10 quotas.

### DIFF
--- a/packages/backend/server/src/core/features/resolver.ts
+++ b/packages/backend/server/src/core/features/resolver.ts
@@ -37,7 +37,7 @@ export class AvailableUserFeatureConfig {
 
   async availableUserFeatures() {
     return this.config.isSelfhosted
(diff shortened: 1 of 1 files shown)


Brooooooklyn (changes requested):
I think it should be configurable via the admin panel. Some users don’t want everyone on their self-hosted server to use Copilot without limitations.

bytemain:
> I think it should be configurable via the admin panel. Some users don’t want everyone on their self-hosted server to use Copilot without limitations.

Yes, this field `availableUserFeatures` define what features users can operate in the admin panel, if here we do not add `UnlimitedCopilot`, user can only control `admin` feature in the admin/config panel.

Brooooooklyn (approved):

PR #8494, opened 2024-10-14
chore(oxlint): configure rule categories in oxlint.json

DonIsaac:
## What This PR Does

Moves category-based configuration from CLI args in `lint:ox` to `oxlint.json`.

We are adding support for the `plugins` field in the next release, and we ironed out how CLI-based rule configuration merges with file-based rule configuration (see: https://github.com/oxc-project/oxc/pull/6088). Rules that should have been enabled before will now be enabled, and due to this I also fixed a violation of `eslint/no-unsafe-optional-chaining` that was not caught before.

This is part of an ongoing effort to keep our ecosystem CI nice and green. Thank you for using oxlint ⚓


pengx17 (approved):
Thanks!

PR #8509, opened 2024-10-16
feat(core): add global error handler

forehalo:
- register error handler at line 1
- no dependency

close AF-1473


forehalo:
* **#8509** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8509?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/bootstrap/global-error-handler.ts
```diff
@@ -0,0 +1,46 @@
+import errorImg from './error.png';
+
+function handleGlobalUnrecoverableError(e: ErrorEvent | PromiseRejectionEvent) {
```
pengx17: I am not sure - some unhandled errors/rejections may not be that disastrous and make the app unusable? The error prompting may be misleading because it may only happend during bootstrap phase some point after app is rendered.

forehalo: errors in renderer should be caught by react error boundary

pengx17: I tried by throwing an error in a service contructor. It seems this is captured in the new handler, instead of in the react error boundary

packages/frontend/core/src/bootstrap/global-error-handler.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (typeof document !== 'undefined') {
+    globalThis.addEventListener(
+      'unhandledrejection',
+      handleGlobalUnrecoverableError
+    );
+    globalThis.addEventListener('error', handleGlobalUnrecoverableError);
+  }
+}
+
+registerGlobalErrorHandler();
```
EYHN, line 80: If the browser version is too low, there is unparseable syntax in the js file, and I suspect this will not be run at all.

should we bundle the error handling into a separate js file ?

forehalo: which one

packages/frontend/core/src/bootstrap/global-error-handler.ts
forehalo: this is bad, error boundary won't stop propagation of errors by default

packages/frontend/core/src/bootstrap/global-error-handler.ts
pengx17: This file (global-error-handler.ts) will be bundled with other js files. I think the most reasonable way is to have an explicit entry file?

forehalo: it won't help but more complex. I will split this into a separated chunk to make sure it will always be loaded first

packages/frontend/core/src/bootstrap/polyfill/browser.ts
```diff
@@ -4,14 +4,8 @@ import { polyfillPromise } from './promise-with-resolvers';
...
-  polyfillEventLoop();
-  await polyfillI18n();
-  await polyfillPromise();
-  await polyfillDispose();
-}
+polyfillResizeObserver();
+polyfillEventLoop();
+await polyfillI18n();
+await polyfillPromise();
+await polyfillDispose();
```
pengx17: I recalled that top-level await caused some issues before but I forget the exact reasons

tools/cli/src/webpack/config.ts
```diff
@@ -404,6 +398,11 @@ export const createConfiguration: (
         maxInitialRequests: Infinity,
         chunks: 'all',
         cacheGroups: {
+          errorHandler: {
+            test: /global-error-handler/,
```
pengx17, line 402: IMO explicit entry is better for maintainance. affine devs often do refactors that the rule here might be missed in the future

forehalo: we need take care more if choosing multi entries

forehalo: you can try doing it if it helps

packages/frontend/core/src/bootstrap/polyfill/browser.ts
forehalo: top level await comes for a long time in apps/setup

pengx17 (approved):

packages/frontend/core/src/bootstrap/setup.ts
```diff
@@ -0,0 +1,3 @@
+import './global-error-handler';
```
pengx17: I think this may be unnecessary for electron

PR #8517, opened 2024-10-16
fix(core): journal nav button icon

forehalo:
close AF-1494

### DIFF
--- a/packages/frontend/core/src/components/root-app-sidebar/journal-button.tsx
+++ b/packages/frontend/core/src/components/root-app-sidebar/journal-button.tsx
@@ -25,7 +25,8 @@ export const AppSidebarJournalButton = ({
   const workbench = useService(WorkbenchService).workbench;
   const location = useLiveData(workbench.location$);
   const { openToday } = useJournalRouteHelper(docCollection);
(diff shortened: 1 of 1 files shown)


forehalo:
* **#8517** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8517?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/components/root-app-sidebar/journal-button.tsx
```diff
@@ -34,12 +34,14 @@ export const AppSidebarJournalButton = ({
...
   );
 
-  const JournalIcon = useLiveData(
-    docDisplayMetaService.icon$(docCollection.id, {
-      compareDate: new Date(),
-    })
-  );
-  const Icon = isJournal ? JournalIcon : TodayIcon;
+  const Icon =
+    isJournal && journalDate
```
CatsJuice: This logic should included in `docDisplayMetaService.icon$`, the issue here should be wrong parameter. Should replace `docCollection.id` with `location.pathname.split('/')[1]`(docId)

CatsJuice (approved):

PR #8525, opened 2024-10-17
feat(core): new "is journal" page property

CatsJuice:
close AF-1450, AF-1451, AF-14552


CatsJuice:
* **#8525** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8525?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/modules/journal/store/journal.ts
```diff
@@ -11,6 +11,17 @@ export class JournalStore extends Store {
...
   }
 
+  allJournalDates$ = LiveData.computed(get => {
+    return new Set(
+      get(this.docsService.list.docs$)
+        .filter(doc => {
+          const journal = get(doc.properties$.selector(p => p.journal));
+          return !!journal && isJournalString(journal);
+        })
+        .map(doc => doc.properties$.value.journal)
```
pengx17: get(doc.properties$).journal?

CatsJuice: Not really sure, I think that should already be observed inside `filter`. It seems to heavy here.

packages/frontend/core/src/desktop/pages/workspace/detail-page/tabs/journal.tsx
```diff
@@ -297,6 +322,12 @@ const ConflictList = ({
     },
     [setTrashModal]
   );
+  const handleRemoveJournalMark = useCallback(
+    (docId: string) => {
+      journalService.removeJournalDate(docId);
```
pengx17, line 329: worth adding a telemetry event

pengx17 (approved):
I think we need to add tests for the duplicated journal logic. However how to add duplicated journal is a question ... Perhaps we can have a backdook that exposes journal service for testing?

CatsJuice:
> I think we need to add tests for the duplicated journal logic. However how to add duplicated journal is a question ... Perhaps we can have a backdook that exposes journal service for testing?

Yes, I am adding some basic test cases. Now we can turn regular page into journal and specify specific dates, so conflicts can be easily triggered.

packages/frontend/core/src/modules/journal/store/journal.ts
pengx17: I think a well designed `computed` should be able to batch the updates

EYHN (approved):

PR #8531, opened 2024-10-17
feat(core): prevent the floating sidebar from showing immediately after the sidebar is closed

JimmFly:
close PD-1763

https://github.com/user-attachments/assets/a8346098-5f87-4297-bdb6-885f5486ce78

### DIFF
--- a/packages/frontend/core/src/modules/app-sidebar/views/sidebar-header/sidebar-switch.tsx
+++ b/packages/frontend/core/src/modules/app-sidebar/views/sidebar-header/sidebar-switch.tsx
@@ -16,15 +16,30 @@ export const SidebarSwitch = ({
 }) => {
   const appSidebarService = useService(AppSidebarService).sidebar;
   const open = useLiveData(appSidebarService.open$);
+  const preventHovering = useLiveData(appSidebarService.preventHovering$);
+  const timeoutRef = useRef<NodeJS.Timeout | null>(null);
   const switchRef = useRef<HTMLDivElement>(null);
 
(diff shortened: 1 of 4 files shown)


JimmFly:
* **#8531** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8531?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

pengx17 (commented):
I think you can add a timer in `AppSidebar` entity to prevent hovering = true when sidebar is just closed

JimmFly:
> I think you can add a timer in `AppSidebar` entity to prevent hovering = true when sidebar is just closed

Maybe it would be better to use the timer as little as possible?

packages/frontend/core/src/modules/app-sidebar/views/sidebar-header/sidebar-switch.tsx
```diff
@@ -16,15 +16,23 @@ export const SidebarSwitch = ({
 }) => {
   const appSidebarService = useService(AppSidebarService).sidebar;
   const open = useLiveData(appSidebarService.open$);
+  const preventHovering = useLiveData(appSidebarService.preventHovering$);
   const switchRef = useRef<HTMLDivElement>(null);
 
   const handleMouseEnter = useCallback(() => {
+    if (preventHovering) {
```
EYHN: You can start listening for the mouseenter event after the sidebar is collapsed, which should achieve the same effect without adding additional states.

EYHN (approved):

PR #8554, opened 2024-10-20
fix: download button shows on refresh

### LINKED ISSUE #8529: ​Download App tip always show when reload page
### What happened?

CleanShot 2024-10-17 at 13 22 36@2x

I think it should stay closed when I actively click the close button.

Now when I click the close icon, and reload page, it will show again.

Maybe we can store the close state to `localStorage`

### Distribution version

Web (app.affine.pro)

_No response_

### Are you self-hosting?

- [X] Yes

### Relevant log output

_No response_

### Anything else?

- https://github.com/bytemain/AFFiNE/blob/7cee5d591106695b49828bed206bee4d9b6d5230/packages/frontend/core/src/modules/app-sidebar/views/app-download-button/index.tsx#L17

the code just use a in memory state

abhipatel80:
fix #8529

### DIFF
--- a/packages/frontend/core/src/modules/app-sidebar/views/app-download-button/index.tsx
+++ b/packages/frontend/core/src/modules/app-sidebar/views/app-download-button/index.tsx
@@ -2,7 +2,7 @@ import { useCatchEventCallback } from '@affine/core/components/hooks/use-catch-e
 import { track } from '@affine/track';
 import { CloseIcon, DownloadIcon } from '@blocksuite/icons/rc';
 import clsx from 'clsx';
-import { useCallback, useState } from 'react';
+import { useCallback, useState, useEffect } from 'react';
 
 import * as styles from './index.css';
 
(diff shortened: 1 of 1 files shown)


packages/frontend/core/src/modules/app-sidebar/views/app-download-button/index.tsx
```diff
@@ -14,10 +14,16 @@ export function AppDownloadButton({
...
 }) {
-  const [show, setShow] = useState(true);
+  const [show, setShow] = useState(null);
+
+  useEffect(() => {
+    const isClosed = localStorage.getItem('appDownloadClosed') === 'true';
+    setShow(!isClosed);
+  }, []);
 
   const handleClose = useCatchEventCallback(() => {
```
pengx17, line 24: @hwangdev97 do we still want to always show the download desktop app button on web?
also the implementation here may follow dismissChangelog in `useAppUpdater`

pengx17:
download button is superseded by https://github.com/toeverything/AFFiNE/pull/8554

PR #8581, opened 2024-10-23
feat(mobile): ios oauth & magic-link login

CatsJuice:


CatsJuice:
* **#8581** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8581?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/modules/cloud/services/auth.ts
```diff
@@ -116,13 +116,18 @@ export class AuthService extends Service {
   ) {
     track.$.$.auth.signIn({ method: 'magic-link' });
     try {
+      const scheme = BUILD_CONFIG.isElectron
+        ? appInfo?.schema
+        : BUILD_CONFIG.isIOS || BUILD_CONFIG.isAndroid
+          ? 'affine'
+          : 'web';
```
pengx17: Are we planning to submit different app versions for canary/beta/stable?

CatsJuice: Scheme in electron is dynamic for each channel, but it's fixed for iOS. Not sure about android yet.

packages/frontend/core/package.json
```diff
@@ -18,6 +18,8 @@
     "@affine/track": "workspace:*",
     "@blocksuite/affine": "0.17.19",
     "@blocksuite/icons": "2.1.69",
+    "@capacitor/app": "^6.0.1",
```
EYHN, line 21: I think we should avoid referencing the capacitor package in core

packages/frontend/core/src/components/affine/auth/send-email.tsx
```diff
@@ -122,7 +122,11 @@ const useSendEmail = (emailType: AuthPanelProps<'sendEmail'>['emailType']) => {
         return trigger({
           email,
           callbackUrl: `/auth/${callbackUrl}?isClient=${
```
EYHN: AppSchema is not needed here?

packages/frontend/core/src/utils/popup.ts
```diff
@@ -1,5 +1,6 @@
 import { DebugLogger } from '@affine/debug';
 import { apis } from '@affine/electron-api';
+import { Browser } from '@capacitor/browser';
 
 const logger = new DebugLogger('popup');
```
EYHN: <i class='graphite__hidden'>[Re: line +15]<br/><br/></i>

We need to abstract this into a popupWindowProvider<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8581?utm_source=unchanged-line-comment">Graphite</a>.</span>

packages/frontend/core/src/components/affine/auth/send-email.tsx
CatsJuice: Seems `client` is not used in `/auth/:authType`

packages/frontend/core/package.json
CatsJuice: but `popupWindow` is implemented in core. Should we wrap a function in `@affine/ios` and import from it?

packages/frontend/core/src/modules/url/services/url.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { ClientSchemaProvider } from '../providers/client-schema';
+import type { PopupWindowProvider } from '../providers/popup-window';
+
+export class UrlService extends Service {
+  // those providers are optional, because they are not always available in some environments
+
+  private readonly popupWindowProvider: PopupWindowProvider | undefined;
+  private readonly clientSchemaProvider: ClientSchemaProvider | undefined;
+  constructor(
+    popupWindowProvider: PopupWindowProvider[],
```
pengx17: why define popupWindowProvider as an array but only the first value is used?

pengx17 (approved):

PR #8608, opened 2024-10-27
feat(server): verify domain record

darkskygit:


packages/backend/server/src/core/auth/controller.ts
```diff
@@ -147,6 +152,33 @@ export class AuthController {
       if (!allowSignup) {
         throw new SignUpForbidden();
       }
+
+      const requireEmailDomainVerification = await this.config.runtime.fetch(
```
EYHN, line 162: Will this cause registered users to be unable to log into their accounts?
We should only check this when registering

darkskygit: this logic only execute that user not register yet

packages/backend/server/src/core/auth/controller.ts
```diff
@@ -30,6 +32,9 @@ import { AuthService } from './service';
 import { CurrentUser, Session } from './session';
 import { TokenService, TokenType } from './token';
 
+// set DNS servers
+setServers(['1.1.1.1', '8.8.8.8']);
```
Brooooooklyn: why we need this? Our Node.js instances are in k8s cluster, there may rpc requires `coredns` in the future, can we remove this?

darkskygit: in the local dev env, if we enable network software like clash tun/surge/tailscale/zerotier, thire DNS hijacking logic may not correctly provide domain records other than DNS. let me add a conditional logic

darkskygit: fixed

Brooooooklyn (approved):

PR #8637, opened 2024-10-30
feat(core): open app in electron app entry

pengx17:
fix PD-208
fix PD-210
fix PD-209
fix AF-1495


pengx17:
* **#8637** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8637?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/frontend/core/src/modules/open-in-app/services/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.initialUrl = window.location.href;
+
+    const maybeDocLink = resolveLinkToDoc(this.initialUrl);
+    let shouldOpenInApp = false;
+
+    if (
+      maybeDocLink &&
+      this.workspacesService.list.workspaces$.value.find(
+        w => w.id === maybeDocLink.workspaceId
+      )?.flavour === WorkspaceFlavour.AFFINE_CLOUD
```
EYHN: Maybe the user is not logged in at this time?
We just need to check if the workspace is not a local workspace.

pengx17: I see, this list only makes sense if the user is logged in? I will read the value from localstorage instead.

packages/frontend/core/src/modules/open-in-app/views/open-in-app-page.tsx
```diff
@@ -0,0 +1,160 @@
...
+        <img src={appIcon} alt={appName} width={120} height={120} />
+
+        <div className={styles.prompt}>
+          <Trans i18nKey="com.affine.auth.open.affine.prompt">
+            Open {appName} app now
+          </Trans>
+        </div>
+
+        <div className={styles.promptLinks}>
+          {maybeDocLink && (
```
EYHN:
```suggestion
          {openHereClicked && (
```

packages/frontend/core/src/modules/open-in-app/services/index.ts
EYHN:
```suggestion
      !this.workspacesService.list.workspaces$.value.some(
        w => w.id === maybeDocLink.workspaceId && w.flavour === WorkspaceFlavour.LOCAL
      )
```

pengx17: will use `getLocalWorkspaceIds` instead

EYHN (approved):

PR #8666, opened 2024-11-01
fix: old document edgelessColorTheme shows auto but always render white

akumatus:
Fix issue [AF-1593](https://linear.app/affine-design/issue/AF-1593).

### DIFF
--- a/packages/frontend/core/src/components/blocksuite/block-suite-editor/specs/custom/root-block.ts
+++ b/packages/frontend/core/src/components/blocksuite/block-suite-editor/specs/custom/root-block.ts
@@ -107,7 +107,9 @@ function createThemeExtension(framework: FrameworkProvider) {
       if (cache) return cache;
 
       const appTheme$ = framework.get(AppThemeService).appTheme.theme$;
-      const docTheme$ = doc.properties$.map(props => props.edgelessColorTheme);
+      const docTheme$ = doc.properties$.map(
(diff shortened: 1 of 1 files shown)


akumatus:
* **#8666** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8666?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

pengx17 (approved):

packages/frontend/core/src/components/blocksuite/block-suite-editor/specs/custom/root-block.ts
```diff
@@ -107,7 +107,9 @@ function createThemeExtension(framework: FrameworkProvider) {
       if (cache) return cache;
 
       const appTheme$ = framework.get(AppThemeService).appTheme.theme$;
-      const docTheme$ = doc.properties$.map(props => props.edgelessColorTheme);
+      const docTheme$ = doc.properties$.map(
+        props => props.edgelessColorTheme || 'system'
```
pengx17, line 111: fallback to appTheme$ instead?

akumatus: This default value logic is also used in UI components, making the use of `edgelessColorTheme || 'system'` more consistent here.

PR #8729, opened 2024-11-07
fix(core): setting height issue

pengx17:
fix AF-1642

PR #8730, opened 2024-11-07
feat(electron): spellcheck setting

pengx17:
fix AF-1647

### DIFF
--- a/packages/frontend/core/src/desktop/dialogs/setting/general-setting/editor/general.tsx
+++ b/packages/frontend/core/src/desktop/dialogs/setting/general-setting/editor/general.tsx
@@ -15,22 +15,26 @@ import {
   SettingRow,
   SettingWrapper,
 } from '@affine/component/setting-components';
+import { useAsyncCallback } from '@affine/core/components/hooks/affine-async-hooks';
 import { ServerConfigService } from '@affine/core/modules/cloud';
+import { DesktopApiService } from '@affine/core/modules/desktop-api';
 import {
   EditorSettingService,
   type FontFamily,
   fontStyleOptions,
 } from '@affine/core/modules/editor-setting';
+import { SpellCheckSettingService } from '@affine/core/modules/editor-setting/services/spell-check-setting';
 import {
   type FontData,
   SystemFontFamilyService,
 } from '@affine/core/modules/system-font-family';
-import { useI18n } from '@affine/i18n';
+import { Trans, useI18n } from '@affine/i18n';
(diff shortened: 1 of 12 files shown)


pengx17:
* **#8730** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8730?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#8728** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8728?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#8729](https://github.com/toeverything/AFFiNE/pull/8729) <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8729?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `canary`

<h2></h2>

packages/frontend/core/src/modules/editor-setting/impls/user-db.ts
```diff
@@ -25,9 +25,8 @@ export class CurrentUserDBEditorSettingProvider
         key,
         value,
       });
-    } else {
-      this.fallback.set(key, value);
     }
+    this.fallback.set(key, value);
```
pengx17: Electron need to read the editor setting on startup. There is no easy way in the electron side to know where to read the editor setting value (global state json or from sqlite db?)
This is a workaround to make sure even if user db is enabled, the settings will also be written to global state json, which should be available when electron is starting.

packages/frontend/core/src/desktop/dialogs/setting/general-setting/editor/general.tsx
```diff
@@ -351,7 +355,7 @@ const NewDocDefaultModeSettings = () => {
   );
 };
 
-export const DeFaultCodeBlockSettings = () => {
+export const DefaultCodeBlockSettings = () => {
```
forehalo: ?

forehalo: I thinks this can be removed.
should be put in 'Editor' settings if new requirement comes

forehalo (approved):

packages/frontend/core/src/modules/editor-setting/impls/user-db.ts
EYHN: Since we can't enable and disable spell-check at runtime, how about move it out of the editor setting?

PR #8734, opened 2024-11-07
feat(infra): introduce op pattern

forehalo:
### TL;DR

Introduces a new Operation Pattern RPC framework for cross-tabs, worker, or even client-server communication.

### What changed?

- Introduced a new Operation Pattern RPC framework that enables:
  - Simple function calls between any client and consumer
  - Stream/subscription based communication
  - Support for transferable objects
  - Automatic message handling and routing
  - Built-in timeout and cancellation
- Added support for multiple transport layers including:
  - Web Workers
  - Shared Workers  
  - BroadcastChannel
  - MessagePort

### How to test?

1. Create an operation:
```ts
class AddOp extends Op<{ a: number; b: number }, number> {}
```

2. Register handler:
```ts
const consumer = new OpConsumer(port);
consumer.register(AddOp, ({ a, b }) => a + b);
```

3. Call operation:
```ts
const client = new OpClient(port);
const result = await client.call(new AddOp({ a: 1, b: 2 })); // 3
```

### Why make this change?

To provide a standardized way of handling communication between different JavaScript contexts (frontend/backend/workers) with:

- Type-safe RPC calls
- Built-in error handling
- Support for streaming data
- Clean abstraction over different transport mechanisms
- Reduced boilerplate code compared to raw message passing

### DIFF
--- a/packages/common/infra/src/op/__tests__/client.spec.ts
+++ b/packages/common/infra/src/op/__tests__/client.spec.ts
@@ -1,0 +1,215 @@
+import { afterEach } from 'node:test';
+
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { OpClient } from '../client';
+import { type MessageHandlers, transfer } from '../message';
+import type { OpSchema } from '../types';
+
+interface TestOps extends OpSchema {
+  add: [{ a: number; b: number }, number];
+  bin: [Uint8Array, Uint8Array];
+  sub: [Uint8Array, number];
+}
+
+declare module 'vitest' {
+  interface TestContext {
+    producer: OpClient<TestOps>;
+    handlers: MessageHandlers;
+    postMessage: ReturnType<typeof vi.fn>;
+  }
+}
+
+describe('op client', () => {
+  beforeEach(ctx => {
+    const { port1 } = new MessageChannel();
+    // @ts-expect-error patch postMessage
+    port1.postMessage = vi.fn(port1.postMessage);
+    // @ts-expect-error patch postMessage
+    ctx.postMessage = port1.postMessage;
+    ctx.producer = new OpClient(port1);
+    // @ts-expect-error internal api
+    ctx.handlers = ctx.producer.handlers;
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it('should send call op', async ctx => {
+    // @ts-expect-error internal api
+    const pendingCalls = ctx.producer.pendingCalls;
+    const result = ctx.producer.call('add', { a: 1, b: 2 });
+
+    expect(ctx.postMessage.mock.calls[0][0]).toMatchInlineSnapshot(`
+      {
+        "id": "add:1",
+        "name": "add",
+        "payload": {
+          "a": 1,
+          "b": 2,
+        },
+        "type": "call",
+      }
+    `);
+    expect(pendingCalls.has('add:1')).toBe(true);
+
+    // fake consumer return
+    ctx.handlers.return({ type: 'return', id: 'add:1', data: 3 });
+
+    await expect(result).resolves.toBe(3);
+
+    expect(pendingCalls.has('add:1')).toBe(false);
+  });
+
+  it('should transfer transferables with call op', async ctx => {
+    const data = new Uint8Array([1, 2, 3]);
+    const result = ctx.producer.call('bin', transfer(data, [data.buffer]));
+
+    expect(ctx.postMessage.mock.calls[0][1].transfer[0]).toBeInstanceOf(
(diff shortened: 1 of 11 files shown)


forehalo:
* **#8734** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8734?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

packages/common/infra/src/op/client.ts
```diff
@@ -0,0 +1,162 @@
...
+    msg => {
+      const ob = this.obs.get(msg.id);
+      if (!ob) {
+        return;
+      }
+
+      ob.complete();
+    };
+
+  call<In, Out>(op: Op<In, Out>): CancelablePromise<Out> {
```
pengx17: It seems there is no type checking on what are the allowed operations to be called. If I construct an random op and call, the client wouldn't complain.

packages/common/infra/src/op/__tests__/client.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+  interface TestContext {
+    producer: OpClient;
+    handlers: MessageHandlers;
+    postMessage: ReturnType<typeof vi.fn>;
+  }
+}
+
+class AddOp extends Op<{ a: number; b: number }, number> {
+  protected override getId(): string {
+    return 'add';
```
pengx17: The generated id should be uniq per Op instance. It is hard to know if two ids will collide. Imagine a client is called with id="add" twice at the same time, the first call will never be resolved.

forehalo: the op id is auto generated and unique, I will add checker in both client and consumer

packages/common/infra/src/op/message.ts
```diff
@@ -0,0 +1,155 @@
...
+      !KNOWN_MESSAGE_TYPES.has(data.type)
+    ) {
+      return;
+    }
+
+    handler(data as any);
+  };
+}
+
+const TRANSFERABLES_CACHE = new Map<any, Transferable[]>();
```
EYHN, line 117: weakmap ?

packages/common/infra/src/op/consumer.ts
```diff
@@ -0,0 +1,188 @@
...
+
+      return ob$.pipe(share(), takeUntil(fromEvent(signal, 'abort')));
+    });
+  }
+
+  destroy() {
+    super.close();
+    this.registeredOpHandlers.clear();
+    this.processing.forEach(controller => {
+      controller.abort();
```
EYHN, line 183: maybe add abort reason for auto recover

```suggestion
      controller.abort("DESTROY");
```

packages/common/infra/src/op/consumer.ts
```diff
@@ -0,0 +1,188 @@
...
+        },
+        error: error => {
+          this.port.postMessage({
+            type: 'return',
+            id: msg.id,
+            error: error as Error,
+          } satisfies ReturnMessage);
+        },
+        complete: () => {
+          this.processing.delete(msg.id);
```
EYHN, line 85: should we send finish message ? Otherwise the caller's promise will never be resolve ？

packages/common/infra/src/op/message.ts
forehalo: items deleted on sending, no need to use weakmap

packages/common/infra/src/op/consumer.ts
forehalo: the path should always fall into next or error, otherwise the business logic get problem

packages/common/infra/src/op/client.ts
```diff
@@ -0,0 +1,206 @@
...
+
+export class OpClient<Ops extends OpSchema> extends AutoMessageHandler {
+  private readonly callIds = new Map<OpNames<Ops>, number>();
+  private readonly pendingCalls = new Map<string, PendingCall>();
+  private readonly obs = new Map<string, Observer<any>>();
+  private readonly options: OpClientOptions = {
+    timeout: 3000,
+  };
+
+  constructor(port: MessageCommunicapable, options: OpClientOptions = {}) {
```
EYHN, line 37: What if the channel is closed external? like, the worker shut down for upgrade

EYHN (approved):

packages/common/infra/src/op/consumer.ts
forehalo: what can be recovered

packages/common/infra/src/op/client.ts
forehalo: channel management should be kept outside of client/consumer

pengx17 (approved):

PR #8735, opened 2024-11-07
fix: edgeless note preview on the editor settings is editable

akumatus:
Fix issue [BS-1779](https://linear.app/affine-design/issue/BS-1779).

The attribute `contenteditable` will be `true` if the document is not read-only. In this case, the note preview can be focused on and users can edit the note content.

But we can not simply make the document read-only. If the user changes the editor settings, we also need to update the document model, which requires that the document is not read-only.

Thus following code sets the document editable before the model updates and resets the document to read-only after the model is updated.


akumatus:
* **#8735** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8735?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `canary`

<h2></h2>

forehalo (approved):

Issue #8752, opened 2024-11-08
​OIDC Provider does not use mandatory scope field


js-0s:
### What happened?

deploy via docker compose behind https: 
image: ghcr.io/toeverything/affine-graphql:stable (9eece242ca7c) (should be unrelated because the code is unchanged in canary branch)

configure oidc using environment + .env:
```
 affine:
    image: ghcr.io/toeverything/affine-graphql:stable
    command:
      ['sh', '-c', 'node ./scripts/self-host-predeploy && node ./dist/index.js']
    networks:
      affine-net:
      affine-front-net:
        aliases:
          - affine
    depends_on:
      affine_cache:
        condition: service_healthy
      affine_db:
        condition: service_healthy
    volumes:
      # custom configurations
      - ./local_data/affine/config:/root/.affine/config
      # blob storage
      - ./local_data/affine/storage:/root/.affine/storage
    logging:
      driver: 'json-file'
      options:
        max-size: '100m'
    restart: unless-stopped
    environment:
      - NODE_OPTIONS="--import=./scripts/register.js"
      - AFFINE_CONFIG_PATH=/root/.affine/config
      - REDIS_SERVER_HOST=cache
      - DATABASE_URL=postgres://affine:affine@db:5432/affine
      - NODE_ENV=production

      - TELEMETRY_ENABLE=false
      - AFFINE_SERVER_HOST=affine.example.com
      - AFFINE_SERVER_HTTPS=true
      - MAIL_HOST=${AFFINE_EMAIL_HOST}
      - MAIL_PORT=${AFFINE_EMAIL_PORT}
      - MAIL_USER=${AFFINE_EMAIL_USERNAME}
      - MAIL_PASSWORD=${AFFINE_EMAIL_PASSWORD}
      - OAUTH_OIDC_ISSUER=${AFFINE_OIDC_ISSUER}
      - OAUTH_OIDC_CLIENT_ID=${AFFINE_OIDC_CLIENT_ID}
      - OAUTH_OIDC_CLIENT_SECRET=${AFFINE_OIDC_CLIENT_SECRET}
      - OAUTH_OIDC_SCOPE=openid email profile offline_access
      - OAUTH_OIDC_CLAIM_MAP_ID=preferred_username
      - OAUTH_OIDC_CLAIM_MAP_EMAIL=email
```

[...]

PR #8753, opened 2024-11-08
fix(server): Do not require scope field in access token

js-0s:
Remove scope that should not be mandatory and is not used (#8752)

### DIFF
--- a/packages/backend/server/src/plugins/oauth/providers/oidc.ts
+++ b/packages/backend/server/src/plugins/oauth/providers/oidc.ts
@@ -14,7 +14,6 @@ const OIDCTokenSchema = z.object({
   access_token: z.string(),
   expires_in: z.number(),
   refresh_token: z.string(),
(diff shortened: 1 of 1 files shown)


packages/backend/server/src/plugins/oauth/providers/oidc.ts
```diff
@@ -14,7 +14,6 @@ const OIDCTokenSchema = z.object({
   access_token: z.string(),
   expires_in: z.number(),
   refresh_token: z.string(),
```
darkskygit, line 16:
```suggestion
  refresh_token: z.string(),
  scope: z.string().optional(),
```

darkskygit: see https://github.com/toeverything/AFFiNE/issues/8752#issuecomment-2465310261

darkskygit (approved):

PR #8844, opened 2024-11-15
fix(electron): respect locale for spellchecker

### LINKED ISSUE #8837: ​Unable to set the document spellcheck language
### What happened?

I'm using the official Electron-based Linux app which is distributed as an AppImage.

The spellcheck can be enabled in the settings. I set the UI language to English and cleared the locale environment. However, the document spellcheck language seems to be picked up from the desktop environment somehow. That is confusing and wrong, and I don't know how to set it properly. It seems that the spellcheck is a built-in feature of electron that needs to be configured.

I propose to implement at least one of the following (from less to more preferred, in my opinion):

1. Pick up the Linux locale variables or describe the mechanism to set the spellcheck language, if there is no setting.
2. Set the spellcheck language in the settings dialog explicitly
3. Set a language tag for each document that is respected by the spellcheck

Thanks!

pengx17:
may fix #8837.
fix AF-1712

MacOS automatically do spellcheck based on given text. This only works for Win/Linux.

### DIFF
--- a/packages/frontend/apps/electron/src/main/ui/handlers.ts
+++ b/packages/frontend/apps/electron/src/main/ui/handlers.ts
@@ -1,6 +1,7 @@
 import { app, nativeTheme, shell } from 'electron';
 import { getLinkPreview } from 'link-preview-js';
 
+import { isMacOS } from '../../shared/utils';
 import { persistentConfig } from '../config-storage/persist';
 import { logger } from '../logger';
 import type { WorkbenchViewMeta } from '../shared-state-schema';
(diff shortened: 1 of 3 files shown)


pengx17:
* **#8844** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8844?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8844?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`

<h2></h2>

packages/frontend/apps/electron/src/main/ui/handlers.ts
```diff
@@ -220,4 +221,14 @@ export const uiHandlers = {
     app.relaunch();
     app.quit();
   },
+  onLanguageChange: async (e, language: string) => {
+    // only works for win/linux
```
forehalo, line 225: any reason or read more link?

forehalo: I find it it electron docs, better quote it https://www.electronjs.org/docs/latest/tutorial/spellchecker#how-to-set-the-languages-the-spellchecker-uses

forehalo (approved):

PR #8877, opened 2024-11-20
feat(core): at menu journal entries

pengx17:
### performance improvement:

fix AF-1790
When at menu appears or user query changes, the `getMenu` function will recomputed the title & icon of every doc in the workspace. In the original implementation, this will create a lot of adhoc LiveData instances but their values are never watched. The overhead is significant if there are many docs in the workspace.

This PR remove LiveData uses in this specific case. Also breaks down the long sync task for smoother user interaction.

depends on https://github.com/toeverything/blocksuite/pull/8780

### journal enhancement

fix AF-1774
added `DateSelectorDialog` for selecting a date through blocksuite;
added `AtMenuConfigService` for constructing the at menu config

### e2e tests
fix AF-1776
fix PD-1942


pengx17:
* **#8911** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8911?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8877** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8877?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8877?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8899** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8899?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`

<h2></h2>

packages/frontend/core/src/modules/doc-display-meta/services/doc-display-meta.ts
```diff
@@ -143,6 +154,41 @@ export class DocDisplayMetaService extends Service {
...
+    const enableEmojiIcon =
+      this.featureFlagService.flags.enable_emoji_doc_icon.value &&
+      options?.enableEmojiIcon !== false;
+    if (enableEmojiIcon) {
+      const { rest } = extractEmojiIcon(docTitle);
+      return rest;
+    }
+
+    // default
+    return docTitle;
```
CatsJuice: maybe extract this logic shared with `title$`?

tools/cli/src/webpack/config.ts
```diff
@@ -149,6 +149,12 @@ export const createConfiguration: (
       alias: {
         yjs: join(workspaceRoot, 'node_modules', 'yjs'),
         lit: join(workspaceRoot, 'node_modules', 'lit'),
+        '@preact/signals-core': join(
```
EYHN, line 152: why ?

packages/frontend/core/src/modules/journal/store/journal.ts
```diff
@@ -37,6 +37,18 @@ export class JournalStore extends Store {
     });
   }
 
+  getDocJournalDate(docId: string) {
```
EYHN: I think we are better using `watch().value`, which is slower but usually not a performance bottleneck.

I think get() is prone to abuse, and the data is not real-time, which leads to inconsistency.

the codes for `get` and `watch` are usually not shared, which causes us to maintain two sets of data reading logic.

tools/cli/src/webpack/config.ts
pengx17: In development with yarn link blocksuite, there will be multiple instances of `@preact/signals-core`, which will cause issues on effect tracking

packages/frontend/core/src/modules/journal/store/journal.ts
pengx17: I thought so and I agree with you especially on the maintenance concerns. However in real profiling creating a lot of adhoc livedata is actually a significant overhead.  Not sure if there is anything can be improved with `LiveData.computed`?

before (with watch)



after (without watch)

packages/frontend/core/src/modules/doc-display-meta/services/doc-display-meta.ts
pengx17: there are some subtle different between this and title$.. I am not sure if there is a good way to share the logic instead of having two almost identical implementations.

EYHN (approved):

forehalo (approved):

PR #8976, opened 2024-12-02
feat(mobile): show doc title on title bar when scroll down

CatsJuice:


CatsJuice:
* **#8976** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8976?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/8976?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`

<h2></h2>

packages/frontend/core/src/mobile/pages/workspace/detail/mobile-detail-page.tsx
```diff
@@ -237,6 +247,12 @@ const MobileDetailPage = ({
     },
     [openJournal]
   );
+
+  useGlobalEvent(
+    'scroll',
```
pengx17, line 252: shall we add `passive` mode for events like `scroll`?

pengx17 (approved):

packages/frontend/core/src/mobile/pages/workspace/detail/mobile-detail-page.tsx
CatsJuice: Seems not matter? As mentioned in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#using_passive_listeners):

> You don't need to worry about the value of `passive` for the basic [scroll](https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll_event) event. Since it can't be canceled, event listeners can't block page rendering anyway.

And other events like `touchstart`, `touchmove` should already be optimized by browser.

pengx17: I see, thanks!

PR #9268, opened 2024-12-24
chore: base url for mobile debug

doodlewind:
Without this PR, the login requests on mobile devices will be sent to `localhost:8080`, which is incorrect.

### DIFF
--- a/packages/frontend/core/src/modules/cloud/constant.ts
+++ b/packages/frontend/core/src/modules/cloud/constant.ts
@@ -33,7 +33,7 @@ export const BUILD_IN_SERVERS: (ServerMetadata & { config: ServerConfig })[] =
       ? [
           {
             id: 'affine-cloud',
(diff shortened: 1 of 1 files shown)


forehalo (approved):

EYHN (approved):

packages/frontend/core/src/modules/cloud/constant.ts
```diff
@@ -33,7 +33,9 @@ export const BUILD_IN_SERVERS: (ServerMetadata & { config: ServerConfig })[] =
       ? [
           {
             id: 'affine-cloud',
-            baseUrl: 'http://localhost:8080',
+            baseUrl: BUILD_CONFIG.isElectron
```
pengx17: I think it is ok to use location.origin for electron build as well

EYHN (approved):

PR #9320, opened 2024-12-26
fix(editor): use nullable inline editor root element

doodlewind:
Fixes `sentry-7906c03b79a54ede819c56cc15ad9889`


Saul-Mirone (approved):
LGTM 🧰

PR #9411, opened 2024-12-28
feat(editor): kanban mobile drag and drop

alsuren:
this is based on https://github.com/toeverything/blocksuite/pull/9048

This PR includes:
* A fix for when the drag and drop can get into a bad state after the browser sends us a pointercancel event 
* An implementation of drag and drop for mobile
  - [ ] kanban/mobile/controller/drag.ts is mostly copy-paste from kanban/pc/controller/drag.ts . Do you want me to DRY it out at all?
  - [x] I have disabled the "mobile databases are readonly" logic from  https://github.com/toeverything/blocksuite/commit/d90ff22f3c0e21b074625a9a727bc3b7d687d08b until someone can give me a bit more context (I have asked in #technical-discussion on discord). If you have a suggested fix for this, please leave a comment (or just push your proposed fix to my branch).
- [x] I haven't worked out the best way to port my tests yet. Should I add the mobile configs to blocksuite/tests-legacy, or try to port them to tests/affine-mobile or somewhere? I've not worked out who to add a kanban view on affine yet, so this might take more time.

This is based on top of https://github.com/toeverything/blocksuite/pull/9044 so it included a basic playwright test. I'm happy to merge this as two separate PRs or close the other one if you prefer.

TODO/known bugs from QAing (some for now, and some for fixing in later issues/PRs):
- [ ] ignore drags if the user didn't stay in the same place for long enough (how long?) before moving
  - [ ] I guess I should display the tilted drag preview as soon as the timeout expires?
- [ ] (later) the horizontal blue drop position indicator doesn't show if you're dropping into an empty column (this is also a problem on desktop affine.fail)
- [ ] (later) it is possible to drop a card into the section below the kanban board and it turns it into a text card, losing information (I might split this out into its own issue though, because it is also a problem on desktop, and pressing undo will get it back where you started)
- [ ] (later) clicking and dragging on the T causes it to render the tilted drag preview, but it stays where it is rather than following your cursor (this is also a problem on desktop affine.fail: it renders the previews as if you're moving the card successfully, but when you release it silently doesn't do anything)
- [ ] (later?) sometimes after you've been playing about for a while, an invisible <affine-mobile-menu> shows up and blocks all interactions with the page. It does not go away when you refresh the page, so you have to open a new browser tab to continue working. (I think this happens when you click any of the 3 dot menus - are they implemented as a pushstate() with hidden data sessionstorage or something? I am tempted to split out a separate PR to enable editing of mobile data views, because the 3dot menus aren't visible on read-only data views)
- [ ] (later) sometimes after you've been playing around for a while, the drag-to-scroll behaviour breaks. It does not fix itself when you refresh the page, so you have to open a new browser tab to continue working. This feels like a chrome bug?

### DIFF
--- a/blocksuite/affine/data-view/src/view-presets/kanban/mobile/card.ts
+++ b/blocksuite/affine/data-view/src/view-presets/kanban/mobile/card.ts
@@ -22,6 +22,7 @@ const styles = css`
     box-shadow: 0px 2px 3px 0px rgba(0, 0, 0, 0.05);
     border-radius: 8px;
     background-color: var(--affine-background-kanban-card-color);
+    touch-action: none;
   }
 
   .mobile-card-header {
--- a/blocksuite/framework/block-std/src/event/control/pointer.ts
+++ b/blocksuite/framework/block-std/src/event/control/pointer.ts
@@ -1,4 +1,4 @@
-import { IS_IPAD } from '@blocksuite/global/env';
+import { IS_IPAD, IS_MOBILE } from '@blocksuite/global/env';
 import { nextTick, Vec } from '@blocksuite/global/utils';
 
 import { UIEventState, UIEventStateContext } from '../base.js';
@@ -176,12 +176,16 @@ class ClickController extends PointerControllerBase {
   }
 }
 
+const MOBILE_DRAG_DELAY_MS = 200;
+
 class DragController extends PointerControllerBase {
+  private _draggingState: 'released' | 'dragging' | 'moved-too-fast' = 'released';
+
   private readonly _down = (event: PointerEvent) => {
     if (this._nativeDragging) return;
 
     if (!event.isPrimary) {
-      if (this._dragging && this._lastPointerState) {
+      if (this._draggingState === 'dragging' && this._lastPointerState) {
         this._up(this._lastPointerState.raw);
       }
       this._reset();
@@ -203,9 +207,18 @@ class DragController extends PointerControllerBase {
       this._move
     );
     this._dispatcher.disposables.addFromEvent(document, 'pointerup', this._up);
+    if (IS_MOBILE) {
+      setTimeout(() => {
+        if (this._draggingState === 'released' && this._startPointerState === pointerState) {
+          this._draggingState = 'dragging';
+          // NOTE: this is a synthetic event. We might end up with a lot of fake dragStart events
+          // followed by context menu events. If this happens, we could make a new event type instead,
+          // and keep draggingState as 'released' until the first dragMove event.
+          this._dispatcher.run('dragStart', createContext(event, pointerState));
+        }
+      }, MOBILE_DRAG_DELAY_MS);
+    }
   };
-
-  private _dragging = false;
 
   private _lastPointerState: PointerEventState | null = null;
 
@@ -230,15 +243,25 @@ class DragController extends PointerControllerBase {
     this._lastPointerState = state;
 
     if (
-      !this._nativeDragging &&
-      !this._dragging &&
+      IS_MOBILE &&
+      this._draggingState === 'released' &&
+      event.timeStamp - this._startPointerState.raw.timeStamp < MOBILE_DRAG_DELAY_MS &&
       isFarEnough(event, this._startPointerState.raw)
     ) {
-      this._dragging = true;
+      this._draggingState = 'moved-too-fast';
+      return;
+    }
+
+    if (
+      !this._nativeDragging &&
+      this._draggingState == 'released' &&
+      isFarEnough(event, this._startPointerState.raw)
+    ) {
+      this._draggingState = 'dragging';
       this._dispatcher.run('dragStart', createContext(event, start));
     }
 
-    if (this._dragging) {
+    if (this._draggingState === 'dragging') {
       this._dispatcher.run('dragMove', createContext(event, state));
     }
   };
@@ -301,7 +324,7 @@ class DragController extends PointerControllerBase {
   };
 
   private readonly _reset = () => {
-    this._dragging = false;
+    this._draggingState = 'released';
     this._startPointerState = null;
     this._lastPointerState = null;
 
@@ -329,7 +352,7 @@ class DragController extends PointerControllerBase {
       last,
     });
 
-    if (this._dragging) {
+    if (this._draggingState === 'dragging') {
       this._dispatcher.run('dragEnd', createContext(event, state));
     }
 
@@ -343,7 +366,7 @@ class DragController extends PointerControllerBase {
     const { host, disposables } = this._dispatcher;
     disposables.addFromEvent(host, 'touchmove', (event: TouchEvent) => {
       if (
-        this._dragging &&
+        this._draggingState === 'dragging' &&
         this._startPointerState &&
         this._startPointerState.raw.pointerType === 'pen'
       ) {
--- a/tests/affine-mobile/e2e/kanban.spec.ts
+++ b/tests/affine-mobile/e2e/kanban.spec.ts
@@ -27,7 +27,7 @@ test('kanban drag and drop', async ({ page }) => {
   }).getByRole('button').nth(1).click();
 
   // NOTE: the following code is the same as the test in
-  // tests/affine-local/e2e/blocksuite/editor.spec.ts apart from the locator names
+  // tests/affine-local/e2e/blocksuite/editor.spec.ts apart from the locator names and waiting for drag preview
 
   const database = page.locator('affine-database');
   await expect(database).toBeVisible();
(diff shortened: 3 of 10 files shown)


zzj3720:
Thank you for your pull request!

To enhance the development experience and code maintainability, we should consider generalizing the current drag-and-drop functionality to support long-press dragging on mobile devices. At present, the key difference between mobile and PC drag-and-drop is the initiation: long-press on mobile versus click on PC. We can make the code adaptable for both platforms.

Previously, the mobile version was set to readonly due to some database issues. Now that these issues have been resolved, we can safely remove this restriction.

Moreover, to ensure consistency and maintainability, please avoid adding new test code to `blocksuite/tests-legacy`. Instead, it would be more efficient to integrate blocksuite’s test code with AFFiNE's existing test framework.

Thank you for your efforts and understanding!

alsuren:
> To enhance the development experience and code maintainability, we should consider generalizing the current drag-and-drop functionality to support long-press dragging on mobile devices. At present, the key difference between mobile and PC drag-and-drop is the initiation: long-press on mobile versus click on PC. We can make the code adaptable for both platforms.

I haven't explicitly implemented the long press thing. I just make all touch events on cards route through the same logic as the desktop version. We are just using pointer events, rather than the web platform's native drag and drop.

If we want to emulate the web platform's native long-press drag and drop behaviour, we would need to add extra logic for that. I have chosen not to do this because:

1) it's fiddly and hard
2) I'm not sure how to say touch-action: none and then translate non-long-press events back into scroll events
3) long press on chrome brings up the context menu by default on mobile (and on desktop Chrome when in mobile device emulation mode). We would need to disable this to make testing possible.

I'm not using the web drag and drop functionality because of 3 and because it's not implemented on mobile Firefox (Which is the browser I use)

> Previously, the mobile version was set to readonly due to some database issues. Now that these issues have been resolved, we can safely remove this restriction.

That's fantastic news. I will clean that up.

> Moreover, to ensure consistency and maintainability, please avoid adding new test code to `blocksuite/tests-legacy`. Instead, it would be more efficient to integrate blocksuite’s test code with AFFiNE's existing test framework.

Okay. I'll have another go at that. 

I noticed in passing that the affine playwright tests take a lot longer per test than the blocksuite ones (I may have picked an unlucky set of tests though, or imagined it). Should I try to write these as vitest tests instead, or is that going to be too much pain?

[Sorry for the edit. I pressed the send button by accident]

blocksuite/affine/data-view/src/view-presets/kanban/mobile/card.ts
```diff
@@ -22,6 +22,7 @@ const styles = css`
     box-shadow: 0px 2px 3px 0px rgba(0, 0, 0, 0.05);
     border-radius: 8px;
     background-color: var(--affine-background-kanban-card-color);
+    touch-action: none;
```
alsuren, line 25: Note that this causes the browser to not scroll if your touch starts on a card. If you are in readonly mode, this might be a bit annoying. Potentially we could make this dynamic. Options include:

a) a css variable that's passed in from above (`touch-action: var(--kanban-card-touch-action)`)
b) dynamically add a .draggable class to the card, and enable this based on that (`.draggable { touch-action: none; }`)
c) dynamic css (🤮 )
d) dynamically set the style property

This is my first contribution to this codebase, and my first time using lit, so I'm not sure what the best approach is here.

tests/kit/utils/page-logic.ts
```diff
@@ -217,3 +217,11 @@ export const addDatabaseRow = async (page: Page, databaseTitle: string) => {
   });
   await db.locator('.data-view-table-group-add-row-button').click();
 };
+
+
+export const addKanban = async (page: Page, title?: string) => {
+  await page.keyboard.press('/');
+  await expect(page.locator('affine-slash-menu .slash-menu')).toBeVisible();
+  await page.keyboard.type('kanban');
+  await page.getByTestId('Kanban View').click();
```
alsuren, line 226: I am struggling to create the equivalent test in @affine-test/affine-mobile.

On https://affine.fail on mobile (firefox), there is a menu that appears above the keyboard, but it doesn't appear for me on chrome with touch emulation enabled. This is also missing locally in chrome on `yarn run -T affine dev -p @affine/mobile`. Because this is missing and there no slash menu, I can't work out how to create a kanban view (there is no / menu on mobile).

Should I try to debug why this is and fix it, or should we merge this PR without a mobile test?

alsuren: I finally worked out what's going on:

a) it goes away when the text area loses focus (which happens when you switch to the elements pane in devtools, but doesn't happen if you're doing it on a usb-debugged phone, because your phone has its own screen for calculating focus). I got around this by putting a breakpoint in the AffineKeyboardToolbarWidget.render() method in the `if (!this._show$.value)` block
b)  it renders **beneath** (as in z-index) the AppTabs widget, and is **pushed up** (as in y) when the on-screen keyboard renders itself and changes the viewport.

I have pushed the following hack as part of the playwright test:

```ts
  // Hack: force the toolbar to be visible (normally it is hidden behind the AppTabs widget),
  // and we can only click it once the virtual keyboard pushes it up, but there is no virtual
  // keyboard in playwright.
  await page.evaluate(() => {
    const el = document.getElementsByTagName("affine-keyboard-toolbar")[0];
    // @ts-expect-error el is actually an AffineKeyboardToolbarWidget
    el._panelHeight$.value = 260;
  })
```

There is definitely a better way to do this, but this will do for now.

blocksuite/affine/data-view/src/view-presets/kanban/mobile/card.ts
alsuren: Readonly mode might not be as common as I first thought, so it probably isn't worth the complexity.

tests/affine-mobile/e2e/kanban.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  expect(cardId).toBeTruthy()
+
+  // drag the card across by one group
+  await page.locator(`[data-card-id="${cardId}"]`).hover()
+  await page.mouse.down();
+  await page.locator(".mobile-group-header").nth(1).hover()
+
+  await page.pause()
+  // drag preview of card should be visible
+  await expect(page.locator(".with-data-view-css-variable .mobilecard-body")).toBeVisible()
```
alsuren:
```suggestion
  await expect(page.locator(".with-data-view-css-variable .mobile-card-body")).toBeVisible()
```
Editing from my phone because I saw the failure notification email in the night. I will test it on my laptop in the morning if nobody hits rerun in the meantime.

zzj3720:
> 2. I'm not sure how to say touch-action: none and then translate non-long-press events back into scroll events
> 3. long press on chrome brings up the context menu by default on mobile (and on desktop Chrome when in mobile device emulation mode). We would need to disable this to make testing possible.

I apologize for the delayed response; I've been on vacation recently.

Regarding the implementation of long-press dragging, this feature is essential for mobile devices as defined by the product design. If it cannot be implemented at the moment, we are open to temporarily not having the drag-and-drop functionality on mobile, which means we won't need to adjust the drag-and-drop feature for mobile devices for now.

About the issue of long-press triggering the context menu, perhaps we could listen for the `contextmenu` event to prevent the default browser behavior. For simulating the long-press event, we might consider listening for the `pointerdown` event followed by the `pointermove`. If the touch does not move beyond a certain distance—though implementing this user experience enhancement can be secondary—we could trigger the drag action; otherwise, this `pointerdown` would not initiate a drag.

Thank you for your understanding!

alsuren:
> I apologize for the delayed response; I've been on vacation recently.

I completely understand. This is the curse of being an open source maintainer: you get a bunch of drive by contributors when everyone is taking time off, and that's exactly when you want to be taking time off.

> Regarding the implementation of long-press dragging, this feature is essential for mobile devices as defined by the product design.

Thanks for your guidance so far. I have learned a lot already.

Do you have design documents that can be shared/screenshotted? I was hoping that I could get away with shipping a half-baked solution to unblock my use-case (sharing a kanban board with my dad, to help me with job hunting). Do I also need to add a drag handle to help with accessibility? 

> otherwise, this pointerdown would not initiate a drag.

Is it okay for this pointerdown to also not cause the page to scroll? (This is what happens if we have touch-events: none. You need to avoid putting your thumb on a card if you want to start scrolling, because the browser will swallow+ignore your touch of you do)

How evergreen can we be about our browser requirements? If long-press is a hard requirement then I'm almost tempted to implement it in terms of
https://web.dev/articles/drag-and-drop (and only use the pointer events logic if that is not supported).

For now, I will try implementing the long press behaviour as you suggested. Thanks again for your feedback.

alsuren:
> Regarding the implementation of long-press dragging, this feature is essential for mobile devices as defined by the product design

Done. It wasn't as painful as I expected, once I found blocksuite/framework/block-std/src/event/control/pointer.ts

blocksuite/framework/block-std/src/event/control/pointer.ts
```diff
@@ -176,12 +176,16 @@ class ClickController extends PointerControllerBase {
   }
 }
 
+const MOBILE_DRAG_DELAY_MS = 200;
+
 class DragController extends PointerControllerBase {
+  private _draggingState: 'released' | 'dragging' | 'moved-too-fast' = 'released';
```
Saul-Mirone, line 182: I don't think we should add mobile only state in framework.

Saul-Mirone (changes requested):

PR #9431, opened 2024-12-30
perf(editor): reduce per frame viewport dom ops

doodlewind:
Currently when panning / zooming on whiteboard, both position and transform of each block component are updated per frame. The positioning part lead to heavy layout costs, which can be reduced.

Before (~35fps):

image

After (~50fps):

image

Tested environment: TibetTravel templet, Windows 11, i5-1130G7 1.1GHz


Saul-Mirone (approved):

forehalo:
nb

PR #9461, opened 2024-12-31
fix(core): split view reordering crash

pengx17:
a workaround fix AF-1940

### DIFF
--- a/packages/frontend/core/src/modules/workbench/view/split-view/split-view.tsx
+++ b/packages/frontend/core/src/modules/workbench/view/split-view/split-view.tsx
@@ -14,7 +14,7 @@ import {
 import { useService } from '@toeverything/infra';
 import clsx from 'clsx';
 import type { HTMLAttributes, RefObject } from 'react';
-import { useCallback, useMemo, useRef, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { createPortal } from 'react-dom';
 
 import type { View } from '../../entities/view';
@@ -49,6 +49,10 @@ export const SplitView = ({
   const [resizingViewId, setResizingViewId] = useState<View['id'] | null>(null);
   const { appSettings } = useAppSettingHelper();
   const workbench = useService(WorkbenchService).workbench;
+
+  // blocksuite's lit host element has an issue on remounting.
+  // Add a workaround here to force remounting after dropping.
+  const [visible, setVisibility] = useState(true);
 
   const sensors = useSensors(
     useSensor(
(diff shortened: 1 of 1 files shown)


pengx17:
* **#9461** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/9461?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/9461?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/toeverything?ref=merge-queue-instructions-in-stack-comment&prId=9461).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/frontend/core/src/modules/workbench/view/split-view/split-view.tsx
```diff
@@ -102,7 +104,15 @@ export const SplitView = ({
         const fromIndex = views.findIndex(v => v.id === active.id);
         const toIndex = views.findIndex(v => v.id === over?.id);
         onMove?.(fromIndex, toIndex);
+        setVisibility(false);
+        const timeoutId = setTimeout(() => {
+          setVisibility(true);
+        });
+        return () => {
+          clearTimeout(timeoutId);
```
forehalo: `useCallback` doesn't have cleanup logic

packages/frontend/core/src/modules/workbench/view/split-view/split-view.tsx
```diff
@@ -102,6 +106,10 @@ export const SplitView = ({
         const fromIndex = views.findIndex(v => v.id === active.id);
         const toIndex = views.findIndex(v => v.id === over?.id);
         onMove?.(fromIndex, toIndex);
+        setVisibility(false);
+        setTimeout(() => {
```
forehalo: it's a potential hook update after unmount, you could do it with an `useEffect`

```ts
useEffect(() => {
  if (!visible) {
    setTimeout(() => {
      setVisible(true)
    }, 0)

    return clear
  }
}, [visible])
```

forehalo: or directly in render, it will be rendered in next update, I think
```ts
if (!visible) {
  setVisible(true)
}
```

pengx17: I think this case is extremely rare, may not worth it adding an useEffect.
The following does not work. I think setting state in render will reschedule a sync render in the same tick.

```
if (!visible) {
  setVisible(true)
}
```

forehalo: rare but it's still not correct. it will cost time debugging if it happens one day especially when you forget about these lines of code

forehalo (approved):

PR #9465, opened 2024-12-31
refactor(core): remove abort event error on init

doodlewind:
Following error always throw on Affine init, which is actually an `Event` from abort controller that can be handled in upstream.

image

### DIFF
--- a/packages/frontend/core/src/desktop/pages/root/index.tsx
+++ b/packages/frontend/core/src/desktop/pages/root/index.tsx
@@ -19,15 +19,9 @@ export const RootWrapper = () => {
     const abortController = new AbortController();
     defaultServerService.server
       .waitForConfigRevalidation(abortController.signal)
-      .then(() => {
-        setIsServerReady(true);
-      })
-      .catch(error => {
-        console.error(error);
-      });
-    return () => {
-      abortController.abort();
-    };
+      .then(() => setIsServerReady(true))
+      .catch(console.error);
+    return () => abortController.abort();
   }, [defaultServerService, isServerReady]);
 
   return (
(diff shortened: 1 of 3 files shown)


pengx17 (approved):

packages/frontend/core/src/desktop/pages/root/index.tsx
```diff
@@ -19,15 +19,9 @@ export const RootWrapper = () => {
...
-      })
-      .catch(error => {
-        console.error(error);
-      });
-    return () => {
-      abortController.abort();
-    };
+      .then(() => setIsServerReady(true))
+      .catch(console.error);
+    return () => abortController.abort();
```
EYHN, line 24: I usually use the following code:

```suggestion
    return () => abortController.abort(MANUALLY_STOP);
```

and catch error by

```
catch (err) {
  if (err === MANUALLY_STOP) {
    // pass
  }
}
```

doodlewind: I found the error is not coming from `useEffect`, but inside `server.waitForConfigRevalidation`, so returning `MANUALLY_STOP` in hooks won't help.

Do you think following update is acceptable?

```
  waitFor(predicate: (v: T) => unknown, signal?: AbortSignal): Promise<T> {
      signal?.addEventListener('abort', reason => {
        reject(MANUALLY_STOP);
      });
    });
  }
```

EYHN: abort(MANUALLY_STOP) will pass MANUALLY_STOP to the abort event as a parameter. So you don't need to change waitForConfigRevalidation.

PR #9471, opened 2024-12-31
fix(editor): too many digits after decimal point in table calculations

OlegDev1:
Fixes **floating-point precision error**, when using **average, median, range functions**

# Changes
## Average
Before:
Screenshot 2024-12-30 at 18 12 09
After:
Screenshot 2024-12-30 at 18 14 44

## Median
Before:
Screenshot 2024-12-31 at 15 18 55
After:
Screenshot 2024-12-31 at 15 20 46

## Range
Before:
Screenshot 2024-12-30 at 18 20 33
After:
Screenshot 2024-12-30 at 18 21 36

### DIFF
--- a/blocksuite/affine/data-view/src/core/statistics/number.ts
+++ b/blocksuite/affine/data-view/src/core/statistics/number.ts
@@ -30,7 +30,9 @@ export const numberStatsFunctions: StatisticsConfig[] = [
       if (numbers.length === 0) {
         return 'None';
       }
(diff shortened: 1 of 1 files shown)


doodlewind (approved):
Thanks for the fix, and happy new year!

OlegDev1:
> Thanks for the fix, and happy new year!

You are welcome! Happy new year to you too, wishing you a fantastic 2025

PR #9600, opened 2025-01-09
refactor(editor): remove default entry

doodlewind:
* `http://localhost:5173/starter/` -> `http://localhost:5173/`
* `http://localhost:5173/?init=affine-snapshot`

The `init` URL param is still valid:

image


Saul-Mirone (approved):
LGTM

PR #9637, opened 2025-01-10
fix(editor): add default value to progress column

### LINKED ISSUE #9624: ​Progress in Database initially to be a Empty rather than Zero
### What happened?

​Progress in Database initially to be a Empty rather than Zero，caused filter unable to get the items whitch Progress = 0 out
CleanShot 2025-01-10 at 04 08 04@2x
CleanShot 2025-01-10 at 04 08 16@2x
CleanShot 2025-01-10 at 04 09 29@2x

### Distribution version

macOS ARM 64 (Apple Silicon)

Other

### Are you self-hosting?

- [X] Yes

### Relevant log output

_No response_

### Anything else?

_No response_

good-jinu:
Fixes #9624 

## What's the issue?

https://github.com/user-attachments/assets/bd622e28-5e68-4db0-95ef-5efb2969e377

1. Add a new column with the data type "progress"
2. Apply a filter to the "progress" column with the condition "= 0"
3. Observe that no rows are displayed in the filtered results
4. Modify the filter condition to "is empty"
5. Observe that all rows are now displayed in the filtered results
6. Change the value of each "progress" field to a non-zero value
7. Change the value of each "progress" field back to zero
8. Apply the filter to the "progress" column with the condition "= 0" again
9. Observe that all rows with a "progress" value of zero are now displayed, which contradicts the initial behavior where no rows were displayed with the same filter condition

## What's the root cause?

New rows are initialized to null by default, resulting in empty values.

## What's changed?

https://github.com/user-attachments/assets/d9e1c9c7-9a5c-4c16-a83c-7bb8d9067384

1. [In code] Added `defaultValue` field within the `PropertyConfig` type
2. [In code] Added `defaultValue` with zero at `progressPropertyModelConfig`
3. Automatically assigned the default value to cells when a new property is added.
4. Updated cell values to the default value whenever a property type is modified.
5. Set newly created rows to the default value for all properties

### DIFF
--- a/blocksuite/affine/block-database/src/data-source.ts
+++ b/blocksuite/affine/block-database/src/data-source.ts
@@ -180,13 +180,25 @@ export class DatabaseBlockDataSource extends DataSourceBase {
 
   propertyAdd(insertToPosition: InsertToPosition, type?: string): string {
     this.doc.captureSync();
+
+    const metaConfig = this.propertyMetaGet(
+      type ?? propertyPresets.multiSelectPropertyConfig.type
+    );
+
     const result = addProperty(
       this._model,
       insertToPosition,
-      databaseBlockAllPropertyMap[
-        type ?? propertyPresets.multiSelectPropertyConfig.type
-      ].create(this.newPropertyName())
+      metaConfig.create(this.newPropertyName())
     );
+
+    updateCells(
+      this._model,
+      result,
+      Object.fromEntries(
+        this.rows$.value.map(r => [r, metaConfig.config.defaultValue])
+      )
+    );
+
     applyPropertyUpdate(this._model);
     return result;
   }
@@ -306,8 +318,13 @@ export class DatabaseBlockDataSource extends DataSourceBase {
       type: toType,
       data: result.property,
     }));
+    const metaConfig = this.propertyMetaGet(toType);
     const cells: Record<string, unknown> = {};
     currentCells.forEach((value, i) => {
+      if (value == null && result.cells[i] == null) {
+        cells[rows[i]] = metaConfig.config.defaultValue;
+      }
+
       if (value != null || result.cells[i] != null) {
         cells[rows[i]] = result.cells[i];
       }
--- a/blocksuite/affine/data-view/src/property-presets/progress/define.ts
+++ b/blocksuite/affine/data-view/src/property-presets/progress/define.ts
@@ -21,4 +21,5 @@ export const progressPropertyModelConfig =
       return value;
     },
     isEmpty: () => false,
+    defaultValue: 0,
   });
--- a/blocksuite/affine/data-view/src/view-presets/table/table-view-manager.ts
+++ b/blocksuite/affine/data-view/src/view-presets/table/table-view-manager.ts
@@ -321,6 +321,14 @@ export class TableSingleView extends SingleViewBase<TableViewData> {
   ): string {
     const id = super.rowAdd(insertPosition);
 
+    this.properties$.value.forEach(column =>
+      this.cellValueSet(
+        id,
+        column.id,
+        this.propertyMetaGet(column.type$.value).config.defaultValue
+      )
+    );
+
     const filter = this.filter$.value;
     if (filter.conditions.length > 0) {
       const defaultValues = generateDefaultValues(filter, this.vars$.value);
--- a/blocksuite/affine/data-view/src/core/property/types.ts
+++ b/blocksuite/affine/data-view/src/core/property/types.ts
@@ -85,6 +85,7 @@ export type PropertyConfig<
       newValue: Value;
     }>
   ) => Value;
+  defaultValue?: Value;
 };
 
 export type DVJSON =


blocksuite/affine/block-database/src/data-source.ts
```diff
@@ -180,13 +180,25 @@ export class DatabaseBlockDataSource extends DataSourceBase {
 
   propertyAdd(insertToPosition: InsertToPosition, type?: string): string {
     this.doc.captureSync();
+
+    const metaConfig = this.propertyMetaGet(
+      type ?? propertyPresets.multiSelectPropertyConfig.type
+    );
```
good-jinu, line 186: `propertyMetaGet` method looks like this
```typescript
  propertyMetaGet(type: string): PropertyMetaConfig {
    return databaseBlockAllPropertyMap[type];
  }
```

It returns the property configuration object corresponding the type.

blocksuite/affine/block-database/src/data-source.ts
```diff
@@ -180,13 +180,25 @@ export class DatabaseBlockDataSource extends DataSourceBase {
...
+      metaConfig.create(this.newPropertyName())
     );
+
+    updateCells(
+      this._model,
+      result,
+      Object.fromEntries(
+        this.rows$.value.map(r => [r, metaConfig.config.defaultValue])
+      )
+    );
```
good-jinu, line 200: it updates cells to default value whenever the property is added

blocksuite/affine/block-database/src/data-source.ts
```diff
@@ -180,13 +180,25 @@ export class DatabaseBlockDataSource extends DataSourceBase {
...
     const result = addProperty(
       this._model,
       insertToPosition,
-      databaseBlockAllPropertyMap[
-        type ?? propertyPresets.multiSelectPropertyConfig.type
-      ].create(this.newPropertyName())
+      metaConfig.create(this.newPropertyName())
     );
+
+    updateCells(
```
zzj3720, line 194: This will cause a large number of update records to exist in the CRDT. It is recommended not to do this. Instead, a default value should be returned when obtaining the value.

blocksuite/affine/data-view/src/property-presets/progress/define.ts
```diff
@@ -21,4 +21,5 @@ export const progressPropertyModelConfig =
       return value;
     },
     isEmpty: () => false,
+    defaultValue: 0,
```
good-jinu, line 24: The progress value always ranges from 0 to 100. When a new progress item is created, it is initially set to 0.

blocksuite/affine/block-database/src/data-source.ts
```diff
@@ -306,8 +318,13 @@ export class DatabaseBlockDataSource extends DataSourceBase {
       type: toType,
       data: result.property,
     }));
+    const metaConfig = this.propertyMetaGet(toType);
     const cells: Record<string, unknown> = {};
     currentCells.forEach((value, i) => {
+      if (value == null && result.cells[i] == null) {
+        cells[rows[i]] = metaConfig.config.defaultValue;
+      }
```
good-jinu, line 326: it updates cells to default value whenever the property is updated to other type.

blocksuite/affine/data-view/src/view-presets/table/table-view-manager.ts
```diff
@@ -321,6 +321,14 @@ export class TableSingleView extends SingleViewBase<TableViewData> {
   ): string {
     const id = super.rowAdd(insertPosition);
 
+    this.properties$.value.forEach(column =>
+      this.cellValueSet(
+        id,
+        column.id,
+        this.propertyMetaGet(column.type$.value).config.defaultValue
+      )
+    );
```
good-jinu, line 330: It initialize the cells to default value when a row is added

good-jinu (commented):
Leave any comment please! 🙏

All feedback, positive or negative, is welcome!

blocksuite/affine/block-database/src/data-source.ts
good-jinu: > This will cause a large number of update records to exist in the CRDT. It is recommended not to do this. Instead, a default value should be returned when obtaining the value.

Thank you for super fast comment! @zzj3720 

I'm not sure of how to implement returning a default value when obtaining the value. Could you please provide more guidance or an example to assist me in resolving this issue? I would greatly appreciate it.

good-jinu:
```diff
/blocksuite/affine/data-view/src/property-presets/progress/define.ts
@@ -8,16 +8,16 @@ export const progressPropertyModelConfig =
     name: 'Progress',
     type: () => t.number.instance(),
     defaultData: () => ({}),
-    cellToString: ({ value }) => value?.toString() ?? '',
+    cellToString: ({ value }) => value?.toString() ?? '0',
     cellFromString: ({ value }) => {
-      const num = value ? Number(value) : NaN;
+      const num = value ? Number(value) : 0;
       return {
-        value: isNaN(num) ? null : num,
+        value: isNaN(num) ? 0 : num,
       };
     },
-    cellToJson: ({ value }) => value ?? null,
+    cellToJson: ({ value }) => value ?? 0,
     cellFromJson: ({ value }) => {
-      if (typeof value !== 'number') return undefined;
+      if (typeof value !== 'number') return 0;
       return value;
     },
     isEmpty: () => false,
```

How about changing all null or undefined values to return 0 like this? Although this change will internally store as null, it will work correctly when filtering. @zzj3720

captainR0bbo:
What is needed to complete this PR? Was just looking to report this same issue as a bug.

zzj3720:
> What is needed to complete this PR? Was just looking to report this same issue as a bug.

This PR has performance concerns. We'll create a new PR to address the bug.

PR #9797, opened 2025-01-20
chore(editor): vscode ts config warning

doodlewind:


blocksuite/tsconfig.json
```diff
@@ -1,7 +1,4 @@
 {
   // exists for easier extending in blocksuite packages
-  "extends": "../tsconfig.web.json",
-  "compilerOptions": {
```
forehalo, line 4: the affine env will be leaked to blocksuite packages

PR #9824, opened 2025-01-21
fix(core): set state isLoading to false after email changing is confirmed

### LINKED ISSUE #9822: ​Changing email will lead to a blank page.
### What happened?

Steps to reproduce:
- Open settings
- Change email

Will observe a blank page.

https://github.com/user-attachments/assets/e3e63082-9045-4726-9fb4-eb274477f9ff

### Distribution version

Web (app.affine.pro)

Firefox

### Are you self-hosting?

- [ ] Yes

### Relevant log output

```shell

```

### Anything else?

_No response_

UNIDY2002:
Otherwise, a fallback screen will always be displayed.

Should fix #9822


packages/frontend/core/src/desktop/pages/auth/confirm-change-email.tsx
```diff
@@ -41,6 +41,7 @@ export const ConfirmChangeEmail: FC<{
           }
           throw err;
         });
+      setIsLoading(false);
```
pengx17: may put it in  `.final` instead

pengx17 (approved):

Brooooooklyn (approved):

PR #10078, opened 2025-02-11
refactor(editor): remove redundant arg in initEmptyEditor

doodlewind:


blocksuite/tests-legacy/utils/actions/misc.ts
```diff
@@ -65,15 +64,13 @@ async function initEmptyEditor({
...
-  multiEditor = false,
 }: {
   page: Page;
   flags?: Partial<BlockSuiteFlags>;
   noInit?: boolean;
-  multiEditor?: boolean;
 }) {
   await page.evaluate(
-    ([flags, noInit, multiEditor]) => {
+    ([flags, noInit]) => {
```
Saul-Mirone, line 73: we should remove this `initEmptyEditor` function as discussed offline

PR #10127, opened 2025-02-12
refactor(editor): move worker renderer to presets with basic test

doodlewind:
This PR migrates the full renderer entry from `@blocksuite/playground` to `@blocksuite/presets` with test case added.

Daily dev usage: `yarn workspace @blocksuite/presets dev` then open http://localhost:5173/renderer.html

image

This `renderer.html` entry reuses the setup logic for running tests in `viewport-renderer.spec.ts`, which makes it easier developing specific logic outside of the vitest test runner and converting it to test cases, towards a more TDD style workflow. I assume we can also add new `my-feature.html` entries for developing other major in-editor features.


L-Sun (approved):
LGTM

Issue #10171, opened 2025-02-13
​Title overflow in edgeless mode in web interface.


samyogdhital:
### What happened?

https://github.com/user-attachments/assets/bd8d7363-e52e-4514-8c1a-1e894412eb0f

See this title overflow in edgeless mode.
Is this the indented behavior or a bug?

I am using self-hosting version, and I am using MS edge browser.

### Distribution version

Web (app.affine.pro)

### What browsers are you seeing the problem on if you're using web version?

Microsoft Edge

### Are you self-hosting?

- [x] Yes

### Relevant log output

```shell

```

PR #10201, opened 2025-02-14
Fix title overflow in edgeless mode

anirbanpati:
This pull request addresses the issue of title overflow in edgeless mode by adding CSS properties to handle text overflow properly in the `v-element.ts` file.

Issue number: #10171


L-Sun:
It seems that the code changes in this PR are unrelated to the issue you are associated with.

Saul-Mirone:
v-element is used in all our rich text blocks. Those styles shouldn't be added to it.

PR #10338, opened 2025-02-21
refactor(editor): move viewport renderer to gfx

doodlewind:
This allows referencing it in `GfxViewportElement`


blocksuite/framework/block-std/src/gfx/viewport/viewport-element.ts
```diff
@@ -62,6 +65,9 @@ export class GfxViewportElement extends WithDisposable(ShadowlessElement) {
       const host = this.host;
       const modelsInViewport = this.getModelsInViewport();
 
+      // const turboRenderer = host.std.get(ViewportTurboRendererIdentifier);
```
Saul-Mirone: comment

Saul-Mirone (changes requested):

blocksuite/framework/block-std/src/gfx/viewport/turbo-renderer/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './turbo-renderer.js';
+export * from './types.js';
```
Saul-Mirone, line 2: Unacceptable block specific concept in types.ts

```ts
export interface ParagraphLayout {
  sentences: SentenceLayout[];
  zoom: number;
}
```

Saul-Mirone (changes requested):

blocksuite/framework/block-std/src/gfx/viewport/turbo-renderer/index.ts
doodlewind: This won't be away very soon, I'll find a workaround to bridge the `TurboRenderer` with `GfxViewportElement`.

Saul-Mirone: You'll need to add extension load point in `TurboRenderer`, and let paragraph block register add it's only renderer.

PR #10448, opened 2025-02-26
fix(editor): list toggle position offset

doodlewind:
Before:

image

After:

image

This centering approach won't affect heading blocks:

image


Saul-Mirone (commented):
Is this by design?

doodlewind:
Fixed

Saul-Mirone (approved):

Issue #8991, opened 2024-12-02
sentry.io connections despite telemetry switched off


Tealk:
### What happened?

Ich habe sowohl am server als auch über die Weboberfläche die telemetrie abeschaltet dennoch wird laut Netzwerkanalyse verbindungen zu sentry.io aufgebaut

### Distribution version

Web (app.affine.pro)

### What browsers are you seeing the problem on if you're using web version?

Firefox

### Are you self-hosting?

- [X] Yes

forehalo:
sentry.io is used for error reporting, the data upload is only used to help us better diagnosing bugs and will not include any privacy regarding data.

Tealk:
Nevertheless, a connection to the domain is established every x seconds, my log is full of connection attempts. Also, if I switch off telemetry, I assume that everything is switched off, at least that's what the GDPR requires.

samyogdhital:
@forehalo, @Tealk any update on this? I am trying to have confidential stuffs here. So, want you make sure that we can turn off every telemetry data.
An environment variable that does this telemetry disabling will solve the problem.

For tracking issues, I think enterprise users give affine team enough data to see the error log. And I am also one of the users of pro version.
But for some separate document I choose to self-host it, so no data is out from there.
Would love to get this feature in self-host version @Affine team.

PR #10543, opened 2025-03-02
fix: prevent sentry from loading when telemetry is disabled

hackeresq:
**What is the issue being solved with this PR?**

A number of other users have noted that Sentry is still sending network requests even when telemetry is disabled (e.g. #8991)

Likely due to the way Sentry is currently configured in Affine, it is sending network requests on many interactions. Sentry is currently hooked into React's router and will send requests when navigating between pages on Affine. This is effectively another layer of telemetry.

Considering there's legitimate reasons to keep exception tracking enabled in the Cloud version, **this PR only affects the self-hosted deployment.** 

**How does this PR resolve the issue?**

When telemetry not enabled and the currently connected cloud is "self-hosted", Sentry will not be loaded and will not fire network requests.


forehalo (approved):

forehalo:
thanks for you contribution. I updated your pr and make sentry follow telemetry setting without in condition of non-selfhost users.

packages/frontend/track/src/sentry.ts
```diff
@@ -0,0 +1,50 @@
...
+        distribution: BUILD_CONFIG.distribution,
+        appVersion: BUILD_CONFIG.appVersion,
+        editorVersion: BUILD_CONFIG.editorVersion,
+      });
+    },
+    enable() {
+      enabled = true;
+    },
+    disable() {
+      enabled = false;
```
hackeresq, line 43: Since the Sentry object has already created, I don't think changing the `enabled` parameter has an effect. Instead, we should probably be updating the Sentry object, like this:

```js
    enable() {
      enabled = true;
      this.init();
    },
    disable() {
      // https://docs.sentry.io/platforms/javascript/guides/react/configuration/draining/
      enabled = false;
      Sentry.getClient()?.close(0)
    },
```

hackeresq (commented):
Hi @forehalo, just did a bit of testing and unfortunately still seeing Sentry when telemetry is disabled.  I think I spotted the issue here.

packages/frontend/track/src/sentry.ts
forehalo: I checked sentry source code, it turns out `.close` will make sentry itself unrecoverable from 'disabled' status. will have another try tmr.

hackeresq: I took another shot at this in case it's helpful. Here's the relevant [branch](https://github.com/hackeresq/AFFiNE/tree/fix-telemetry-sentry).

Essentially the Sentry BrowserSessions integration doesn't use the `beforeSend()` hook. 
* I created a custom integration that listens for browser session events and blocks them.
* Also added dynamic loading to the integrations only when telemetry is enabled.
* Finally made the default value of `enabled` to be false in the Sentry module until it's set by the AppSettingsHelper.

Interested to hear your thoughts!

forehalo: the `beforeSend` is used but there are `beforeSendSpan` and `beforeSendTransaction` to be overrided but this approach is a little bit verbose. The Integration way is also too complicated. I will do it by just updating options and test pass

PR #10546, opened 2025-03-02
fixes(core): confusing default Shape Text Color in default settings UI

hackeresq:
This PR does two things:

## 1. Fixes default Shape Text Color in default settings UI

The settings for default shape text color had a weird bug that caused the color and label to be mismatched (screenshot below). This PR ensures labels and colors will always be matched. 

image

## 2. Adds "transparent" as a default option for shapes and notes

Also, noticed that there was no "transparent" option for shapes and notes in the default colors. Will be a nice quality of life update for users.

### DIFF
--- a/packages/frontend/core/src/desktop/dialogs/setting/general-setting/editor/utils.ts
+++ b/packages/frontend/core/src/desktop/dialogs/setting/general-setting/editor/utils.ts
@@ -5,6 +5,7 @@ import {
   type Palette,
   resolveColor,
 } from '@blocksuite/affine/blocks';
+import { DefaultTheme } from '@blocksuite/affine-model';
 import { isEqual } from 'lodash-es';
 import { useTheme } from 'next-themes';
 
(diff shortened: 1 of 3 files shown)


pengx17 (approved):
LGTM

packages/frontend/core/src/desktop/dialogs/setting/general-setting/editor/utils.ts
```diff
@@ -26,8 +27,10 @@ export const usePalettes = (
     if (isDark) {
```
fundon, line 27: <p class='graphite__hidden'><i>[Re: line +21]</i></p>
Adding a parameter `isShapeText` might be better

```ts
if (!isShapeText && isDark) {
      if (key === 'Black') {
        key = 'White';
      } else if (key === 'White') {
        key = 'Black';
      }
    }
```
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/10546?utm_source=unchanged-line-comment">Graphite</a>.</p>

hackeresq: I don't think the issue is specific to shape text (although that's where I noticed it). This will be problematic for any colors that are pureBlack or pureWhite since they'll be inverted to acomodate the current theme.

hackeresq:
Closing since #10765 resolves this. 

Will submit separate PR for transparency.

PR #10590, opened 2025-03-03
feat: use tags when creating a new doc using template

hackeresq:
Loving the new template functionality!

However, I noticed the tags assigned to a template are not copied to a new document. This is unexpected. Ideally, the template's tags are copied over to the newly created document.

This PR also copies over tags from the template to the newly created doc.


EYHN (approved):

hackeresq:
@EYHN is there anything else needed for this PR?  Hoping we can merge and include in the next release.

hackeresq:
No longer using Affine.

PR #10624, opened 2025-03-05
perf(editor): lazy DOM update with idle state in gfx viewport

doodlewind:
Currently, `GfxViewportElement` hides DOM blocks outside the viewport using `display: none` to optimize performance. However, this approach presents two issues:

1. Even when hidden, all top-level blocks still undergo frequent CSS transform updates during viewport panning and zooming.
2. Hidden blocks cannot access DOM layout information, preventing `TurboRenderer` from updating the complete canvas bitmap.

To address this, this PR introduces a refactoring that divides all top-level edgeless blocks into two states: `idle` and `active`. The improvements are as follows:

1. Blocks outside the viewport are set to the `idle` state, meaning they no longer update their DOM during viewport panning or zooming. Only `active` blocks within the viewport are updated frame by frame.
2. For `idle` blocks, the hiding method switches from `display: none` to `visibility: hidden`, ensuring their layout information remains accessible to `TurboRenderer`.

[Screen Recording 2025-03-07 at 3.23.56 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.dev/media/video/lEGcysB4lFTEbCwZ8jMv/4bac640b-f5b6-4b0b-904d-5899f96cf375.mov)

While this minimizes DOM updates, it introduces a trade-off: `idle` blocks retain an outdated layout state. Since their positions are updated using a lazy update strategy, their layout state remains frozen at the moment they were last moved out of the viewport:

To resolve this, the PR serializes and stores the viewport field of the block at that moment on the `idle` block itself. This allows the correct layout, positioned in the model coordinate system, to be restored from the stored data.

### DIFF
--- a/blocksuite/affine/shared/src/viewport-renderer/renderer-utils.ts
+++ b/blocksuite/affine/shared/src/viewport-renderer/renderer-utils.ts
@@ -1,5 +1,10 @@
-import type { EditorHost } from '@blocksuite/block-std';
-import { type Viewport } from '@blocksuite/block-std/gfx';
+import type { EditorHost, GfxBlockComponent } from '@blocksuite/block-std';
+import {
+  clientToModelCoord,
+  GfxBlockElementModel,
+  GfxControllerIdentifier,
+  type Viewport,
+} from '@blocksuite/block-std/gfx';
 import { Pane } from 'tweakpane';
 
 import { getSentenceRects, segmentSentences } from './text-utils.js';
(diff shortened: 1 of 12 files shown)


doodlewind:
* **#10624** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/10624?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/10624?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/toeverything?ref=merge-queue-instructions-in-stack-comment&prId=10624).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

blocksuite/framework/block-std/src/gfx/viewport.ts
```diff
@@ -491,4 +500,21 @@ export class Viewport {
...
+      top: this.top,
+      viewportX: this.viewportX,
+      viewportY: this.viewportY,
+      zoom: this.zoom,
+      viewScale: this.viewScale,
+    });
+  }
+
+  deserializeRecord(record?: string) {
+    const result = JSON.parse(record || '{}') as ViewportRecord;
```
Saul-Mirone: add catch for JSON.parse

blocksuite/affine/shared/src/viewport-renderer/renderer-utils.ts
```diff
@@ -23,58 +29,109 @@ export function syncCanvasSize(canvas: HTMLCanvasElement, host: HTMLElement) {
...
+
+  const [clientX, clientY] = clientCoord;
+  const viewportInternalX = clientX - left;
+  const viewportInternalY = clientY - top;
+  const modelX = viewportX + viewportInternalX / zoom / viewScale;
+  const modelY = viewportY + viewportInternalY / zoom / viewScale;
+  return [modelX, modelY];
+}
+
+function getSurfaceBlock(doc: Store) {
```
Saul-Mirone: We have at least 4 same function. Can you merge them into one?

doouding: We could just use `gfx.surface` if the `std` is accessible.

doodlewind: Thanks for the advice. Current `getSurfaceBlock` impls have been unified and lifted up to `block-std` to be reused. I think we can remove other similar impls by using `gfx.surface` (they are actually querying `SurfaceComponent` rather than `SurfaceModel`), this refactoring should be responsible for standalone PRs.

doouding (approved):

blocksuite/affine/shared/src/viewport-renderer/renderer-utils.ts
```diff
@@ -23,58 +27,104 @@ export function syncCanvasSize(canvas: HTMLCanvasElement, host: HTMLElement) {
...
+  const doc = host.doc;
+  const surface = gfx.surface;
+
+  const surfaceChildren = surface?.children || [];
+  const rootChildren = doc.root?.children || [];
+  const nonSurfaceChildren = rootChildren.filter(c => c.id !== surface?.id);
+  const blocks = [...surfaceChildren, ...nonSurfaceChildren];
+  const components = blocks
+    .map(block => gfx.view.get(block.id))
+    .filter(Boolean) as GfxBlockComponent[];
```
L-Sun: You can use `gfx.gfxElements.filter(e => e instanceof GfxBlockElementModel)` here

blocksuite/affine/shared/src/viewport-renderer/renderer-utils.ts
```diff
@@ -23,58 +27,104 @@ export function syncCanvasSize(canvas: HTMLCanvasElement, host: HTMLElement) {
...
+) {
+  const { left, top, viewportX, viewportY, zoom, viewScale } = viewport;
+
+  const [clientX, clientY] = clientCoord;
+  const viewportInternalX = clientX - left;
+  const viewportInternalY = clientY - top;
+  const modelX = viewportX + viewportInternalX / zoom / viewScale;
+  const modelY = viewportY + viewportInternalY / zoom / viewScale;
+  return [modelX, modelY];
+}
```
L-Sun: I think we should move this function to `gfx/viewport.ts` and modify `Viewport.toModelCoordFromClientCoord` to ensure they have the same implementation.🤔

blocksuite/affine/shared/src/viewport-renderer/renderer-utils.ts
doodlewind: Thanks for the hint!

blocksuite/affine/shared/src/viewport-renderer/renderer-utils.ts
doodlewind: Great advice, thanks!

PR #10632, opened 2025-03-05
Support all allowed schemes as per RFC 3986

### LINKED ISSUE #10456: Support third party url links
### Description

Title. Mainly, zotero://select links and other possible low hanging integrations. Also see #8669 for a previous request that was turned down without any reasoning.

### Use case

Academic writers, and linking to other databases.

### Anything else?

_No response_

- [x] Yes I'd like to help by submitting a PR!

DhruvaSambrani:
[RFC3986](https://www.rfc-editor.org/rfc/rfc3986#section-3.1) 
Fixes #10456


forehalo:
@Saul-Mirone take a look

Saul-Mirone:
Sorry, we don't plan to accept this pr because after some discussion. We feel this change will make it possible to attach other apps if their schemes have security issues. As an alternative, we plan to make a redirection page.

image

This will be safer for users.

DhruvaSambrani:
> if their schemes have security issues

Right I  thought this may be a concern ~, but I couldn't think of a possible way to exploit this? Or atleast, not worse than allowing a subset of the schemes? Any concerns with the receiver being compromised also exists with current schemes, and the "Embed preview" can be fixed too.~

> ~make a redirection page~

~1. Is this only to prevent accidental clicks to suspicious links? Instead, would you consider storing user-saved "safe" domains and protocols? Like Discord, where clicking external links prompts the user to agree to open it?
2. How would it work in an offline setting?~ 

EDIT: reading more about it, there are many other apps where custom schemes are blocked. I am not a security expert, and so I retract this.

DhruvaSambrani:
Can we instead support a user allowed list of custom schemes?

PR #10766, opened 2025-03-11
refactor(editor): generic layout type support for turbo renderer

doodlewind:
This PR refactored the turbo renderer architecture to support multiple block layout types.

- New base class `BlockLayoutPainter` and `BlockLayoutProvider` are introduced for writing extendable per-block layout querying and painting logic.
- Paragraph-specific lines are all moved into dedicated classes (`ParagraphLayoutProvider` and `ParagraphLayoutPainter`) under the `/variants/paragraph` dir.
- The `renderer-utils.ts` doesn't contain paragraph-specific logic now.
- The `text-utils.ts` is also now scoped for paragraph only.
- Worker messages are now strongly typed.

Upcoming PR should further implement the block registration system using extension API. The `variants` dir could still exist, since there will be similar rendering logic that can be reused among block types (i.e., between paragraph block and list block).


doodlewind:
* **#10766** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/10766?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/10766?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/toeverything?ref=merge-queue-instructions-in-stack-comment&prId=10766).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

blocksuite/affine/gfx/turbo-renderer/src/variants/paragraph/paragraph-layout-provider.ts
```diff
@@ -0,0 +1,89 @@
+import type { GfxBlockComponent } from '@blocksuite/block-std';
+import { clientToModelCoord } from '@blocksuite/block-std/gfx';
+
+import { BlockLayoutProvider } from '../../layout/block-layout-provider';
+import type { Rect } from '../../types';
+import type { ParagraphLayout } from './paragraph-painter';
+import { getSentenceRects, segmentSentences } from './text-utils';
+
+export class ParagraphLayoutProvider extends BlockLayoutProvider<ParagraphLayout> {
```
Saul-Mirone: Can this be put in the paragraph block package?

doodlewind: Will do in next PR. The `variants` is somewhat needed for sharing deps between paragraph / list / code blocks, etc.

blocksuite/affine/gfx/turbo-renderer/src/turbo-renderer.ts
```diff
@@ -37,6 +43,7 @@ export class ViewportTurboRendererExtension extends LifeCycleWatcher {
 
   static override setup(di: Container) {
     di.addImpl(ViewportTurboRendererIdentifier, this, [StdIdentifier]);
+    layoutProviders.register(new ParagraphLayoutProvider());
```
Saul-Mirone: The register behavior should not be called manually. I think paragraph layout should be an extension and register itself

doodlewind: Thanks for the advice, that's exactly what I need.

Just tried a few options, seems the API design with least boilerplate is to pass entities in `ConfigExtensionFactory`:

```ts
function setupEditor() {
  // ...
  const extensions = [
    TurboRendererConfigFactory({
      // already works
      options: {
        zoomThreshold: 1,
        debounceTime: 1000,
      },
      // TODO
      blockLayoutProviders: [...],
      blockPainters: [...],
    }),
    ViewportTurboRendererExtension
  ]
  // ...
}
```

In this way, each `BlockLayoutProvider` and `BlockPainter` would not be extensions (just plain class), but still can be configured and registered in downstream, with no *paragraph*-related concepts hard-coded in upstream anymore. Do you think this approach could fit the best practice?

doodlewind: There are no paragraph related code in the `turbo-renderer` upstream now. They are now moved to the `block-paragraph` package.

Saul-Mirone: I think the `blockLayoutProviders` and `blockLayoutPainters` can also be extension.

Saul-Mirone (approved):

blocksuite/affine/gfx/turbo-renderer/src/turbo-renderer.ts
doodlewind: Migrated, but the adoption to extension and new package structure has cost too much time this week. I thought there are still rooms to improve in terms of extension API in turbo renderer, which is better to be worked out in housekeeping stage if possible.

PR #10828, opened 2025-03-13
fix(editor): database block e2e test flaky

zzj3720:


Saul-Mirone (approved):
@Saul-Mirone approved this PR from Slack with [Graphite](https://app.graphite.dev/github/pr/toeverything/AFFiNE/10828)

PR #10921, opened 2025-03-17
feat(editor): support member node

Flrande:
Close [BS-2793](https://linear.app/affine-design/issue/BS-2793/inline-member)


packages/frontend/core/src/modules/at-menu-config/services/index.ts
```diff
@@ -310,12 +322,112 @@ export class AtMenuConfigService extends Service {
...
+              blockId: block.blockId,
+              mode: mode as GraphqlDocMode,
+            })
+            .catch(error => {
+              console.error(error);
+            });
+
+          const inlineRange = inlineEditor.getInlineRange();
+          if (inlineRange && inlineRange.length === 0) {
+            inlineEditor.insertText(inlineRange, ' ', {
```
EYHN, line 382: should wait for mentionUser success, before insert

EYHN, line 382: should wait for mentionUser success, before insert

packages/frontend/core/src/modules/at-menu-config/services/index.ts
```diff
@@ -310,12 +322,112 @@ export class AtMenuConfigService extends Service {
...
+
+          notificationService
+            .mentionUser(id, workspaceId, {
+              id: docId,
+              title: docTitle,
+              blockId: block.blockId,
+              mode: mode as GraphqlDocMode,
+            })
+            .catch(error => {
+              console.error(error);
```
EYHN, line 377:
```suggestion
              const err = UserFriendlyError.fromAny(error);
              notify.error({
                title: I18n[`error.${err.name}`](err.data),
              });
```

blocksuite/affine/rich-text/src/inline/presets/affine-inline-specs.ts
```diff
@@ -214,6 +214,28 @@ export const FootNoteInlineSpecExtension = InlineSpecExtension(
   }
 );
 
+export const AffineMemberInlineSpecExtension = InlineSpecExtension(
```
Saul-Mirone, line 217: Should move to blocksuite/affine/inlines/member

PR #10937, opened 2025-03-17
Fix: Prevent Middle Mouse Paste (#7035)

### LINKED ISSUE #7035: Fix X11 accidental middle mouse paste​
### What happened?

On Linux when you use middle mouse to move around, as soon as you release the middle mouse button a new textbox is created with the primary clipboard contents.

Expected behavior: Either the middle mouse does not paste, or at minimum, it does not paste when first having moved the mouse to pan around.

### Distribution version

Web (app.affine.pro)

Chrome

### Are you self-hosting?

- [ ] Yes

### Relevant log output

_No response_

### Anything else?

_No response_

Naresh-official:
## Description
This PR fixes an issue where releasing the middle mouse button after panning on Linux causes an unintended paste action due to the system's primary clipboard behavior.

## Changes Made
- Updated isMiddleButtonPressed function in blocksuite/affine/shared/src/utils/event.ts to properly detect middle-click for panning vs. pasting.
- Prevented pasting behavior if the middle mouse button was used for panning first.
- Added a check to differentiate between intentional pasting vs. panning release.
- Ensured compatibility across different platforms (Linux, Windows, macOS).

### DIFF
--- a/blocksuite/affine/shared/src/utils/event.ts
+++ b/blocksuite/affine/shared/src/utils/event.ts
@@ -29,7 +29,11 @@ export enum MOUSE_BUTTON {
 }
 
 export function isMiddleButtonPressed(e: MouseEvent) {
-  return (MOUSE_BUTTONS.AUXILIARY & e.buttons) === MOUSE_BUTTONS.AUXILIARY;
+  if ((MOUSE_BUTTONS.AUXILIARY & e.buttons) === MOUSE_BUTTONS.AUXILIARY) {
+    e.preventDefault(); // Prevent unintended default behavior
+    return true;
+  }
+  return false;
 }
 
 export function isRightButtonPressed(e: MouseEvent) {


blocksuite/affine/shared/src/utils/event.ts
```diff
@@ -29,7 +29,11 @@ export enum MOUSE_BUTTON {
 }
 
 export function isMiddleButtonPressed(e: MouseEvent) {
-  return (MOUSE_BUTTONS.AUXILIARY & e.buttons) === MOUSE_BUTTONS.AUXILIARY;
+  if ((MOUSE_BUTTONS.AUXILIARY & e.buttons) === MOUSE_BUTTONS.AUXILIARY) {
+    e.preventDefault(); // Prevent unintended default behavior
```
Saul-Mirone, line 33: don't add side effect for validators.

Saul-Mirone (changes requested):

pengx17:
I think this PR can be safely closed and in favor of https://github.com/toeverything/AFFiNE/pull/11496? btw I searched for isMiddleButtonPressed and it seems it is not being used anywhere at all.

PR #10942, opened 2025-03-17
feat(editor): modify the interaction of attachment block

### LINKED ISSUE #10781: [Feature Request]: Modify the interaction for attachment block
### Description

Change the interaction of the **attachment block** so that the toolbar is displayed only when selected, instead of on hover. When selected, the primary color border will appear and the background color need to be removed, same as the linked doc block style. This applies to both card view and embed view.

https://github.com/user-attachments/assets/073dfaaa-50df-4f5e-91d5-07e07965d777

### Use case

_No response_

### Anything else?

_No response_

OlegDev1:
Closes #10781
# Demo
### Before:
Screenshot 2025-03-17 at 23 17 48

Screenshot 2025-03-17 at 23 24 47
### After:
Screenshot 2025-03-17 at 23 18 26

Screenshot 2025-03-17 at 23 18 58

# Changes
Unlike linked document, that uses `renderEmbed` from `EmbedBlockComponent`, attachment block and other similar blocks use different approach to change the style when the block is selected. They use `BlockSelection`, that is called from `CaptionedBlockComponent`.
I verified whether `CaptionedBlockComponent` current block corresponds to the model of attachment block or image block (which is basically an attachment block in embed view). And then the necessary props are passed to `BlockSelection`. Though, `affine-page-image` that is used in attachment block embed view doesn't have a model, so I used a model of `affine-image`


doodlewind (dismissed):
Thanks for the nice improvement! cc @Svaney-ssman to double check after landing canary

Svaney-ssman:
Looks great for the changes, please also move the block toolbar to align to the left while the whole image block is selected
CleanShot 2025-03-18 at 10 20 49@2x

OlegDev1:
Thanks, will work on that

OlegDev1:
Done, the change might affect other components as-well though

Screenshot 2025-03-20 at 23 18 38

OlegDev1:
fixing merge conflict

OlegDev1:
Fixed. Though the change will certainly affect other blocks now

OlegDev1:
Noticed that it didn't work. Now everything works as expected

PR #11081, opened 2025-03-21
refactor(server): config system

forehalo:


forehalo:
* **#11242** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/11242?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11081** <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/11081?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/toeverything/AFFiNE/11081?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/toeverything?ref=merge-queue-instructions-in-stack-comment&prId=11081).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/backend/server/src/app.module.ts
```diff
@@ -225,40 +159,33 @@ export function buildAppModule() {
...
-    if (!plugin) {
-      new Logger('AppBuilder').warn(`Unknown plugin ${name}`);
-      return;
-    }
+    .useIf(() => env.selfhosted, WorkerModule, SelfhostModule)
 
-    factor.use(plugin);
-  });
+    // gcloud
+    .useIf(() => env.gcp, GCloudModule);
```
forehalo, line 189: add gcp env in deploy

packages/backend/server/src/base/config/loader.ts
```diff
@@ -0,0 +1,33 @@
...
+  #config: DeepReadonly<AppConfig> = this.loadDefault();
+
+  constructor(private readonly overrides: DeepPartial<AppConfig>) {}
+
+  get config() {
+    return this.#config;
+  }
+
+  override(updates: DeepPartial<AppConfig>) {
+    // TODO: do the validations
```
forehalo: validate with descriptors

packages/backend/server/src/base/job/queue/executor.ts
```diff
@@ -98,26 +94,19 @@ export class JobExecutor
...
-            ret[`job/queues.${queue}.concurrency`] = true;
-            return ret;
-          },
-          {} as {
-            [key in `job/queues.${Queue}.concurrency`]: true;
-          }
-        )
-        // TODO(@forehalo): fix the override by [payment/service.spec.ts]
-      )) ?? {};
+  setConcurrency(queue: Queue, concurrency: number) {
```
forehalo, line 103: watch on config.changed

packages/backend/server/src/base/storage/providers/fs.ts
```diff
@@ -40,7 +44,9 @@ export class FsStorageProvider implements StorageProvider {
     config: FsStorageConfig,
     public readonly bucket: string
   ) {
-    this.path = resolve(config.path, bucket);
+    this.path = config.path.startsWith('~')
```
forehalo: ～/

packages/backend/server/src/core/config/service.ts
```diff
@@ -1,17 +1,110 @@
...
+        })),
+      ])
+    );
+  });
+
+  async updateConfig(
+    user: string,
+    updates: Array<{ key: string; value: any }>
+  ) {
+    const promises = await this.models.appConfig.save(user, updates);
```
forehalo: Validate first

packages/backend/server/src/core/mail/config.ts
```diff
@@ -1,16 +1,24 @@
...
-defineStartupConfig('mailer', {});
+defineModuleConfig('mailer', {
+  enabled: {
+    desc: 'Whether enabled mail service.',
+    default: false,
+  },
+  SMTP: {
+    desc: 'The SMTP configuration for the mail service.',
+    link: 'https://nodemailer.com/smtp/',
+    default: {},
```
forehalo: Make top layer config

packages/backend/server/scripts/self-host-predeploy.js
```diff
@@ -1,17 +1,10 @@
...
-const SELF_HOST_CONFIG_DIR = '/root/.affine/config';
-
-function generateConfigFile() {
-  const content = fs.readFileSync('./dist/config/affine.js', 'utf-8');
-  return content.replace(
-    /(^\/\/#.*$)|(^\/\/\s+TODO.*$)|("use\sstrict";?)|(^.*lint-disable.*$)/gm,
-    ''
-  );
-}
+const SELF_HOST_CONFIG_DIR = `${homedir()}/.affine/config`;
```
fengmk2, line 7: For existing deployments where the container user is not root, is this considered an incompatible update? Will it cause the private key to be regenerated again, and what impact will it have?

packages/backend/server/src/app.module.ts
fengmk2: Will self-host deployment on GCP also be forced to open? The `PLATFORM` environment variable is set by the user, right? The GCP environment won't set it automatically?

packages/backend/server/src/core/storage/config.ts
```diff
@@ -1,34 +1,44 @@
...
-
-export interface StorageStartupConfigurations {
-  avatar: StorageConfig<{
-    publicLinkFactory: (key: string) => string;
-    keyInPublicLink: (link: string) => string;
-  }>;
-  blob: StorageConfig;
+export interface Storages {
+  avatar: {
+    storage: ConfigItem<StorageProviderConfig>;
```
forehalo, line 9: incase there is requirement to separate blobs with different s3/r2 accounts/tokens, I brings storage 
 provider config here

packages/backend/server/src/core/storage/config.ts
```diff
@@ -1,34 +1,44 @@
...
+      provider: 'fs',
+      bucket: 'avatars',
+      config: {
+        path: '~/.affine/storage',
+      },
+    },
+  },
+  'blob.storage': {
+    desc: 'The config of storage for all uploaded blobs(images, videos, etc.).',
+    default: {
```
forehalo, line 41: need to add schema for different provider

```
"oneOf": [
  { provider: 'fs' },
  { provider: 'r2', config: {  accountId: 'xxx' } },
  // ...
]
```

packages/backend/server/src/core/sync/gateway.ts
```diff
@@ -218,8 +191,6 @@ export class SpaceSyncGateway
     @MessageBody()
     { spaceType, spaceId, clientVersion }: JoinSpaceMessage
   ): Promise<EventResponse<{ clientId: string; success: true }>> {
-    await this.assertVersion(client, clientVersion);
```
forehalo, line 221: check in auth now

packages/backend/server/src/plugins/copilot/config.ts
```diff
@@ -1,30 +1,71 @@
...
+    },
+    link: 'https://github.com/openai/openai-node',
+  },
+  'providers.fal': {
+    desc: 'The config for the fal provider.',
+    default: {
+      apiKey: '',
+    },
+  },
+  'providers.google': {
```
forehalo: should we call it `providers.gemini`

packages/backend/server/src/base/job/queue/config.ts
```diff
@@ -1,61 +1,84 @@
...
+        attempts: 5,
+        // should remove job after it's completed, because we will add a new job with the same job id
+        removeOnComplete: true,
+        removeOnFail: {
+          age: 24 * 3600 /* 1 day */,
+          count: 500,
+        },
       },
     },
+    link: 'https://api.docs.bullmq.io/interfaces/v5.QueueOptions.html',
```
fengmk2, line 44: Hope we can remember to make changes here when the bullmq version is upgraded later

packages/backend/server/src/plugins/license/service.ts
```diff
@@ -22,35 +21,22 @@ interface License {
 }
 
 @Injectable()
-export class LicenseService implements OnModuleInit {
+export class LicenseService {
```
forehalo, line 24: disable this injection when env.selfhosted

packages/backend/server/src/base/metrics/opentelemetry.ts
```diff
@@ -81,39 +81,58 @@ export abstract class OpentelemetryFactory {
...
+  private async setup() {
+    if (this.config.metrics.enabled) {
+      if (!this.#sdk) {
+        this.#sdk = this.create();
+      }
+      this.#sdk.start();
+      this.logger.log('OpenTelemetry SDK started');
+    } else {
+      await this.#sdk?.shutdown();
+      this.#sdk = null;
```
fengmk2, line 134: Will the sdk become null after the `config.changed` event is triggered?

packages/backend/server/src/core/config/service.ts
```diff
@@ -1,17 +1,110 @@
...
+    const overrides: DeepPartial<AppConfig> = {};
+    // only take successfully saved configs
+    promises.forEach(promise => {
+      if (promise.status === 'fulfilled') {
+        set(updates, promise.value.id, promise.value.value);
+      }
+    });
+    this.configLoader.override(overrides);
+    this.event.emit('config.changed', { updates: overrides });
+    this.event.broadcast('config.changed.broadcast', { updates: overrides });
```
fengmk2, line 88: If a node restarts or crashes during the `broadcast`, or for some special reason does not listen to this `broadcast` event, is there a possibility of configuration inconsistency?

Once this situation occurs, we need to wait until the next process restart to pull the configuration from the database to achieve eventual consistency.

At this stage, this low-probability event should still be acceptable; in fact, this is a function of the configuration service.

packages/backend/server/scripts/self-host-predeploy.js
forehalo: deployment in docker always use root user

packages/backend/server/src/app.module.ts
forehalo: it won't

packages/backend/server/src/base/metrics/opentelemetry.ts
forehalo: what's the point?

packages/backend/server/src/core/config/service.ts
forehalo: add a cron to fetch latest config every 5 mins?

fengmk2: I just happened to think of this potential issue, so let's ignore it for now. 
Since the change frequency is very low, the probability of it occurring is even lower. Let's wait and see a better solution if a problem really arises later.

packages/backend/server/src/base/metrics/opentelemetry.ts
fengmk2: Then there's no way to continue reporting data? Do we need to reinitialize a new SDK instance?

packages/backend/server/src/core/config/service.ts
```diff
@@ -1,17 +1,110 @@
...
+    user: string,
+    updates: Array<{ key: string; value: any }>
+  ) {
+    const promises = await this.models.appConfig.save(user, updates);
+
+    const overrides: DeepPartial<AppConfig> = {};
+    // only take successfully saved configs
+    promises.forEach(promise => {
+      if (promise.status === 'fulfilled') {
+        set(updates, promise.value.id, promise.value.value);
```
fengmk2: should use `overrides` not `updates`

packages/backend/server/src/core/config/service.ts
```diff
@@ -1,17 +1,110 @@
...
+  async updateConfig(
+    user: string,
+    updates: Array<{ key: string; value: any }>
+  ) {
+    const promises = await this.models.appConfig.save(user, updates);
+
+    const overrides: DeepPartial<AppConfig> = {};
+    // only take successfully saved configs
+    promises.forEach(promise => {
+      if (promise.status === 'fulfilled') {
```
fengmk2, line 80: need to log the error result

packages/backend/server/src/base/config/loader.ts
```diff
@@ -0,0 +1,33 @@
...
+    this.#config = Object.freeze(merge({}, this.#config, updates));
```
fengmk2: Should it only be merged into the `this.#config` object instance? Instead of creating a new object instance? Otherwise, the Config that has already been injected will still reference the old config instance?

```ts
merge(this.#config, updates)
```

packages/backend/server/src/core/mail/sender.ts
```diff
@@ -17,25 +17,51 @@ export type SendOptions = Omit<SendMailOptions, 'to' | 'subject' | 'html'> & {
...
-      this.smtp = createTransport(config);
-    } else if (this.config.node.dev) {
+  @OnEvent('config.changed')
+  onConfigChanged(event: Events['config.changed']) {
+    if ('mailer' in event.updates) {
+      this.setup();
+    }
+  }
+
+  setup() {
```
fengmk2: private method

packages/backend/server/src/base/metrics/opentelemetry.ts
forehalo: it's in `enabled` branch

.docker/selfhost/schema.json
```diff
@@ -0,0 +1,890 @@
...
+        }
+      }
+    },
+    "crypto": {
+      "type": "object",
+      "description": "Configuration for crypto module",
+      "properties": {
+        "privateKey": {
+          "type": "string",
+          "description": "undefined\n@default \"\"\n@environment `AFFINE_PRIVATE_KEY`",
```
darkskygit: undefined in description

packages/backend/server/src/core/config/service.ts
```diff
@@ -1,17 +1,121 @@
...
+    this.event.broadcast('config.changed.broadcast', { updates: overrides });
+  }
+
+  @OnEvent('config.changed.broadcast')
+  onConfigChangedBroadcast(updates: DeepPartial<AppConfig>) {
+    this.configFactory.override(updates);
+    this.event.emit('config.changed', { updates });
+  }
+
+  @Cron(CronExpression.EVERY_5_MINUTES)
```
fengmk2: Let's not add this scheduled full update for now, as it will trigger a full config.changed event every 5 minutes, causing many clients to reinitialize. 
For instance, smtp will recreate the tcp connection, and there are some clients for which we are unsure if repeated initialization might lead to memory leak issues.

packages/backend/server/src/base/event/index.ts
```diff
@@ -6,7 +6,10 @@ import { EventHandlerScanner } from './scanner';
 
 const EmitProvider = {
   provide: EventEmitter2,
-  useFactory: () => new EventEmitter2(),
+  useFactory: () =>
+    new EventEmitter2({
+      maxListeners: 100,
```
fengmk2, line 11: Why does this need to be increased to 100?

fengmk2 (approved):
LGTM

Added via Giphy

packages/backend/server/src/base/event/index.ts
forehalo: the default was 10, it keep warning us about 'config.init' and 'config.changed'

packages/backend/server/src/core/config/service.ts
forehalo: we need a diff here

PR #11087, opened 2025-03-22
fix(editor): ui and edgeless bugs

### LINKED ISSUE #10559: When dragging in Edgeless, the last selected element will be selected
### What happened?

When I use the middle mouse button to drag on Edgeless and release it, the last element I selected is reselected.
#### Expected behavior:
“Not selected” state at the end of the drag and drop.
#### bug video:
https://github.com/user-attachments/assets/2976e8df-9552-4e43-a98c-964b69a29225

### LINKED ISSUE #10981: [Bug]: The icon of the link under edgeless is not aligned.
### What happened?

Image

### Distribution version

macOS ARM 64 (Apple Silicon)

### App Version

_No response_

_No response_

### Are you self-hosting?

- [ ] Yes

### Self-hosting Version

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

OlegDev1:
Fixes #10981


blocksuite/framework/block-std/src/gfx/tool/tool-controller.ts
```diff
@@ -524,7 +524,6 @@ export class ToolController extends GfxExtension {
       ...currentTool.activatedOption,
       type: toolNameStr,
     } as GfxToolsFullOptionValue;
-    currentTool.activate(currentTool.activatedOption);
```
Saul-Mirone: @doouding PTAL

doouding (changes requested):
Hi Oleg, I don't think we can delete currentTool.activate, because it is used to activate the tool. The problem with this bug is that the last selection state is not handled properly. You could just leave the second issue alone, the selection logic is being refactoring right now, and it's not easy to make this change. I'll take it.

OlegDev1:
> Hi Oleg, I don't think we can delete currentTool.activate, because it is used to activate the tool. The problem with this bug is that the last selection state is not handled properly. You could just leave the second issue alone, the selection logic is being refactoring right now, and it's not easy to make this change. I'll take it.

Thanks, I have reverted to the first issue fix. The second issue fix is now not present in this PR. After debugging, I have noticed that `currentTool.activate` is only used to set the selection state after the drag has ended. But yes, the selection state needs to be changed, thanks for taking this issue

doouding (approved):

Saul-Mirone:
Seems there's one unit test fail.

OlegDev1:
> Seems there's one unit test fail.

It was timed out, I guess you should run it again

PR #11356, opened 2025-04-01
fix(editor): unexpectedly remove native selection of menu and image caption

zzj3720:


Saul-Mirone (approved):

Saul-Mirone (commented):
I've fixed this in #11366

PR #11544, opened 2025-04-08
feat(editor): support image preview for attachment columns

zzj3720:
close: BS-2634

### DIFF
--- a/packages/frontend/core/src/blocksuite/database-block/properties/file/view.tsx
+++ b/packages/frontend/core/src/blocksuite/database-block/properties/file/view.tsx
@@ -1,6 +1,7 @@
 import { Popover, uniReactRoot } from '@affine/component';
 import { Button } from '@affine/component/ui/button';
 import { Menu, MenuItem } from '@affine/component/ui/menu';
+import { PeekViewService } from '@affine/core/modules/peek-view';
 import {
   type Cell,
   type CellRenderProps,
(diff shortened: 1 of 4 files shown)


packages/frontend/core/src/blocksuite/database-block/properties/file/view.tsx
```diff
@@ -456,8 +457,11 @@ const useFilePreview = (
...
 ): {
   preview: ReactNode;
+  onPreview?: () => void;
   fileType: 'uploading' | 'loading' | 'image' | 'file';
 } => {
+  const peekView = useService(PeekViewService).peekView;
+
   const uploadProgress = useSignalValue(
     file.type === 'uploading'
       ? fileUploadManager?.getUploadProgress(file.id)
```
pengx17, line 506: I think `fileUploadManager.getFileInfo` has memory leak. The image url from `URL.createObjectURL(blob)` is not being revoked once it is no longer used.

packages/frontend/core/src/blocksuite/database-block/properties/file/view.tsx
```diff
@@ -486,9 +490,33 @@ const useFilePreview = (
...
     }
+    const onPreview = () => {
+      peekView
+        .open({
+          type: 'image-list',
+          data: {
+            image: {
+              url: loadFileData.url,
+              caption: file.name,
+            },
```
pengx17: it seems the `total` prop is not passed in

packages/frontend/core/src/blocksuite/database-block/properties/file/view.tsx
zzj3720: There is a dispose function used for revoking.
```ts
dispose(): void {
    this.fileLoadMap.forEach(fileLoadData => {
      const url = fileLoadData.value?.url;
      if (url) {
        URL.revokeObjectURL(url);
      }
    });

    this.uploadProgressMap.clear();
    this.fileLoadMap.clear();
  }
```

pengx17 (approved):

PR #11820, opened 2025-04-18
ci: Auto-update AppStream on release

eaglesemanation:
Newly added metainfo file got immediately outdated. Unless there is any desire to manually keep release notes up to date - this should be a good enough solution.

Alternatively, this could be converted into a release gate. It would prevent running a release workflow unless metainfo file already has a matching version.


.github/workflows/release-desktop.yml
```diff
@@ -51,6 +51,19 @@ jobs:
...
+        if: ${{ github.event.inputs.build-type == 'stable' && !github.event.inputs.is-draft && !github.event.inputs.is-pre-release }}
+        uses: ./.github/actions/update-appstream
+        with:
+          version: ${{ steps.version.outputs.APP_VERSION }}
+          path: packages/frontend/apps/electron/resources/affine.metainfo.xml
+      - name: Commit new AppStream file
+        uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: 'chore(client): Add v${{ steps.version.outputs.APP_VERSION }} release to affine.metainfo.xml'
+          file_pattern: packages/frontend/apps/electron/resources/affine.metainfo.xml
```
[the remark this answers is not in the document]
eaglesemanation: git-auto-commit-action will not create a commit unless there are any changes to repo. So this check would be redundant.

pengx17 (commented):
Instead of adding a new action, maybe you can modify the existing `set-version.sh` script? That should be the script that the maintainer will run before each release. cc @forehalo @Brooooooklyn

eaglesemanation:
Rewrote it as a script. Been a while since I wrote JS, tried to minimize extra dependencies, hope that's good

packages/frontend/apps/electron/resources/affine.metainfo.xml
```diff
@@ -39,35 +37,38 @@
...
+    <ul>
+      <li>Write your way to better productivity. Build up your content like blocks and let your ideas run wild.</li>
+      <li>Draw and visualise with ease and creativity. Visualise your creativity with others. No constraints, limited only by your imagination.</li>
+      <li>Plan, track, and collaborate efficiently. Stay on top of your workload and achieve more in less time.</li>
+      <li>AI partner helps you better write, draw and plan. Let you think bigger, create faster, work smarter in anytime, anywhere</li>
+    </ul>
   </description>
-
-  <launchable type="desktop-id">pro.affine.AFFiNE.desktop</launchable>
+  <launchable type="desktop-id">affine.desktop</launchable>
```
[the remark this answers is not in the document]
eaglesemanation:
```
# ar x affine-0.21.3-stable-linux-x64.deb
# tar tvf data.tar.zst | grep desktop
-rw-r--r-- root/root       202 2025-04-18 05:45 ./usr/share/applications/affine.desktop
```
Looks fine

pengx17 (approved):

scripts/update-appstream.mjs
```diff
@@ -0,0 +1,44 @@
+import fs from 'node:fs/promises';
```
forehalo: pushing to releases will soon make it quite big(including beta and canary). if the old releases aren't that important, we could simply replace the `version=x` to the latest.

eaglesemanation: Oh, set-version is being executed for beta and canary as well? I'd exclude those. For stable releases I don't think that's too big of a concern. And even if it becomes one - you could migrate to external releases file: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases

I double-checked Flathub submission guidelines, they don't explicitly state that all releases need to be included. But I'd prefer to keep it in, it would probably affect how app store render version history.

forehalo: if a downgrade option is not provided then the version history is meaningless, and we can't promise downgrading the app won't cause data lost cause it's hard to do forward-compatibility. I will directly update your pr before it gets merged.

PR #12051, opened 2025-04-29
fix(server): read the .env file on time

congzhou09:
Problem in @affine/server:
At the time of the globalThis.env's creation in packages\backend\server\src\env.ts, the .env file has not yet been read into process.env, which causes some configurations in the .env file to be invalid. For example, the server always print "AFFiNE Server is running in [affine] mode" even when the DEPLOYMENT_TYPE="selfhosted" configuration is set.

Solution:
Read the .env file into process.env before the first import of packages\backend\server\src\env.ts in packages/backend/server/src/prelude.ts using dotenv, which is already available via @nestjs/config in NestJS framework.

## Summary by CodeRabbit

- **Chores**
  - Improved the initialization process for environment configuration to ensure more controlled and reliable setup. No changes to user-facing features or functionality.

### DIFF
--- a/packages/backend/server/src/prelude.ts
+++ b/packages/backend/server/src/prelude.ts
@@ -1,12 +1,12 @@
 import 'reflect-metadata';
-import './dotenv';
-import './env';
 
 import { existsSync, readFileSync } from 'node:fs';
 import { homedir } from 'node:os';
 import { join } from 'node:path';
 
 import { config } from 'dotenv';
+
+import { createGlobalEnv } from './env';
 
 const CUSTOM_CONFIG_PATH = `${homedir()}/.affine/config`;
 
@@ -38,3 +38,4 @@ function load() {
 }
 
 load();
+createGlobalEnv();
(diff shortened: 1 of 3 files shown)


packages/backend/server/src/prelude.ts
```diff
@@ -1,4 +1,5 @@
 import 'reflect-metadata';
+import './dotenv';
```
forehalo: better doing it by removing the `gloablThis.env = new Env()` in `./env.ts` and creating it after `load()`

congzhou09: OK, let me sort out the process flow of this section (including isPrivateKeyFromEnv and load()) later.

congzhou09: > better doing it by removing the `gloablThis.env = new Env()` in `./env.ts` and creating it after `load()`

Done.
I didn't notice the codes in `load()`. Thank you for your guidance!

forehalo (approved):

PR #12083, opened 2025-04-30
feat(deploy-kube): Provide Kubernetes manifests for deploying AFFiNE …

jnpacker:
…to k3s

This is a first kick at the can in supplying a Kubernetes manifest. This works on k3s, but should also be usable on any of the cloud provider Kubernetes offerings. The core AFFiNE service might just need to be modified depending on your provider.

Usually Kubernetes manifests are multiple files, and this is how I deploy it. I combined it into a single file to simplify the packaging.

If there is interested, this could be migrated to a helm chart, in which case it should distributed back into multiple yaml files.

I also used `*.yml` to match other yaml files in the repository, but I do prefer to use `*.yaml`

I do have one question on the migration container (pod). This looks more like a job in docker, though `kompose` converted it to a full deployment, so I think the fact that it runs, then terminates, eventually gets seen as a failure by Kubernetes and so the pod is run again.  This doesn't seem to hurt anything, but it does come across as a bunch of pod restarts.

## Summary by CodeRabbit

- **New Features**
	- Added comprehensive documentation for deploying the service on Kubernetes, including step-by-step setup instructions and configuration guidance.
	- Introduced Kubernetes deployment manifests to simplify running the application and its dependencies (PostgreSQL and Redis) with persistent storage and external access.

### DIFF
--- a/deploy-kube.md
+++ b/deploy-kube.md
@@ -1,0 +1,32 @@
+# How to install on Kubernetes
+
+## Required
+  1. k3s (this is where this file was tested)
+  2. Metal Load Balancer (on k3s) to expose the external IP on the k3s cluster (see `type: LoadBalancer`)
+
+## Modify the manifest
+Modify the following file `./deploy-kube.yml`
+```
+# Update the AFFINE_SERVER_EXTERNAL_URL (2x occurances)
+            - name: AFFINE_SERVER_EXTERNAL_URL
+              value: http://<MY_URL_OR_IPADDRESS>
+
+# Update the DATABASE_URL (2x occurances)
+            - name: DATABASE_URL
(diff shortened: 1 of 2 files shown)


forehalo (commented):
it's not right time to introduce another k8s manifest. we will make our own one more flexible to be able to deploy at any k8s providers.

at `.github/helm/affine`

patroldo:
Hey, @forehalo

Just curious about status of helm chart? I checked github/helm/affine, but i see quite a lot of difference compared to official docker-compose provided. Some more service out there and haven't yet found migration job

Could you please help, if any time soon are you going to finalize and release Affine helm chart or batch of kubernetes manifests?

If not soon - i can make helm chart from docker-compose and host a repo for it in my github as temporal unofficial and share it with you

PR #12342, opened 2025-05-17
fix(core): don’t open multiple settings dialogs

### LINKED ISSUE #12336: [Bug]: AFFiNE opens multiple settings modals when using keyboard shortcut to open them
### What happened?

On macOS, pressing `⌘,` to open the settings modal opens a new settings modal, even if one was already open.

You can press it 20 times to get 20 settings modals that you then have to close one by one 😅

This happens both on the desktop app and on a browser (Chrome, in my case)

### Distribution version

macOS ARM 64 (Apple Silicon)

### App Version

0.21.6

_No response_

### Are you self-hosting?

- [ ] Yes

### Self-hosting Version

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

bomberstudios:
Fixes #12336

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented multiple 'setting' dialogs from being opened at the same time. If a 'setting' dialog is already open, attempting to open another will now focus the existing dialog instead.


forehalo:
im gonna close this because although the code can prevent duplicated settings modal to be open, the hardcode stuff is quite not cool.

bomberstudios:
Would you accept a PR that did not have a check for `settings`, but would return any type of dialog if it's already open?

Issue #8716, opened 2024-11-05
Disable demo workspace


kendofriendo:
@darkskygit reference to #6141 does not address this issue. Disabling sign-up does not disable the demo workspace. 

Anyone with the url of the selfhosted instance is able to use the demo workspace.

_Originally posted by @ylbeethoven in https://github.com/toeverything/AFFiNE/discussions/5975#discussioncomment-8802555_

RichardQian:
What's the problem this feature is trying to solve?

a demo workspace only exists on visitor's machine, you may essentially view it as a fancy placeholder page, anything the user do in the demo workspace will not change the data on your server ( unless , they successfully signup and migrate unto private cloud)

JokerQyou:
> Anyone with the url of the selfhosted instance is able to use the demo workspace.

This part concerned me for a while, but it turns out guests are only getting a newly created **local** demo workspace, not related to the cloud-synced one. Still, admins with sign-up disabled might want the feature to turn off demo workspace for guests altogether.

lolokraus:
Since self hosted Affine can run on a public server I would agree that admins should be ale to turn off the demo workspace for guests.

AhmedAburady:
I opted out of it completely because of this

melihsd:
I don't understand the philosophy behind it. Why is this behaviour forced on a self-hosted instance and not configurable? 

Many people want to keep their Affine instance private for various reasons. This is confusing for people because they want to run only a cloud version, but they are taken to a local demo, which is the most bizarre starting page. This is very opinion-based and should be configurable. For example, add a config to disable local projects so that visitors are prompted with a login page like it always was.

I must admit, I really enjoy using affine but sometimes I don't get the complexity of it.

JackPala:
Had to remove my Affine instances because of the open frontend. Too much hypothetical risk to expose that many vectors outwardly to the public. Also don't want to use a secondary auth layer to lock it down as that will only confuse the users.

Hopefully when this issue is resolved and the demo workspace can be removed (and is ideally removed by default) then I'd be happy to revisit this software

BigPebble123:
I also removed AFFiNE from my home server.

It's not just about that specific feature; the whole software feels more like a work in progress than a fully functioning application.

melihsd:
I am now happy with outline

BigPebble123:
Thanks! I'll definitely check it out since I'm in need of a note-taking app, and AFFiNE isn't really stable 😅

[...]

Issue #12256, opened 2025-05-13
[Feature Request]: Optional Demo Skipping for Self-Hosted Users


blakey108:
### Description

Provide an option to disable the automatic creation or display of the demo/starter document when accessing the login page or starting a new workspace in self-hosted environments.

Add a configuration option (e.g. in environment variables or admin settings) to disable the automatic demo/starter doc in self-hosted deployments.

Retain the demo experience for first-time logins only, or make it user-dismissible after the initial view.

### Use case

Self-hosted users typically access the platform with specific intent and are not just trialing the product. The repeated display of the demo/starter doc — even when not logged in — creates confusion and friction, especially for returning users. This feature would improve UX for system admins and regular users in self-hosted setups by reducing unnecessary distractions during login or workspace creation.

### Anything else?

This feature would help align the self-hosted experience with user expectations around control and customization. While the demo is valuable for onboarding, forcing it on every visit (including before login) undermines its purpose in non-cloud contexts.

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

Issue #12219, opened 2025-05-11
Open internet can use a hosted Affine without being signed in?


JackPala:
It seems that on affine instances with a reverse proxy, users can access affine instances in local mode without needing to sign in. Is there a way to require sign in?

Issue #10839, opened 2025-03-13
[Feature Request]: Disable the local storage workspace on self hosted solution


6im0n:
### Description

Trying to find a way (or feature) to disable the local storage workspace, the ideal solution is prompting to user to login when an user is trying to access to the app.

![Image](https://github.com/user-attachments/assets/ce244d1a-2073-4d88-932d-0bbc4e2c3a18)

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

hwangdev97:
Aight, so AFFiNE's a local-first app, which means we always think about users who might not log in and just use it locally. 

But, if we're just talkin' about self-hosted, some orgs or teams do need that feature. So, if someone's using the self-hosted version, we could think about adding an option in the **admin panel**. If there are any pull requests for that, we'd be happy to merge 'em, but we're not gonna actively set it up ourselves. 😉

geotang01:
Where is result?

PR #12366, opened 2025-05-19
Do not create local workspaces if disabled

OliverRM:
On first launch, the frontend creates a local Demo Workspace. This behavior has until now ignored the `enable_local_workspace` flag, causing a local workspace to be created even if the flag is disabled. This PR changes the behavior to _not_ create a Demo Workspace if the flag is disabled. Instead, a UI is shown that offers the user to sign in.

### Suggestions for future improvement:
The UI shown when no Demo Workspace has been created is as of now very basic. I suggest to replace the UI with a more appealing one in the future.

### Addressed issues:
#8716 #10839 #12219  #12256

## Summary by CodeRabbit

- **New Features**
  - Added a feature flag to control the availability of local workspace creation. This feature will only be enabled for users when the corresponding flag is active.

### DIFF
--- a/packages/frontend/core/src/desktop/pages/index/index.tsx
+++ b/packages/frontend/core/src/desktop/pages/index/index.tsx
@@ -1,4 +1,5 @@
 import { DesktopApiService } from '@affine/core/modules/desktop-api';
+import { FeatureFlagService } from '@affine/core/modules/feature-flag';
 import { WorkspacesService } from '@affine/core/modules/workspace';
 import {
   buildShowcaseWorkspace,
@@ -127,7 +128,14 @@ export const Component = ({
     desktopApi?.handler.ui.pingAppLayoutReady().catch(console.error);
   }, [desktopApi]);
 
+  const featureFlagService = useService(FeatureFlagService);
+  const localWorkspaceEnabled =
+    featureFlagService.flags.enable_local_workspace.value;
+
   useEffect(() => {
+    if (!localWorkspaceEnabled) {
+      return;
+    }
     setCreating(true);
     createFirstAppData(workspacesService)
       .then(createdWorkspace => {
@@ -148,7 +156,7 @@ export const Component = ({
       .finally(() => {
         setCreating(false);
       });
-  }, [jumpToPage, openPage, workspacesService]);
+  }, [localWorkspaceEnabled, jumpToPage, openPage, workspacesService]);
 
   if (navigating || creating) {
     return fallback ?? <AppContainer fallback />;


forehalo:
this flag can disable demo workspace creating but it's not the one controls the creating behavior of a server. it only takes effect on user side

OliverRM:
Thank you @forehalo for your comment. The `enable_local_workspace` controls whether unauthenticated users can create new workspaces. To be more precise, `onNewWorkspace` in packages/frontend/core/src/components/workspace-selector/user-with-workspace-list/index.tsx checks the flag and redirects the user to sign in instead of creating the workspace:
``` typescript
const onNewWorkspace = useCallback(() => {
  if (
    !isAuthenticated &&
    !featureFlagService.flags.enable_local_workspace.value
  ) {
    return openSignInModal();
  }
  // Create workspace
  ...
}
```

This is done in the frontend. Now here comes the bug: When a user opens AFFiNE for the first time, packages/frontend/core/src/desktop/pages/index/index.tsx will create a local demo workspace, ignoring the flag. This way, unauthenticated users will have a local workspace even if they have been deactivated for unauthenticated users in the frontend. This causes issues #8716 #10839 #12219 #12256 .

My PR fixes this be changing packages/frontend/core/src/desktop/pages/index/index.tsx to also check for the feature to be enabled before created a new local workspace. If not, the user is offered to sign in. This also makes the check in packages/frontend/core/src/desktop/pages/index/index.tsx align with packages/frontend/core/src/components/workspace-selector/user-with-workspace-list/index.tsx where the flag is check in combination with a check whether the user is authenticated.

``` typescript
if (!localWorkspaceEnabled) {
  return; // Will render workspace selector offering sign in
}
setCreating(true);
createFirstAppData(workspacesService)
...
```

The change fixes a frontend bug and does not affect the backend. Local workspaces are handled by the frontend. Therefore, the check also has to be done by the frontend.

Let me know if there is anything else I can clarify.

forehalo:
simple test: disable `Allow create local workspace` in experimental features, open a new browser and visit affine, a demo workspace will still be created. the setting is remembered in per browser context.

all the issues you listed, want the feature that allows system admins to control whether anonymous users can use the system. and the flag `enable_local_workspace` is not the one

OliverRM:
What do you mean the “the flag enable_local_workspace is not the one”? What else should it be? Looking at packages/frontend/core/src/modules/feature-flag/constant.ts
``` typescript
  enable_local_workspace: {
    category: 'affine',
    displayName:
      'com.affine.settings.workspace.experimental-features.enable-local-workspace.name',
    description:
      'com.affine.settings.workspace.experimental-features.enable-local-workspace.description',
    configurable: isCanaryBuild,
    defaultState: false,
  },
```
it seems to be it.

“the setting is remembered in per browser context.” That is not an issue. All I am fixing, is that in some places the frontend checks the flag before creating local workspaces and in some places it does not. It should always or we should abandon the flag. Only checking sometimes is inconsistent.

“all the issues you listed, want the feature that allows system admins to control whether anonymous users can use the system.”
I just checked the issues again:
- #8716 is specifically about the creation of demo workspaces.
- #10839 is specifically about local workspaces.
- #12219 suggests, closing AFFiNE complete to unauthenticated users.
- #12256 is again about the demo workspaces

I have to correct myself: My PR fixes #8716 #10839 #12256 , not #12219

We have to distinguish between two different things: Limiting the creation of local workspaces by unauthenticated users and blocking access to the system completely. I am addressing the first one, which of my understanding is what the feature flag is for. Completely blocking access to the system is a totally different thing, it has more complex consequences (like potentially breaking sharing links), and requires a different approach (we would have to change the server to limit what static files it serves).

OliverRM:
Let me put it in different words:

Before the PR:
Flag enabled: When anonymous users open AFFiNE for the first time, a local demo workspace is created 👍 and they can create further local workspaces 👍 
Flag disabled: When anonymous users open AFFiNE for the first time, a local demo workspace is created which the unauthenticated user can use as he wishes to do so❗ but they can’t create further local workspaces 👍

After the PR:
Flag enabled: When anonymous users open AFFiNE for the first time, a local demo workspace is created 👍 and they can create further local workspaces 👍 
Flag disabled: When anonymous users open AFFiNE for the first time, they are asked to sign in and NO local demo workspace is created ✅. Neither can they create further local workspaces 👍

OliverRM:
@forehalo Could you please be more specific what I need to change? I'd happily adjust my PR according to your feedback.

PR #12611, opened 2025-05-28
fix(editor): connector not added as frame child

### LINKED ISSUE #12428: [Bug]: Arrow inside of frame not copied to image
### What happened?

I created two shapes connected by an arrow, created a frame of it and then added another shape connected by an arrow. Then I copied the frame as an image.
The image (pasted in the Edgeless below the frame) only contains the arrow that has been created before the frame. The arrow that connects to the third shape (and has been created once the frame already existed) is missing.

Image

### Distribution version

macOS ARM 64 (Apple Silicon)

### App Version

0.21.6

_No response_

### Are you self-hosting?

- [x] Yes

### Self-hosting Version

0.21.6

### Relevant log output

```shell

```

### Anything else?

_No response_

doufa:
Fix: #12428 

Connectors had invalid bounds: When elementAdded fires, connectors have bounds of [0,0,0,0] because their path/bounds are calculated in a separate microtask by the connector-watcher. Frame detection waits for connector bounds calculation update to complete.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when adding connectors and shapes to frames by deferring frame detection to ensure elements are fully initialized before grouping.
  - Ensured all elements within a frame move together correctly when the frame is moved.

- **Tests**
  - Added an end-to-end test verifying connectors and shapes created simultaneously inside a frame are properly contained and move together.

### DIFF
--- a/blocksuite/affine/blocks/frame/src/frame-manager.ts
+++ b/blocksuite/affine/blocks/frame/src/frame-manager.ts
@@ -241,20 +241,35 @@ export class EdgelessFrameManager extends GfxExtension {
       surfaceModel.elementAdded.subscribe(({ id, local }) => {
         const element = surfaceModel.getElementById(id);
         if (element && local) {
-          const frame = this.getFrameFromPoint(element.elementBound.center);
-
-          // if the container created with a frame, skip it.
-          if (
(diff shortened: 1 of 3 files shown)


L-Sun (approved):
LGTM!

PR #12749, opened 2025-06-08
fix(core): correct emoji extraction logic using regex

NorkzYT:
https://github.com/user-attachments/assets/ef612f34-0388-49a2-bcad-0cac07a5f785

This PR solves the issue where a majority of emoji's are unable to become the document or folders icon.

The regex used is below with the test string of a variety of emoji's:
https://regex101.com/r/0anB6Z/1

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved emoji extraction to better handle a wider variety of emoji combinations at the start of text.

- **Tests**
	- Expanded test coverage for emoji extraction, including additional cases for single and combined emojis followed by digits, and updated negative test scenarios.


NorkzYT:
Perhaps in the future there can be a built-in pop-up that you can select Emoji's, Icons, and an Upload function as well.
Similar to what Notion has which is shown below: 

image
image
image

forehalo (approved):

PR #12810, opened 2025-06-13
feat(server): embedding search tool

darkskygit:
fix PD-2659
fix AI-184

### DIFF
--- a/packages/backend/server/src/plugins/copilot/context/service.ts
+++ b/packages/backend/server/src/plugins/copilot/context/service.ts
@@ -1,4 +1,5 @@
 import { Injectable, OnApplicationBootstrap } from '@nestjs/common';
+import { ModuleRef } from '@nestjs/core';
 
 import {
   Cache,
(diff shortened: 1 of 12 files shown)


packages/backend/server/src/plugins/copilot/context/service.ts
```diff
@@ -192,6 +191,42 @@ export class CopilotContextService implements OnApplicationBootstrap {
...
+        embedding,
+        workspaceId,
+        topK * 2,
+        threshold
+      ),
+    ]);
+
+    if (!fileChunks.length && !workspaceChunks.length) return [];
+
+    return this.embeddingClient.reRank(
```
fengmk2: Keep adding an `await` call instead of returning a promise directly

packages/backend/server/src/plugins/copilot/providers/provider.ts
```diff
@@ -98,6 +110,49 @@ export abstract class CopilotProvider<C = any> {
...
+          continue;
+        }
+        switch (tool) {
+          case 'webSearch': {
+            tools.web_search_exa = createExaSearchTool(this.AFFiNEConfig);
+            tools.web_crawl_exa = createExaCrawlTool(this.AFFiNEConfig);
+            break;
+          }
+          case 'semanticSearch': {
+            const ac = this.moduleRef.get(AccessController, { strict: false });
```
fengmk2, line 140: use Inject AccessController and CopilotContextService instead of moduleRef.get will be better?

darkskygit: there is a circular reference

fengmk2 (approved):

PR #12829, opened 2025-06-16
chore: disable code block html preview

Saul-Mirone:
## Summary by CodeRabbit

- **Chores**
  - Updated the default setting for the code block HTML preview feature flag; it is now enabled only in experimental (canary) builds.


packages/frontend/core/src/modules/feature-flag/constant.ts
```diff
@@ -280,7 +280,7 @@ export const AFFINE_FLAGS = {
     description:
       'com.affine.settings.workspace.experimental-features.enable-code-block-html-preview.description',
     configurable: isCanaryBuild,
-    defaultState: true,
+    defaultState: false,
```
L-Sun:
```suggestion
    defaultState: isCanaryBuild,
```

Should be better if set it to `isCanaryBuild` for passing e2e tests?

L-Sun (approved):

PR #12966, opened 2025-06-30
fix(core): fix client crash

EYHN:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when displaying OAuth provider icons by handling cases where the provider may not be recognized, preventing potential errors during authentication.

### DIFF
--- a/packages/frontend/core/src/components/affine/auth/oauth.tsx
+++ b/packages/frontend/core/src/components/affine/auth/oauth.tsx
@@ -110,7 +110,10 @@ interface OauthProviderProps {
 }
 
 function OAuthProvider({ onContinue, provider }: OauthProviderProps) {
-  const { icon } = OAuthProviderMap[provider];
+  const { icon } =
+    provider in OAuthProviderMap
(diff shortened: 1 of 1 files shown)


pengx17 (approved):

fengmk2 (approved):

darkskygit (approved):

CatsJuice (approved):

packages/frontend/core/src/components/affine/auth/oauth.tsx
```diff
@@ -110,7 +110,10 @@ interface OauthProviderProps {
 }
 
 function OAuthProvider({ onContinue, provider }: OauthProviderProps) {
-  const { icon } = OAuthProviderMap[provider];
```
fengmk2, line 113: The logic here should be refactored; if the provider map does not exist, the login button should not be displayed

EYHN: It seems that even it not in oauth provider map, oauth can still work, just no icon

PR #12971, opened 2025-06-30
feat: title of session

darkskygit:
fix AI-253


packages/backend/server/schema.prisma
```diff
@@ -443,6 +443,7 @@ model AiSession {
   promptName      String    @map("prompt_name") @db.VarChar(32)
   promptAction    String?   @default("") @map("prompt_action") @db.VarChar(32)
   pinned          Boolean   @default(false)
+  title           String?   @db.VarChar(64)
```
fengmk2: why max size is 64?

fengmk2 (approved):

packages/backend/server/src/plugins/copilot/session.ts
```diff
@@ -520,8 +544,78 @@ export class ChatSessionService {
...
+        title ||
+        !messages.length ||
+        messages.filter(m => m.role === 'user').length === 0 ||
+        messages.filter(m => m.role === 'assistant').length === 0
+      ) {
+        return;
+      }
+
+      {
+        const title = await this.chatWithPrompt('Summary as title', {
```
fengmk2, line 606: Should check the max size?

packages/backend/server/schema.prisma
darkskygit: early ideas, removed

PR #13007, opened 2025-07-03
feat(server): use new content reader

darkskygit:
partial fix AI-280

### DIFF
--- a/packages/backend/server/src/core/doc-service/controller.ts
+++ b/packages/backend/server/src/core/doc-service/controller.ts
@@ -48,12 +48,12 @@ export class DocRpcController {
   async getDocMarkdown(
     @Param('workspaceId') workspaceId: string,
     @Param('docId') docId: string,
-    @Query('aiEditable') aiEditable: boolean
+    @Query('aiEditable') aiEditable?: string
   ) {
     const result = await this.docReader.getDocMarkdown(
       workspaceId,
       docId,
-      aiEditable
+      aiEditable === 'true'
     );
     if (!result) {
       throw new NotFound('Doc not found');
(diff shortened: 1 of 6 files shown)


packages/backend/server/src/core/doc-service/controller.ts
```diff
@@ -47,9 +47,14 @@ export class DocRpcController {
   @Get('/workspaces/:workspaceId/docs/:docId/markdown')
   async getDocMarkdown(
     @Param('workspaceId') workspaceId: string,
-    @Param('docId') docId: string
+    @Param('docId') docId: string,
+    @Query('aiEditable') aiEditable: boolean
```
fengmk2: aiEditable is string，should convert it to bool

```ts
aiEditable === 'true'
```

fengmk2: like fullContent does
image

darkskygit: good point, fixed

packages/backend/server/src/plugins/copilot/tools/doc-edit.ts
```diff
@@ -16,8 +16,9 @@ export const buildContentGetter = (ac: AccessController, doc: DocReader) => {
       .doc(docId)
       .can('Doc.Read');
     if (!canAccess) return undefined;
-    const content = await doc.getFullDocContent(options.workspace, docId);
-    return content?.summary.trim() || undefined;
+    const content = await doc.getDocMarkdown(options.workspace, docId, true);
+    console.log(`Doc content for ${docId}:`, content);
```
fengmk2: remove

fengmk2 (approved):

PR #13026, opened 2025-07-04
feat(server): paginated list endpoint

darkskygit:
fix AI-323

### DIFF
--- a/packages/backend/server/src/models/copilot-session.ts
+++ b/packages/backend/server/src/models/copilot-session.ts
@@ -265,26 +265,31 @@ export class CopilotSessionModel extends BaseModel {
       userId: true,
       workspaceId: true,
       docId: true,
-      pinned: true,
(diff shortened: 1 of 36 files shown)


packages/backend/server/src/models/copilot-session.ts
```diff
@@ -330,8 +335,22 @@ export class CopilotSessionModel extends BaseModel {
...
+  }
+
+  async count(options: ListSessionOptions) {
+    return await this.db.aiSession.count({
+      where: this.getListConditions(options),
+    });
+  }
+
+  async list(options: ListSessionOptions) {
+    console.log(
```
fengmk2: use logger.debug instead

darkskygit: oops

fengmk2 (approved):

PR #13028, opened 2025-07-04
feat(server): update session after doc deletion

darkskygit:
## Summary by CodeRabbit

* **New Features**
  * Sessions associated with a deleted document are now automatically updated to remove the document reference.

* **Improvements**
  * Enhanced session management to better handle documents that have been deleted. 

No visible changes to the user interface; these updates improve backend handling of session and document relationships.

### DIFF
--- a/packages/backend/server/src/plugins/copilot/session.ts
+++ b/packages/backend/server/src/plugins/copilot/session.ts
@@ -47,6 +47,10 @@ declare global {
     'copilot.session.generateTitle': {
       sessionId: string;
     };
+    'copilot.session.deleteDoc': {
+      workspaceId: string;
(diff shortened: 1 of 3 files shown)


packages/backend/server/src/plugins/copilot/session.ts
```diff
@@ -580,6 +580,24 @@ export class ChatSessionService {
     return provider.text(cond, [...prompt.finish({}), msg], config);
   }
 
+  @OnJob('copilot.workspace.deleteDoc')
```
fengmk2: A job handler can only have one job with the same job name

fengmk2 (approved):

PR #13052, opened 2025-07-06
fix(server): use config.server.host to be consistent with port

### LINKED ISSUE #12882: [Bug]: AFFINE_SERVER_HOST Seems not working?
### What happened?

I tried to set `AFFINE_SERVER_HOST` from localhost to my own host address, but it seems always pointed to localhost. 

Tried the `AFFINE_SERVER_PORT` its working just fine. Any suggest?

Image

### Distribution version

macOS ARM 64 (Apple Silicon)

### App Version

_No response_

_No response_

### Are you self-hosting?

- [x] Yes

### Self-hosting Version

latest

### Relevant log output

```shell

```

### Anything else?

_No response_

yifeiyin:
fixes #12882

## Summary by CodeRabbit

* **Chores**
  * Updated server to use a configurable host address based on settings, improving flexibility for deployment environments.
  * Changed default server host to listen on all network interfaces by default, enhancing accessibility.

### DIFF
--- a/packages/backend/server/src/server.ts
+++ b/packages/backend/server/src/server.ts
@@ -60,11 +60,10 @@ export async function run() {
   app.useWebSocketAdapter(adapter);
 
   const url = app.get(URLHelper);
-  const listeningHost = '0.0.0.0';
 
-  await app.listen(config.server.port, listeningHost);
+  await app.listen(config.server.port, config.server.host);
 
   logger.log(`AFFiNE Server is running in [${env.DEPLOYMENT_TYPE}] mode`);
-  logger.log(`Listening on http://${listeningHost}:${config.server.port}`);
+  logger.log(`Listening on http://${config.server.host}:${config.server.port}`);
   logger.log(`And the public server should be recognized as ${url.baseUrl}`);
 }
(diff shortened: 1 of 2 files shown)


fengmk2:
You cannot directly use the host configuration; it is recommended to add a new configuration called listenHost to achieve this. This will maintain compatibility.

forehalo (approved):

packages/backend/server/src/server.ts
```diff
@@ -60,11 +60,10 @@ export async function run() {
   app.useWebSocketAdapter(adapter);
 
   const url = app.get(URLHelper);
-  const listeningHost = '0.0.0.0';
 
-  await app.listen(config.server.port, listeningHost);
+  await app.listen(config.server.port, config.server.host);
```
fengmk2, line 64: After supporting multi-host deployment, this host configuration cannot currently be used directly as the listen host https://github.com/toeverything/AFFiNE/pull/12950

yifeiyin:
Thanks for the added context fengmk2. I feel like adding listenHost would adds little value and further complicates the existing multiple configs for multi-host. I am happy to contribute again if there's a strong use case for it.

Also the [original issue](https://github.com/toeverything/AFFiNE/issues/12882) can be closed since we are listening on all interfaces by default now.

Closing the PR here

PR #13181, opened 2025-07-11
feat(server): add cron job for session cleanup

darkskygit:
fix AI-338


fengmk2 (approved):

fengmk2:
It seems we need to check the CI, it has been failing continuously

PR #13225, opened 2025-07-15
fix(server): delay send mail if retry many times

darkskygit:
fix AF-2748

## Summary by CodeRabbit

* **New Features**
  * Improved mail sending job with adaptive retry delays based on elapsed time, enhancing reliability of email delivery.

* **Chores**
  * Updated job payload to include a start time for better retry management.
  * Added an internal delay utility to support asynchronous pause in processes.

### DIFF
--- a/packages/backend/server/src/base/utils/promise.ts
+++ b/packages/backend/server/src/base/utils/promise.ts
@@ -1,3 +1,5 @@
+import { setTimeout } from 'node:timers/promises';
+
 import { defer as rxjsDefer, retry } from 'rxjs';
 
 export class RetryablePromise<T> extends Promise<T> {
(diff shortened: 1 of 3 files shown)


packages/backend/server/src/base/utils/promise.ts
```diff
@@ -48,3 +48,7 @@ export function defer(dispose: () => Promise<void>) {
     [Symbol.asyncDispose]: dispose,
   };
 }
+
+export function sleep(ms: number): Promise<void> {
```
pengx17, line 54: use `setTimeout` from `node:timers/promises`?

pengx17 (approved):

packages/backend/server/src/base/utils/promise.ts
darkskygit: done

PR #13253, opened 2025-07-17
feat(server): improve transcript

darkskygit:
fix AF-2758
fix AF-2759

### DIFF
--- a/packages/backend/server/src/plugins/copilot/transcript/service.ts
+++ b/packages/backend/server/src/plugins/copilot/transcript/service.ts
@@ -59,7 +59,7 @@ export class CopilotTranscriptionService {
     return prompt?.optionalModels[hasAccess ? 1 : 0];
   }
 
-  async submitTranscriptionJob(
+  async submitJob(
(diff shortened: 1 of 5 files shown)


pengx17 (approved):

packages/backend/server/src/plugins/copilot/transcript/service.ts
```diff
@@ -49,6 +49,15 @@ export class CopilotTranscriptionService {
     private readonly providerFactory: CopilotProviderFactory
   ) {}
 
+  private async getModel(userId: string) {
+    const prompt = await this.prompt.get('Transcript audio');
+    const hasAccess = await this.models.userFeature.has(
+      userId,
+      'unlimited_copilot'
+    );
+    return prompt?.optionalModels[Number(hasAccess)];
```
pengx17: A bit hack here. Better add some comments for clearance

darkskygit: done

PR #13259, opened 2025-07-18
fix(editor): allow right click on reference

EYHN:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved click event handling on reference elements to prevent unintended behavior from right-clicks.


pengx17:
I think auxclick here is to open link in new tab

fengmk2 (approved):

PR #13348, opened 2025-07-29
feat(server): attachment embedding

darkskygit:
## Summary by CodeRabbit

* **New Features**
  * Added support for managing "blobs" in Copilot context, including adding and removing blobs via new GraphQL mutations and UI fields.
  * Introduced tracking and querying of blob embeddings within workspaces, enabling search and similarity matching for blob content.
  * Extended Copilot context and workspace APIs, schema, and UI to display and manage blobs alongside existing documents and files.

* **Bug Fixes**
  * Updated context and embedding status logic to handle blobs, ensuring accurate status reporting and embedding management.

* **Tests**
  * Added and updated test cases and snapshots to cover blob embedding insertion, matching, and removal scenarios.

* **Documentation**
  * Updated GraphQL schema and TypeScript types to reflect new blob-related fields and mutations.

* **Chores**
  * Refactored and cleaned up code to support new blob entity and embedding logic, including renaming and updating internal methods and types.


packages/backend/server/src/__tests__/copilot.spec.ts
```diff
@@ -1519,15 +1519,18 @@ test('TextStreamParser should process a sequence of message chunks', t => {
 });
 
 // ==================== context ====================
-test('should be able to manage context', async t => {
-  const { context, prompt, session, event, jobs, storage } = t.context;
+test.only('should be able to manage context', async t => {
```
fengmk2: should remove it

packages/backend/server/src/plugins/copilot/context/resolver.ts
```diff
@@ -130,38 +150,44 @@ export class CopilotContextType {
...
   @Field(() => SafeIntResolver)
   createdAt!: number;
 }
 
-registerEnumType(ContextEmbedStatus, { name: 'ContextEmbedStatus' });
-
 @ObjectType()
-class CopilotContextDoc extends CopilotDocType {
-  @Field(() => String, { nullable: true })
-  error!: string | null;
```
fengmk2, line 164: Does the `error` field maintain compatibility with a version? Otherwise, after the release of 0.24, if the 0.23 client hasn't been upgraded, this request will result in an error.

darkskygit: this field never used after 0.22

fengmk2 (approved):

PR #13429, opened 2025-08-06
chore(cli): remove legacy hooks and registration files, update depend…

ygcaicn:
This update migrates the CLI's TypeScript runtime from ts-node to tsx.

Key objectives:

- Eliminate Warnings: Completely resolves the departure warnings generated by ts-node.

- Improve Performance: tsx is built on esbuild, resulting in faster compilation and startup times.

- Simplify Configuration: Removed related legacy hook files to streamline the project.

## Summary by CodeRabbit

* **Chores**
  * Removed legacy CLI files and binaries, streamlining the command-line tool.
  * Updated the CLI to use a new execution method with `tsx` instead of `ts-node`.
  * Cleaned up package configuration by removing obsolete exports and dependencies.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -74,7 +74,7 @@ jobs:
         name: Wait for approval
         with:
           secret: ${{ secrets.GITHUB_TOKEN }}
-          approvers: forehalo,fengmk2
+          approvers: forehalo,fengmk2,darkskygit
           minimum-approvals: 1
           fail-on-denial: true
           issue-title: Please confirm to release docker image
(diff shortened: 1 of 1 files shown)


ygcaicn:
This looks more complicated than I thought. tsx runs into a series of problems when starting NestJS. Since it's based on esbuild, this is not easy to solve. Does anyone have any solutions?

How can we get rid of ts-node and resolve the large number of DeprecationWarnings?

> DeprecationWarning: fs.Stats constructor is deprecated.

I'm not sure if nestjs/cli can solve this.

PR #13458, opened 2025-08-10
fix(editor): fix import zip with cjk filename

### LINKED ISSUE #12721: [Bug]: When I import a Markdown ZIP package that includes media files, the Chinese filenames appear garbled.
### What happened?

Image

### Distribution version

macOS x64 (Intel)

### App Version

0.22

_No response_

### Are you self-hosting?

- [ ] Yes

### Self-hosting Version

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

ygcaicn:
fix #12721 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of non‑ASCII filenames when unzipping archives: the extractor now tries alternative encodings and validates results so filenames are preserved and displayed correctly after extraction. This change reduces corrupt or garbled names while keeping existing extraction behavior otherwise unchanged.

### DIFF
--- a/blocksuite/affine/widgets/linked-doc/src/transformers/utils.ts
+++ b/blocksuite/affine/widgets/linked-doc/src/transformers/utils.ts
@@ -65,6 +65,98 @@ export class Unzip {
     this.unzipped = fflate.unzipSync(new Uint8Array(await blob.arrayBuffer()));
   }
 
+  private fixFileNameEncoding(fileName: string): string {
+    try {
+      // check if contains non-ASCII characters
+      if (fileName.split('').some(char => char.charCodeAt(0) > 127)) {
+        // try different encodings
+        const fixedName = this.tryDifferentEncodings(fileName);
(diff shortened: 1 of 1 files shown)


blocksuite/affine/widgets/linked-doc/src/transformers/utils.ts
```diff
@@ -65,6 +65,98 @@ export class Unzip {
...
+
+      for (const encoding of encodings) {
+        try {
+          const decoder = new TextDecoder(encoding);
+          const result = decoder.decode(bytes);
+
+          // check if decoded result is valid
+          if (result && this.isValidDecodedString(result)) {
+            return result;
+          }
```
darkskygit, line 107:
```suggestion
          const encoded = new TextEncoder().encode(result);
          if (encoded.length === bytes.length && encoded.every((b, i) => b === bytes[i])) {
            return result;
          }
```

L-Sun (approved):

PR #13601, opened 2025-09-18
feat: improve model resolve

darkskygit:
fix AI-419


EYHN (approved):

packages/backend/server/src/plugins/copilot/session.ts
```diff
@@ -93,6 +100,50 @@ export class ChatSession implements AsyncDisposable {
...
+            plan: SubscriptionPlan.AI,
+          } as any);
+        isUserAIPro = subscription?.status === SubscriptionStatus.Active;
+      }
+    } catch {
+      // payment not available -> skip checks
+      paymentEnabled = false;
+    }
+
+    if (paymentEnabled) {
```
akumatus, line 133: No need for this `paymentEnabled` assertion, because `isUserAIPro` is false by default.

packages/backend/server/src/plugins/copilot/prompt/prompts.ts
```diff
@@ -2099,6 +2099,13 @@ Below is the user's query. Please respond in the user's preferred language witho
       'codeArtifact',
       'blobRead',
     ],
+    proModels: [
+      'gemini-2.5-pro',
+      'claude-opus-4@20250514',
+      'claude-sonnet-4@20250514',
+      'claude-3-7-sonnet@20250219',
+      'claude-3-5-sonnet-v2@20241022',
+    ],
```
akumatus, line 2108: Maybe it's better to use a `isProModels` function instead of `proModels` config in the prompt?

packages/backend/server/src/plugins/copilot/session.ts
```diff
@@ -93,6 +100,50 @@ export class ChatSession implements AsyncDisposable {
...
+      if (isUserAIPro) {
+        if (!requestedModelId) {
+          const firstPro = this.proModels[0];
+          return normalize(firstPro);
+        }
```
akumatus, line 138: No need for this `firstPro` logic.

packages/backend/server/src/plugins/copilot/prompt/prompts.ts
darkskygit: dont hardcode in code

packages/backend/server/src/plugins/copilot/session.ts
darkskygit: need to distinguish selfhost

packages/backend/server/src/plugins/copilot/session.ts
darkskygit: pro user use pro model if not provided

PR #13638, opened 2025-09-23
feat(component): emoji and icon picker

CatsJuice:
CleanShot 2025-09-23 at 17 11 13

## Summary by CodeRabbit

- **New Features**
  - Icon Picker added with Emoji and Icon panels, search/filtering, recent selections, color selection, skin tone options, and smooth group navigation.

- **Documentation**
  - Storybook example added to preview and test the Icon Picker.

- **Chores**
  - Bumped icon library dependency to a newer minor version.
  - Added emoji data dependency to support the Emoji Picker.


steffenrapp:
@CatsJuice it's really awesome to have an icon picker for docs and folders instead of setting the first character to an emoji! Is is also planned from your side to create some kind of migration that moves the first emoji to the new icon field so that emojis that have been set previously don't get "lost"? Right now in the canary version the emoji is displayed in the title instead of the icon field.

CatsJuice:
> @CatsJuice it's really awesome to have an icon picker for docs and folders instead of setting the first character to an emoji! Is is also planned from your side to create some kind of migration that moves the first emoji to the new icon field so that emojis that have been set previously don't get "lost"? Right now in the canary version the emoji is displayed in the title instead of the icon field.

Yes, we are currently developing a new way to set icons for doc/tag/collection/folder. However, due to technical reasons, data migration will not be implemented for now. This means the previously set icons will need to be reconfigured.

EYHN (approved):

PR #13670, opened 2025-09-29
feat(server): refresh subscription

darkskygit:
## Summary by CodeRabbit

* **New Features**
  * Added an on-demand mutation to refresh the current user's subscriptions, syncing with RevenueCat when applicable and handling Stripe-only cases.
  * Subscription variant normalization for clearer plan information and consistent results.

* **Tests**
  * Added tests for refresh behavior: empty state, RevenueCat-backed multi-step sync, and Stripe-only scenarios.

* **Client**
  * New client operation to invoke the refresh mutation and retrieve updated subscription fields.

### DIFF
--- a/packages/backend/server/src/plugins/payment/resolver.ts
+++ b/packages/backend/server/src/plugins/payment/resolver.ts
@@ -567,10 +567,21 @@ export class UserSubscriptionResolver {
       },
     });
 
+    const subscriptions = current.reduce(
+      (r, s) => {
+        if (s.plan in SubscriptionPlan) {
+          r[s.plan as SubscriptionPlan] = s.provider;
+        }
+        return r;
+      },
+      {} as Record<SubscriptionPlan, Provider>
+    );
+
     // has revenuecat subscription or no subscription at all
     const shouldSync =
       current.length === 0 ||
-      current.some(s => s.provider === Provider.revenuecat);
+      subscriptions.pro === Provider.revenuecat ||
+      subscriptions.ai === Provider.revenuecat;
 
     if (shouldSync) {
       try {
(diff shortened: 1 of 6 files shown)


L-Sun (approved):

packages/backend/server/src/plugins/payment/resolver.ts
```diff
@@ -534,6 +542,59 @@ export class UserSubscriptionResolver {
...
+            SubscriptionStatus.PastDue,
+          ],
+        },
+      },
+    });
+
+    // has revenuecat subscription or no subscription at all
+    const shouldSync =
+      current.length === 0 ||
+      current.some(s => s.provider === Provider.revenuecat);
```
EYHN: If the user has no revenuecat subscription, the subscription will not be synced?

What is the point of refreshing the subscription in this way?

darkskygit: stripe will also have subscription records; when subscribing with stripe, it should not go through iap
so only attempt to sync when:
1. an rcat record exists
2. no records

EYHN: I'm afraid some people will subscribe to Pro on Stripe and then subscribe to AI on IAP

EYHN: The easiest way is to remove this condition, if there is nothing bad in this sync

darkskygit: therefore, the frontend should check via `subscriptions.gql` whether a subscription already exists before triggering the iap
neither rcat nor the app store has a api capable of refusing charges

EYHN: @Lakr233 We should add this check into the paywall?

Lakr233: Ah, that’s a rather good point! Perhaps we could include eligible SKUs as a parameter when showing the paywall to the user. We could also hide or make some tabs invisible to the user.

Keep me updated when new API is available.

Issue #13703, opened 2025-10-07
[Bug]: Notion side user avatar becomes big pictures after Export + Import


xingjianan:
### What happened?

On my notion side, I have a lot of pages in an inlined database, and have a column which contains a list of Person(user).
 
After exporting to HTML on the Notion side, and then import to AFFiNE, the avatars became a lot of big pictures on the top of the migrated pages.

The exported HTML from Notion looks like as follow:
```
Person</th>
<td>
<span class="user"><img src="xxx.jpeg" class="icon user-icon"/>Name with avatar</span>
<span class="user"><span class="icon text-icon user-icon"><span class="user-icon-inner">C</span></span>CXXXX(Another name without avatar)</span>
<span class="user"><span class="icon text-icon user-icon"><span class="user-icon-inner">K</span></span>KXXXX(2nd name without avatar)</span>
```

I think the notion-html adapters can just replace any html node with "user-icon" class to a "space", hence the post replacement html will look like:
`<span class="user"> Name with avatar</span><span class="user"> CXXXX(Another name without avatar)</span><span class="user"> KXXXX(2nd name without avatar)</span>

### Distribution version

macOS x64 (Intel)

### Are you self-hosting?

- [x] Yes

### Relevant log output

```shell

```

PR #13718, opened 2025-10-09
fix: strip user-icon avatars from Notion HTML import (#13703)

ankitsharma38:
## Problem
When importing Notion HTML exports to AFFiNE, user avatars in database Person columns were being imported as large images at the top of pages instead of being properly handled.

## Solution
- Added `notionHtmlUserIconToDeltaMatcher` to specifically handle elements with `user-icon` class
- These elements are now replaced with spaces, cleaning up the imported content
- Added test case to verify the fix works correctly

## Example
Before: Large avatar images appeared at the top of imported pages
After: Clean text like "Name with avatar CXXXX(Another name without avatar) KXXXX(2nd name without avatar)"

## Testing
- Added unit test case that verifies user icons are properly converted to spaces
- Existing tests continue to pass

## Summary by CodeRabbit

- New Features
  - Notion HTML imports now skip user icon elements, inserting a space instead to prevent icon artifacts in text and produce cleaner paragraphs.
- Tests
  - Added a snapshot test covering user icon handling during Notion HTML conversion to ensure consistent results.
- Chores
  - Removed iOS Swift Package.resolved to avoid committing resolved dependency state and reduce churn.


darkskygit (changes requested):
dont change file mode and remove non-required file changes

PR #13849, opened 2025-11-03
Fix emoji icon overlap

vetash:
Fix icon on doc editor:

изображение

изображение

## Summary by CodeRabbit

* **Style**
  * Updated button component text overflow handling to allow alternative clipping behavior based on layout context.

### DIFF
--- a/packages/frontend/component/src/ui/button/button.css.ts
+++ b/packages/frontend/component/src/ui/button/button.css.ts
@@ -193,7 +193,6 @@ export const content = style({
   // in case that width is specified by parent and text is too long
   textOverflow: 'ellipsis',
   whiteSpace: 'nowrap',
-  overflow: 'hidden',
 });
 
 export const icon = style({


packages/frontend/component/src/ui/button/button.css.ts
```diff
@@ -193,7 +193,6 @@ export const content = style({
   // in case that width is specified by parent and text is too long
   textOverflow: 'ellipsis',
   whiteSpace: 'nowrap',
-  overflow: 'hidden',
```
CatsJuice, line 196: Since this is a very general, globally used component, directly removing it here wouldn’t be appropriate — it’s quite likely to affect the behavior of some buttons elsewhere. The root cause of the emoji issue probably isn’t here.

vetash: Ok, i'll try to find another way fo fix that

Issue #13966, opened 2025-11-16
[Bug]: When renaming a database, pressing Enter creates a database row


mwhouser:
### What happened?

For a Table View:

1. Click the name of the database to rename it.
2. Edit the name.
3. Press Enter.

It starts a new row. Instead, it should complete the rename.

### Distribution version

Windows x64

### App Version

0.25.4

### Are you self-hosting?

- [x] Yes

### Self-hosting Version

0.25.4

### Relevant log output

```shell

```

ddybing:
I've submitted a PR https://github.com/toeverything/AFFiNE/pull/13975 to fix.

ddybing:
The fix has been merged and is available in the canary channel. I assume it will be available in the next stable update.

PR #13975, opened 2025-11-17
fix(core): pressing ENTER on database title now switches focus instead of creating new record

ddybing:
Initial bug report: Issue https://github.com/toeverything/AFFiNE/issues/13966

Description of bug: When a database header/title is in focus and the user presses ENTER, a new record is created and shown to the user. 

Expected outcome: When the user presses enter in the header title field, the new title should be applied and then the title field should loose focus. 

Short summary of fix: When the ENTER key is pressed within the title, the `onPressEnterKey()` function is called. As of now, this calls the function `this.dataViewLogic.addRow?.('start');` which creates a new record. In this fix, this has been changed to `this.input.blur()` which instead essentially switches focus away from the title field and does not create a new record, as expected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Modified Enter key behavior in the database title field. Pressing Enter now blurs the input instead of automatically inserting a new row.

### DIFF
--- a/blocksuite/affine/blocks/database/src/components/title/index.ts
+++ b/blocksuite/affine/blocks/database/src/components/title/index.ts
@@ -176,7 +176,7 @@ export class DatabaseTitle extends SignalWatcher(
   private readonly isFocus$ = signal(false);
 
   private onPressEnterKey() {
-    this.dataViewLogic.addRow?.('start');
+    this.input.blur();
   }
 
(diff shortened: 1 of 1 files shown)


ddybing:
This is my first time contributing. I saw the Lint error and have tried to fix that now.

L-Sun (approved):

PR #13982, opened 2025-11-19
feat(core): find todo list based on 'checkbox' search query

ddybing:
This feature enhances the /slash command by allowing users to search for 'checkbox' and have the to-do list item show up as a result. Users come from different systems and environments, and some may use the name 'checkbox' but be confused as they cannot find it in the search menu. 

This is achieved by adding a `searchAlias`  property on the to-do list item block that contains the string `checkbox`. 

## Summary by CodeRabbit

* **New Features**
  * Added search-alias support for slash menu items so entries can be found by alternative terms.
  * To-do List entry now includes "checkbox" as an additional searchable alias to improve discoverability.
  * Slash menu search results updated to reflect alias-driven matches (additional item appears when searching).


darkskygit (approved):

darkskygit:
have type error

ddybing:
@darkskygit Sorry, should be fixed now. Had forgotten to add the searchAlias property to the TextConversionConfig interface.

darkskygit:
affect some test,need to update

ddybing:
Test is expecting 4 properties, but the object now has 5. Do we update the tests, or should I try to find another way to implement this feature?

darkskygit:
> Test is expecting 4 properties, but the object now has 5. Do we update the tests, or should I try to find another way to implement this feature?

update the test

ddybing:
I have updated e2e test to adjust for the increase in item properties. I will await the result from the build and test workflow.

darkskygit (approved):

ddybing:
I see there are still some errors with one of the tests. Shall I look into it @darkskygit ?

PR #13988, opened 2025-11-20
feat(server): make OIDC refresh_token field optional

preved911:
Hello. I faced with OIDC implementation, which doesn't return refresh_token, and it doesn't work with affine. This patch should solve the problem.

## Summary by CodeRabbit

* **Bug Fixes**
  * OIDC authentication now accepts tokens that may lack a refresh token, improving compatibility with providers that do not return refresh tokens and reducing sign-in errors for affected accounts.
  * Minor stability improvements to token handling to prevent unexpected authentication failures.


hayd1n:
I encountered the same issue when using [Synology SSO Server](https://www.synology.com/dsm/7.3/software_spec/sso_server) for OIDC authentication — the SSO Server doesn't return a `refresh_token`, which causes authentication to fail. I can confirm that this patch effectively resolves the problem. 👍

darkskygit:
the server side might need this field in future processes.

PR #14000, opened 2025-11-22
Fix: Disable GFM auto-linking for Windows-style file paths

S-A-Adit:
This PR prevents `remark-gfm` from auto-linking strings like `api\views\project.py`, which incorrectly turn into `<api\views\project.py>` when pasted into Affine.

### Key Changes
- Added a custom unified plugin to detect and convert backslash-containing auto-links into plain text nodes.
- Ensures Windows file paths are pasted and rendered correctly without unwanted angle brackets or link transformations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect autolinking of file paths in Markdown content, ensuring file path references are displayed as plain text rather than clickable links

### DIFF
--- a/blocksuite/affine/shared/src/adapters/markdown/markdown.ts
+++ b/blocksuite/affine/shared/src/adapters/markdown/markdown.ts
@@ -204,10 +204,40 @@ export class MarkdownAdapter extends BaseAdapter<Markdown> {
       .replace(/&#x20;\n/g, ' \n');
   }
 
-  private _markdownToAst(markdown: Markdown) {
+ private _markdownToAst(markdown: Markdown) {
     const processor = unified()
       .use(remarkParse)
       .use(remarkGfm)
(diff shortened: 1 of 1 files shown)


blocksuite/affine/shared/src/adapters/markdown/markdown.ts
```diff
@@ -204,10 +204,40 @@ export class MarkdownAdapter extends BaseAdapter<Markdown> {
...
+
+            for (let i = 0; i < node.children.length; i++) {
+              const child = node.children[i];
+
+              if (
+                child.type === 'link' &&
+                child.children.length === 1 &&
+                child.children[0].type === 'text' &&
+                child.url === child.children[0].value &&
+                (child.url as string).includes('\\')
```
darkskygit, line 228: prefer `typeof child.url === 'string'`

PR #14015, opened 2025-11-26
fix(editor): surface-block canvas size not fitting its container in edgeless mode when zoom is not 1

congzhou09:
### Problem
●In edgeless mode, the embed-edgeless-doc's content does not match the size of its outer block when zoom ≠ 1.
●The follwing image and video show the issue at zoom = 0.5.

图片

https://github.com/user-attachments/assets/ea7e7cc4-64ae-4747-8124-16c4eea6458e

### Reason and resolve
●The issue occurs because the surface-block canvas uses the container’s dimensions obtained from getBoundingClientRect(), which are already affected by the CSS transform. The canvas is then transformed again together with the container, causing the size mismatch.
●To keep all drawing operations in the surface-block’s original coordinate space, we apply a reverse transform to the canvas.

### After

https://github.com/user-attachments/assets/6c802b81-d520-44a0-9f01-78d0d60d37b8

## Summary by CodeRabbit

* **Bug Fixes**
  * Canvas rendering now properly responds to viewport zoom levels. Visual scaling is applied dynamically to ensure canvases align correctly with viewport scaling, providing consistent and accurate rendering during zoomed interactions while preserving original canvas dimensions.

### DIFF
--- a/blocksuite/affine/blocks/surface/src/renderer/canvas-renderer.ts
+++ b/blocksuite/affine/blocks/surface/src/renderer/canvas-renderer.ts
@@ -113,7 +113,7 @@ export class CanvasRenderer {
    * It is not recommended to set width and height to 100%.
    */
   private _canvasSizeUpdater(dpr = window.devicePixelRatio) {
-    const { width, height } = this.viewport;
+    const { width, height, viewScale } = this.viewport;
     const actualWidth = Math.ceil(width * dpr);
     const actualHeight = Math.ceil(height * dpr);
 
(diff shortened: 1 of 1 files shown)


congzhou09:
There is a momentary content-size mismatch in the "after" video during the first resize after changing the zoom. The cause has been identified as a separate bug, and I will submit a follow-up PR to address it.

图片

darkskygit (approved):

congzhou09:
> There is a momentary content-size mismatch in the "after" video during the first resize after changing the zoom. The cause has been identified as a separate bug, and I will submit a follow-up PR to address it.
> 图片

the follow-up PR: https://github.com/toeverything/AFFiNE/pull/14019

### After these two PR

https://github.com/user-attachments/assets/6a2592d3-e921-492a-9fd7-3d3e3a56bb5b

PR #14016, opened 2025-11-26
fix: improved user feedback for proxy file size limits

### LINKED ISSUE #13992: [Bug]: Attachment size limit applies incorrectly
### What happened?

Hello!

When I try to attach a file (a PDF) that is larger than 1 MB, AFFiNE shows an error "File size exceeds the 100 MB limit" which it is not. Compressing the file so it becomes smaller than 1 MB helps and AFFiNE allows that file to be uploaded. The problem is present in browser and in macOS desktop app.

Is there a bunch of zeroes missing in file size checking?

### Distribution version

macOS ARM 64 (Apple Silicon)

### App Version

0.25.5

Firefox

### Are you self-hosting?

- [x] Yes

### Self-hosting Version

0.25.5

### Relevant log output

```shell

```

### Anything else?

An example of the error and the real file size:
Image

ddybing:
**Summary:**
This PR improves the user feedback when encountering an HTTP 413 (_CONTENT_TOO_LARGE)_ error caused by a file size limit in the proxy / ingress controller in a self-hosted environment. 
fix #13992

**Example scenario:**
A self-hosted environment serves AFFiNE through an nginx proxy, and the `client_max_body_size` variable in the configuration file is set to a smaller size (e.g. 1MB) than AFFiNE's own file size limit (typically 100MB). Previously, the user would get an error saying the file is larger than 100MB regardless of file size, as all of these cases resulted in the same internal error. With this fix, the  _CONTENT_TOO_LARGE_ error is now handled separately and gives better feedback to the user that the failing upload is caused by a fault in the proxy configuration. 

**Screenshot of new error message**

1MB_now

**Affected files:**
  1. packages/common/nbstore/src/storage/errors/over-size.ts
  2. packages/common/nbstore/src/impls/cloud/blob.ts

I'm open to any suggestions in terms of the wording used in the message to the user. The fix has been tested with an nginx proxy.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved user-facing error messages for file upload failures. When an upload exceeds the file size limit, users now receive a clearer message indicating that the upload was stopped by the network proxy due to the size restriction, providing better understanding of why the upload was rejected.

### DIFF
--- a/packages/common/nbstore/src/impls/cloud/blob.ts
+++ b/packages/common/nbstore/src/impls/cloud/blob.ts
@@ -116,7 +116,10 @@ export class CloudBlobStorage extends BlobStorageBase {
         throw new OverSizeError(this.humanReadableBlobSizeLimitCache);
       }
       if (userFriendlyError.is('CONTENT_TOO_LARGE')) {
-        throw new OverSizeError(this.humanReadableBlobSizeLimitCache);
+        throw new OverSizeError(
+          null,
+          'Upload stopped by network proxy: file size exceeds the set limit.'
+        );
       }
       throw err;
     }
--- a/packages/common/nbstore/src/storage/errors/over-size.ts
+++ b/packages/common/nbstore/src/storage/errors/over-size.ts
@@ -1,6 +1,10 @@
 export class OverSizeError extends Error {
-  constructor(limit: string | null) {
-    const formattedLimit = limit ? `${limit} ` : '';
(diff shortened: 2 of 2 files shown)


ddybing:
Related to issue: https://github.com/toeverything/AFFiNE/issues/13992#issuecomment-3574257646

darkskygit (approved):

Issue #13290, opened 2025-07-22
[Bug]: Wrong shortcut for windows app.


foogao:
### What happened?

The shortcut of `Copy private link` is wrong for the windows version of AFFiNE desktop app, it is showing the key in macOS keyboard.

<img width="1402" height="946" alt="Image" src="https://github.com/user-attachments/assets/642247f4-971d-4158-a9a2-d06a383d5a1a" />

That's it.

### Distribution version

Windows x64

### App Version

0.23.2

### Are you self-hosting?

- [ ] Yes

### Relevant log output

```shell

```

ddybing:
This seems to be an issue even in 0.25.7. I'll look into it.

ddybing:
I have created PR https://github.com/toeverything/AFFiNE/pull/14088 to fix this.

PR #14088, opened 2025-12-11
fix(core): fixed keyboard shortcut help for Windows and Linux

ddybing:
This PR is related to issue https://github.com/toeverything/AFFiNE/issues/13290

Keyboard shortcut for copying a private link works as expected, but the overview of shortcuts shows the Mac shortcut for Windows, web and Linux users. This fix shows the correct (Ctrl+Shift+C) shortcut to the aforementioned users. 

I have not tested this on a Mac (neither in browser nor in the app), but ideally this should not have an impact for Mac users as the logic for showing the correct shortcut is already implemented.

Affected files: 
- packages/frontend/core/src/components/hooks/affine/use-shortcuts.ts

Old: 
old_shortcut

New:
Keyboard shortcut fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the keyboard shortcut for copying private links on Windows from Command+Shift+C to Ctrl+Shift+C.

### DIFF
--- a/packages/frontend/core/src/components/hooks/affine/use-shortcuts.ts
+++ b/packages/frontend/core/src/components/hooks/affine/use-shortcuts.ts
@@ -86,7 +86,7 @@ export const useWinGeneralKeyboardShortcuts = (): ShortcutMap => {
       [t('expandOrCollapseSidebar')]: ['Ctrl', '/'],
       [t('goBack')]: ['Ctrl', '['],
       [t('goForward')]: ['Ctrl', ']'],
-      [t('copy-private-link')]: ['⌘', '⇧', 'C'],
+      [t('copy-private-link')]: ['Ctrl', '⇧', 'C'],
     }),
     [t]
   );


ddybing:
@darkskygit

darkskygit:
> @darkskygit

The ⌘(Command) and Ctrl keys on the keyboard are not the same key; they need to be displayed separately on different systems
image

ddybing:
> > @darkskygit
> 
> The ⌘(Command) and Ctrl keys on the keyboard are not the same key; they need to be displayed separately on different systems


Hi. There is one set of shortcuts for Windows and another for Mac. I've only updated the one for Windows, as it contained a Mac key. I don't have a Mac to test with but doesn't this switch automatically to show Mac shortcuts when run on a Mac?

darkskygit:
> > > @darkskygit
> > 
> > 
> > The ⌘(Command) and Ctrl keys on the keyboard are not the same key; they need to be displayed separately on different systems
> 
> Hi. There is one set of shortcuts for Windows and another for Mac. I've only updated the one for Windows, as it contained a Mac key. I don't have a Mac to test with but doesn't this switch automatically to show Mac shortcuts when run on a Mac?

sry, i hadn't expand the code outside the diff before; now the pr lgtm

darkskygit (approved):

Issue #13962, opened 2025-11-15
[Feature Request]: Properly import dates from Notion


mwhouser:
### Description

When importing databases from Notion, dates are imported as Text. They should be imported as dates.

Just changing the field from Text to Date erases all entries in that column.

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

ddybing:
I'll look into it and submit a PR if I get something working. Right now there is no logic in the import process that looks for date fields. Also, date fields exported from Notion are not formatted the same way as other fields such as multi-select, so some extensions to the logic will be needed.

PR #14111, opened 2025-12-15
feat(core): adapt date fields in database for notion import

ddybing:
This is related to issue/feature request https://github.com/toeverything/AFFiNE/issues/13962. 

This PR extends the Notion import functionality to properly handle date fields from databases. Previously, these were imported as text (see photo below), which served little purpose. These Notion date fields are now parsed as actual dates, and imported to AFFiNE as epoch time (which is what the date field in AFFiNe expects). Because of this, even date fields with time (e.g. 09:00 AM) are also handled correctly - although they are only shown as dates, since AFFiNE's `Date` field does not support time.

Tested with several Notion imports both with and without time, and they all seem to work correctly. 

Affected files: 
- blocksuite/affine/blocks/database/src/adapters/notion-html.ts

Old: 
image

New: 
image

## Summary by CodeRabbit

* **New Features**
  * Enhanced Notion imports with automatic date column detection. When importing Notion databases, date fields are now automatically recognized, properly configured as date columns, and formatted correctly. This improvement ensures accurate data preservation, eliminates manual type corrections, and provides a streamlined import experience for all users working with date-rich Notion databases.

### DIFF
--- a/blocksuite/affine/blocks/database/src/adapters/notion-html.ts
+++ b/blocksuite/affine/blocks/database/src/adapters/notion-html.ts
@@ -15,6 +15,7 @@ const ColumnClassMap: Record<string, string> = {
   typesCheckbox: 'checkbox',
   typesText: 'rich-text',
   typesTitle: 'title',
+  typesDate: 'date',
 };
 
 const NotionDatabaseToken = '.collection-content';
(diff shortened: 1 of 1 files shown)


ddybing:
Side note: We could "support" time in dates by importing date fields from Notion with time included as "Created Time" fields instead, but I've omitted that as now as it's not really the same.

darkskygit (approved):

Issue #14094, opened 2025-12-12
[Bug]: Journal > Pick Day > No autofocus (don't force me to use mouse click, just let me type..)


grand-lotus-iroh:
### What happened?

Can you make it so when I click Create Daily Journal it just autofocuses the input field so I don't have to left click?

<img width="763" height="534" alt="Image" src="https://github.com/user-attachments/assets/20fc1bd2-dba3-44e0-a99b-0506801fc951" />

As soon as I click that blue button above I would like to be able to type without clicking again.

### Distribution version

Windows x64

### Are you self-hosting?

- [ ] Yes

### Relevant log output

```shell

```

ddybing:
I submitted a PR which should do just this. It puts focus on the last block in the journal when opening it. If the journal is empty this means the first paragraph, if there is already text in the document this creates a new block right under the last one and puts focus there.

PR #14122, opened 2025-12-18
feat(core): focus on text body when opening journal

ddybing:
Related to issue https://github.com/toeverything/AFFiNE/issues/14094

This PR makes it so that focus is put on the input body when loading a journal. A check is made when loading the document whether it is a normal document or a journal document. If it is a journal document, the last noteblock in the document is focused on. This does not change how the title is focused on normal documents. 

This makes it more effortless to use the journal, as you don't have to click on the body of the journal after opening/creating it. 

## Summary by CodeRabbit

* **New Features**
  * Improved editor focus for journal documents: when opening or switching to a journal the cursor now auto-positions to the end of the last note entry (or the input area) after a short, smooth delay for faster typing and reliable focus behavior.

* **Bug Fixes**
  * Added safeguards and error handling to make automatic focus more robust across load and editor states.

### DIFF
--- a/packages/frontend/core/src/desktop/pages/workspace/detail-page/detail-page.tsx
+++ b/packages/frontend/core/src/desktop/pages/workspace/detail-page/detail-page.tsx
@@ -188,13 +188,7 @@ const DetailPageImpl = memo(function DetailPageImpl() {
   useRegisterBlocksuiteEditorCommands(editor, isActiveView);
 
   const journalService = useService(JournalService);
-
-  const isJournal = useCallback(
-    (docId: string): boolean => {
-      return !!journalService.journalDate$(docId).value;
-    },
-    [journalService]
-  );
+  const isJournal = !!useLiveData(journalService.journalDate$(doc.id));
 
   const onLoad = useCallback(
     (editorContainer: AffineEditorContainer) => {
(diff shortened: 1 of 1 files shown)


ddybing:
I need to optimise this further by wrapping isJournal in a callback. I'll fix it when I'm back on my computer and push a new commit.

packages/frontend/core/src/desktop/pages/workspace/detail-page/detail-page.tsx
```diff
@@ -184,10 +187,43 @@ const DetailPageImpl = memo(function DetailPageImpl() {
...
   useRegisterBlocksuiteEditorCommands(editor, isActiveView);
 
+  const journalService = useService(JournalService);
+
+  const isJournal = useCallback(
+    (docId: string): boolean => {
+      return !!journalService.journalDate$(docId).value;
+    },
+    [journalService]
+  );
```
darkskygit: const isJournal = !!useLiveData(journalService.journalDate$(doc.id));

ddybing: Yes, that's what I was thinking.

ddybing: Oh wait, yes, seems like I had already implemented it. I had an issue with a commit yesterday that wouldn't finish properly after linting, but seems like the change I did was added after all. I thought I had to rewrite it and commit once more. So if it looks good to you @darkskygit, I think we are good to go.

darkskygit: use `useLiveData` instead of `useCallback`

ddybing: @darkskygit Check new commit, updated to use `useLiveData` instead of `useCallBack`

darkskygit (approved):

PR #14211, opened 2026-01-05
feat(server): add LISTEN_ADDR env var for allowing server to listen on ipv6

yiding:
The old code hardcoded 0.0.0.0 which means the server only listened for ipv4 connections, making it not work on ipv6-only networks.

This change adds a LISTEN_ADDR env var which allows the server to bind to ipv6 as well.

## Summary by CodeRabbit

* **New Features**
  * Server listen address is now configurable via the LISTEN_ADDR environment variable (default: 0.0.0.0), enabling IPv4/IPv6 or interface-specific binding.
  * Configuration schemas and admin UI now expose the listen address option so deployments can view and override it.

### DIFF
--- a/.docker/selfhost/schema.json
+++ b/.docker/selfhost/schema.json
@@ -595,6 +595,11 @@
           "description": "Multiple hosts the server will accept requests from.\n@default []",
           "default": []
         },
+        "listenAddr": {
+          "type": "string",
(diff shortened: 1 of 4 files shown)


darkskygit:
please use `yarn affine server genconfig` to sync the schema

darkskygit (approved):

PR #14244, opened 2026-01-10
feat(editor): add shift-based angle snapping for line drawing and anchor dragging

S-A-Adit:
This PR adds ⇧ Shift angle snapping when drawing new lines/arrows and when dragging existing anchors.
While Shift is held, drag directions snap to 15° increments (matching tldraw behavior). Releasing Shift restores freeform movement.
This improves precision and consistency when creating and editing linear geometry.

## Summary by CodeRabbit

* **New Features**
  * Shift+drag snapping: holding Shift snaps drag, rotate, and resize to 15° increments during movement and on completion, preserving distance from the start point.

* **Tests**
  * Added unit tests covering no-snap, distance preservation, and snaps at 0°, 45°, 90°, 180° and near -15°.

* **Chores**
  * Pointer package exports updated to expose the new snapping capability.


S-A-Adit:
Hi @darkskygit could you please take a look and let me know if this works as expected?
Thanks!

pickaxe828:
Does this behave as intended on your side?

Cuz when I tried to press ⇧ Shift when creating/ modify an arrow (tried all three types)
It has no snapping effect at all and just follows my mouse pointer
(I ran the code without server)

In [here](https://github.com/toeverything/AFFiNE/pull/14244/changes#diff-8071ae74165e9e617476929ef8b406ff34be8cf56696a3a57129562a42dd9038R400), and with adding some `console.log`, 
I find that `onDragEnd` and `onDragMove` are only called upon dragging of a whole component on the canva, 
But not when anchors of an arrow are modified

So this doesn't seem to be the correct place to put `snapDragAngle` in, 
And hence might not be a correct implementation

(Or maybe this is because I ran the client without server, tho that would change such behavior is unlikely)

darkskygit:
The current code has type errors and cannot pass the unit tests.

Please modify the code locally and ensure that `yarn test` / `yarn lint` / `yarn typecheck` run without errors. 

Frequently syncing the main branch does not help move this branch toward being merged into the main branch.

PR #14329, opened 2026-01-27
fix(mobile): fixed toolbar position

passabilities:
## Summary
Fixed the text formatting toolbar not working properly on mobile web browsers.

## Problem
The toolbar had multiple issues on mobile devices:
- It would render off-screen or be covered by the virtual keyboard
- The flag-based rendering system caused visibility issues on mobile
- Long-press text selection didn't trigger the toolbar
- Wide toolbars could overflow the viewport

## Solution
- Use fixed positioning at bottom of screen on mobile devices
- Position toolbar above virtual keyboard using Visual Viewport API
- Handle toolbar visibility directly via `selectionchange` event
- Bypass flag-based rendering system on mobile to avoid rendering issues
- Add `touchend` listener to handle long-press text selection
- Limit toolbar max-width to viewport minus padding
- Enable horizontal scrolling for overflow content

## Test plan
- [x] Tested on mobile Safari (iOS)
- [x] Tested on mobile Chrome (Android)
- [x] Verified desktop browsers still work correctly
- [x] Verified the toolbar is fixed to the bottom of the screen and above virtual keyboard
- [x] Verified long-press text selection triggers toolbar

## Summary by CodeRabbit

* **Improvements**
  * Mobile toolbar now anchors to the bottom, adapts width, and repositions dynamically to stay above on-screen keyboards.
  * Toolbar visibility is context-aware, showing when native-like text selections occur and hiding otherwise; touch interactions are handled for reliable toggling.
  * Desktop experience and public APIs remain unchanged.

### DIFF
--- a/blocksuite/affine/widgets/toolbar/src/toolbar.ts
+++ b/blocksuite/affine/widgets/toolbar/src/toolbar.ts
@@ -280,8 +280,9 @@ export class AffineToolbarWidget extends WidgetComponent {
         zIndex: 'var(--affine-z-index-popover, 1000)',
         display: 'none',
         opacity: '0',
-        width: 'max-content',
-        touchAction: 'none',
+        maxWidth: 'calc(100vw - 32px)',
+        overflowX: 'auto',
+        touchAction: 'pan-x',
       });
       document.body.append(toolbar);
       disposables.add(() => toolbar.remove());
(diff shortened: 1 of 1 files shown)


passabilities:
I've updated the description to include screenshots of the problem I'm having on mobile

darkskygit (approved):

PR #14458, opened 2026-02-16
fix: electron app unused blobs not accurate

### LINKED ISSUE #14191: [Bug]: unused blobs not accurate
### What happened?

windows version 0.25.7

Image

showed some unused blobs. but these blobs were actually in use. 

online version seems good.

Image

i deleted unused blobs, and now i lost my figures and cannot find them back.

it is a shared workspace. in the other user who create the page, these images still exist. but after I clean the unused blobs, i cannot get the image back in the online version, or my local version. Maybe we need a "force" sync button to force upload the contents again?

### Distribution version

Windows x64

### App Version

0.25.7

_No response_

### Are you self-hosting?

- [x] Yes

### Self-hosting Version

0.25.1 stable-5c0e3b8

### Relevant log output

```shell

```

### Anything else?

_No response_

kthreadd:
fix #14191

In the Electron app, `fts_flush_index` must be called after inserting a blob (e.g., image) into a document, as there is no better opportunity to update the index in the database.

Failure to write the blob index to the database will result in incorrect information about unused blobs.

Note that `fts_flush_index` in current memory_indexer version only succeeds right after a blob insertion.

## Summary by CodeRabbit

* **Refactor**
  * Adjusted SQLite indexer refresh interval behavior to use default settings instead of a custom configuration.

### DIFF
--- a/packages/frontend/apps/electron-renderer/src/app/effects/events.ts
+++ b/packages/frontend/apps/electron-renderer/src/app/effects/events.ts
@@ -1,4 +1,3 @@
-import { DesktopApiService } from '@affine/core/modules/desktop-api';
 import { WorkspaceDialogService } from '@affine/core/modules/dialogs';
 import type { SettingTab } from '@affine/core/modules/dialogs/constant';
 import { DocsService } from '@affine/core/modules/doc';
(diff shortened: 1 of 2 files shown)


darkskygit:
I don’t think this is the root cause. The only function of fts_flush_index is to flush the modified index to disk; before the next restart of client, reads and writes always operate on the in-memory index structures.

kthreadd:
Hi @darkskygit 

I also think this is not the root cause, but I'm investigating why `fts_flush_index` fails to persist modified indexes to disk.

My observations:
Every user operation triggers a new indexing job in `subscribeDocUpdate`, following this flow:

```plaintext
tryNativeCrawlDocData
    |
deleteByQuery
    |
insert (ftsAddDocument is called per block; new indexes are only stored in memory)
    |
refreshIfNeed (30s interval in Electron—this rarely triggers actual index flushing to disk)
```

Even after waiting for the 30s timer to expire, operating on the same unflushed document still results in `fts_flush_index` not working as expected.

The new index is only written to disk if and only if `fts_flush_index` is explicitly called right after the user operation.

Based on these findings, I believe the current patch serves as a temporary fix for the issue, and the root problem likely resides in the memory_indexer.

PR #14513, opened 2026-02-24
[helm chart] Support external secrets for Postgres, plus other small changes

postmaxin:
I am deploying AFFiNE to a cluster where I use the [Postgres Operator](https://github.com/zalando/postgres-operator) to deploy my databases. Both this operator, and the AFFiNE chart, are opinionated about the format of the Postgres secret. This change both adds the Postgres username to the secret, and allows you to choose the field names for username/password.

I then use chart values like the following to deploy AFFiNE:

```yaml
global:
  database:
    secretName: "affine-website-im-owner-user.db.credentials.postgresql.acid.zalan.do"
    secretUsernameField: "username"
    secretPasswordField: "password"
    host: db.mori.svc
    name: affine_website_im
    user: affine_website_im
```

I also have made the following changes to the chart, in order to deploy it to my cluster (self-hosted K3S):

* New value `global.app.environment`: The current chart uses the K8S namespace as the name of the AFFiNE environment by default. `global.app.environment` lets you override this, so I don't have to place my AFFiNE in a namespace called "production".

* Doesn't crash if indexer is not configured

* Removed `global.indexer.username`/`password` as they are not used by the code

* Removed references to enabling/disabling Redis as AFFiNE will crash if there is no Redis

* Created a sub-chart, "common", to contain things that were previously being copy-pasted across three charts

## Summary by CodeRabbit

* **New Features**
  * Added a shared chart delivering reusable environment and service configuration templates.
  * Added centralized templates for Postgres, Redis, and Indexer environment setup.

* **Chores**
  * Refactored deployments to use the new reusable env templates.
  * Added configurable deployment environment variable.
  * Switched Indexer auth to API-key mode with an explicit enable flag.
  * Made database secret handling configurable (secret name and field names).
* **Dependencies**
  * Frontend and GraphQL charts now include the shared chart as a dependency.

### DIFF
--- a/.github/helm/affine/charts/common/Chart.yaml
+++ b/.github/helm/affine/charts/common/Chart.yaml
@@ -1,0 +1,6 @@
+apiVersion: v2
+name: common
+description: templates and helpers common to the affine chart and its sub-charts
+type: application
+version: 0.0.0
+appVersion: "0.26.1"
--- a/.github/helm/affine/charts/common/Chart.yaml
+++ b/.github/helm/affine/charts/common/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: common
 description: templates and helpers common to the affine chart and its sub-charts
-type: application
+type: library
 version: 0.0.0
 appVersion: "0.26.1"
--- a/.github/helm/affine/charts/common/templates/_helpers.tpl
+++ b/.github/helm/affine/charts/common/templates/_helpers.tpl
@@ -1,0 +1,3 @@
+{{- define "common.environment" -}}
+{{- .Values.global.app.environment | default .Release.Namespace -}}
+{{- end -}}
--- a/.github/helm/affine/charts/common/templates/_indexer_env.yaml
+++ b/.github/helm/affine/charts/common/templates/_indexer_env.yaml
@@ -14,4 +14,3 @@
       key: indexer-apiKey
 {{- end }}
 {{- end }}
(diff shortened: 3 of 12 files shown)


darkskygit:
the helm charts in this repo is  design for internal testing only
there are no plans to support other deployment needs
please maintain the helm charts yourself

postmaxin:
> the helm profile in this report is for internal testing only there are no plans to support other deployment needs please maintain the helm profiles yourself

Reading discussion #6325 , it sounded like people wanted to use this helm chart for release instead of https://github.com/toeverything/helm-charts .

Why don't you want the chart to work for real deployments? Wouldn't that make it easier for people to use AFFiNE?

darkskygit:
> > the helm profile in this report is for internal testing only there are no plans to support other deployment needs please maintain the helm profiles yourself
> 
> Reading discussion #6325 , it sounded like people wanted to use this helm chart for release instead of https://github.com/toeverything/helm-charts .
> 
> Why don't you want the chart to work for real deployments? Wouldn't that make it easier for people to use AFFiNE?

at present, we do not plan to maintain an official helm chart

the helm charts in this repo are only for automated testing and the default value and scale are not suitable for production deployments

currently, we officially recommend using Docker for deployment

if you have sufficient operations and maintenance experience, you can maintain helm charts yourself

postmaxin:
Alright, I've copied this work into it's own repository.

PR #14577, opened 2026-03-05
fix(editor): note-edgeless-block dimensions mismatch content at non-100% scale

congzhou09:
### Problem
●In edgeless mode, when the `note-edgeless-block` is scaled below 100%, its outer dimension becomes larger than its content region. This extra invisible region will block some user interactions such as clicks and hovers on editing elements underneath.

note-elem-block-click

https://github.com/user-attachments/assets/3b719b25-0d7e-496b-9507-6aa65ed0a797

### Solution
●The root cause is that `transform: scale(...)` CSS property (which implements the scale) is currently applyed to its **inner root element** instead of itself, and the solution is to move this CSS property to the proper place.

### After

https://github.com/user-attachments/assets/e2dbd75d-c2ea-460d-90a1-5cc13e12d5b8

## Summary by CodeRabbit

* **Refactor**
  * Centralized CSS scaling for graphics and edgeless note blocks into dedicated public methods; rendering now uses these methods instead of inline transform calculations.
* **Tests**
  * Updated end-to-end checks to read scale directly from the edgeless note element and use a more flexible transform-matching pattern.

### DIFF
--- a/blocksuite/affine/blocks/note/src/note-edgeless-block.ts
+++ b/blocksuite/affine/blocks/note/src/note-edgeless-block.ts
@@ -221,6 +221,12 @@ export class EdgelessNoteBlockComponent extends toGfxBlockComponent(
     }
   }
 
+  override getCSSScaleVal(): number {
+    const baseScale = super.getCSSScaleVal();
+    const extraScale = this.model.props.edgeless?.scale ?? 1;
+    return baseScale * extraScale;
+  }
+
   override getRenderingRect() {
     const { xywh, edgeless } = this.model.props;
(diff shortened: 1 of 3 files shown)


darkskygit (approved):

darkskygit:
some test failed, need to fix

congzhou09:
> some test failed, need to fix


e2e-failed

Fixed all failed e2e tests in <e2e/edgeless/note/scale.spec.ts>
- Changed the target element for retrieving the `scale` style property.
- Updated the regex pattern for parsing the scale value.

darkskygit (approved):

PR #14602, opened 2026-03-08
fix(security): run Docker container as non-root user

### LINKED ISSUE #14584: [Bug]: Docker image runs as root
### What happened?

As advised from the Affine team - recreating as an issue for review rather than a security issue.

**Summary**
It appears Affine currently runs as root within its docker container. This could lead to privilege escalation attacks and if the attacker was able to get out of the container, they could breach other containers and potentially the host.

**Details**
The default container needs to write to high-privilege directories within the container. This appears to be hardcoded. This means that unless the sysadmin customised a container build, Affine would need to operate as a privileged/root user. I can see it tried to write to:
/root/.affine/
/app/src/

**PoC**
This seems to be present in the default images/builds for Affine.

It can be manually mitigated using a Dockerfile which gives a non-root user access to the hardcoded locations in the container. But this is not ideal.

Recommend hardcoded high-privilege directories are changed to lower (user) locations instead in the default image/build.

Here is the Dockerfile example to mitigate the issue:

```
FROM ghcr.io/toeverything/affine:stable

# match your host user if you want; 1000:1000 is the usual default
ARG APP_UID=1000
ARG APP_GID=1000

RUN set -eux; \
    groupadd -g "${APP_GID}" affine || true; \
    useradd -m -u "${APP_UID}" -g "${APP_GID}" -s /usr/sbin/nologin affine || true; \
    mkdir -p /home/affine/.affine/storage /home/affine/.affine/config; \
    chown -R "${APP_UID}:${APP_GID}" /home/affine; \
    chown "${APP_UID}:${APP_GID}" /app/src/schema.gql

ENV HOME=/home/affine
USER ${APP_UID}:${APP_GID}
```

and example Docker compose

```
services:
  affine:
    build:
      context: .
      args:
        APP_UID: 1000
        APP_GID: 1000
    user: "1000:1000"
    volumes:
      - ./storage:/home/affine/.affine/storage
      - ./config:/home/affine/.affine/config

  affine_migration:
    build:
      context: .
      args:
        APP_UID: 1000
        APP_GID: 1000
    user: "1000:1000"
    volumes:
      - ./storage:/home/affine/.affine/storage
      - ./config:/home/affine/.affine/config
```

Noting that the host should also have directories limited for the binds: (example)

```
sudo chown -R 1000:1000 ./storage ./config
```

**Impact**
Privilege escalation attack, container breach as privileged user.

The severity is relatively high but the actual risk of it happening is very low.

### Distribution version

Linux

### App Version

_No response_

_No response_

### Are you self-hosting?

- [x] Yes

### Self-hosting Version

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

JasonOA888:
Fixes #14584

Add non-root user to Docker container for security.

- Create affine user with configurable UID/GID
- Set proper ownership
- Run as non-root user

## Summary by CodeRabbit

* **Chores**
  * Application process in the container now runs as a non-root user.
* **Configuration**
  * Default host mount paths for storage and configuration updated from /root/.affine to /home/affine/.affine.

### DIFF
--- a/.docker/selfhost/compose.yml
+++ b/.docker/selfhost/compose.yml
@@ -14,8 +14,8 @@ services:
         condition: service_completed_successfully
     volumes:
       # custom configurations
-      - ${UPLOAD_LOCATION}:/root/.affine/storage
-      - ${CONFIG_LOCATION}:/root/.affine/config
+      - ${UPLOAD_LOCATION}:/home/affine/.affine/storage
+      - ${CONFIG_LOCATION}:/home/affine/.affine/config
     env_file:
       - .env
     environment:
@@ -29,8 +29,8 @@ services:
     container_name: affine_migration_job
     volumes:
       # custom configurations
-      - ${UPLOAD_LOCATION}:/root/.affine/storage
-      - ${CONFIG_LOCATION}:/root/.affine/config
+      - ${UPLOAD_LOCATION}:/home/affine/.affine/storage
+      - ${CONFIG_LOCATION}:/home/affine/.affine/config
     command: ['sh', '-c', 'node ./scripts/self-host-predeploy.js']
     env_file:
       - .env
--- a/.github/deployment/node/Dockerfile
+++ b/.github/deployment/node/Dockerfile
@@ -27,9 +27,19 @@ RUN apt-get update && \
   apt-get install -y --no-install-recommends openssl libjemalloc2 && \
   rm -rf /var/lib/apt/lists/*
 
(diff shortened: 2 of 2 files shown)


uZer:
Hi!
Could we have a bit more info about why this PR has been closed?

Containers running with root or privileged user is a big security concern, especially in our current context where supply chain attacks are frequent. I'd like to try and help mitigate the issue if I can, but need more info about the issues with this PR :)

Cheers,

PR #14654, opened 2026-03-14
feat: resend workspace invite with backoff

ibex088:
## Summary

This PR adds a **Resend invite** action for pending workspace members and prevents duplicate resend emails by **deduplicating invitation sends in the notification queue** per workspace and invitee email.

https://www.loom.com/share/ab0e2ca3c7dc401cae2a2039bed2729b

### Backend
- Added `resendMemberInvite(workspaceId, inviteId): Boolean!`.
- Reused the existing invitation email pipeline via `notification.sendInvitation`.
- Added queue-based deduplication for invitation sends using a deterministic per-workspace, per-email job id.
- Resend now only succeeds for **Pending** invites.
- If an older queued invite exists for the same email and a new invite replaces it, the stale queued job is removed so the latest invite can be sent.
- Added backend test coverage for resend behavior, duplicate suppression, replacement invites, and non-pending rejection.

### Frontend
- Added a **Resend invite** action for pending members in member options.
- Shows:
  - success notification when resend is queued
  - **Invitation already queued** notification when a duplicate resend is blocked
  - localized error when the member has no email
- Added a component test covering the **already queued** notification path.
- Added new i18n strings and regenerated i18n typings.

## Behavior

- Admins/owners can resend invites for pending members.
- Duplicate resend attempts for the same workspace/email are blocked while an invitation send is already queued.
- Resend is rejected for non-pending invite states.
- If an invite is revoked/replaced before the queued send runs, the queue is updated so the latest invite is the one that gets emailed.

## Why

Invite emails can fail or be missed, so admins need a retry path. At the same time, resend should not enqueue duplicate invitation emails for the same recipient. This PR adds a simple recovery flow while using queue-level deduplication to guard against duplicate sends.

## Tests

- Backend e2e coverage for:
  - resend for pending member
  - duplicate resend suppression
  - revoke + re-invite replacement behavior
  - reject resend for non-pending team invite
- Frontend component coverage for:
  - **Invitation already queued** notification

## Summary by CodeRabbit

* **New Features**
  * Workspace admins/owners can resend pending member invites; shows success or “already queued” feedback and blocks resends if no email.

* **Behavior**
  * Server now prevents duplicate invitation notifications and ensures replaced/revoked invites produce notifications for the latest invite.

* **Localization**
  * Added UI strings for resend action, cooldown hint, success, and queued notifications.

* **Tests**
  * E2E and component tests cover resend, cooldown/prevention, revoke/replace, and error cases.


packages/backend/server/src/core/workspaces/resolvers/member.ts
```diff
@@ -234,6 +285,95 @@ export class WorkspaceMemberResolver {
...
+
+    await this.cache.set(
+      key,
+      {
+        attempt: nextAttempt,
+        nextAllowedAt: nextAllowedAt.toISOString(),
+        lastSentAt: now.toISOString(),
+      },
+      { ttl: MEMBER_RESEND_INVITE_BACKOFF_TTL_MS }
+    );
```
darkskygit: I think manually managing backoff via the cache is not a good idea.
Please consider managing the sending queue through a queue component, and only initiate a new sending event when the corresponding email address does not exist in the queue.

ibex088: https://github.com/toeverything/AFFiNE/pull/14654/commits/9eb680036d35c8460441acd98362f835c6b64b8a

ibex088: hope this is what you meant

PR #14665, opened 2026-03-15
fix(editor): restore horizontal scrolling in kanban board full-width layout

### LINKED ISSUE #14531: [Bug]: Kanban board is broken
### What happened?

I have a Kanban board that was working fine in previous versions.

But after upgrading to 0.26.x, there are many things about Kanban that just aren't working anymore.

* Horizontal scrolling is broken. Using the mouse to scroll left/right doesn't work at all anymore, and the scrollbar is simply gone.
* Dragging a card from one stack to another resets the stack order (ie. the stacks/lists reorder themselves)
* Dragging a card up/down within a stack simply does not work anymore.

This is using version 0.26.3 of the windows client, and also web interface of self-hosting using 0.26.3.

### Distribution version

Windows x64

### App Version

0.26.3

Firefox

### Are you self-hosting?

- [x] Yes

### Self-hosting Version

0.26.3

### Relevant log output

```shell

```

### Anything else?

_No response_

CTDave001:
## Summary

Fixes the horizontal scrolling issue in the kanban board when using full-width page layout. The scrollbar wasn't showing up because the kanban container was expanding to match its content width instead of staying constrained to the viewport.

Closes #14531

## Problem

In full-width mode, `virtualPadding` is `0`, which meant the negative-margin/padding technique used for scroll alignment was essentially a no-op. Without any width constraint, the `KanbanViewUI` element would grow to fit all its child groups, so `scrollWidth === clientWidth` and no scrollbar appeared.

## Fix

- Added `contain: inline-size` to the kanban view element so its width comes from the parent, not its children
- Swapped out inline padding on the scroll container for spacer divs, since padding inflates `clientWidth` while spacer divs only add to `scrollWidth`

## Files changed

- `blocksuite/affine/data-view/src/view-presets/kanban/pc/kanban-view-ui-logic.ts` — the actual fix
- `blocksuite/affine/data-view/src/core/group-by/trait.ts` — removed stray comment
- `blocksuite/affine/data-view/src/__tests__/kanban.unit.spec.ts` — added test for scroll containment

## Testing

- Added a unit test that checks `KanbanViewUI` applies `contain: inline-size` on connect
- All 41 tests pass (40 existing + 1 new)

**To verify manually:**
1. Create a database with kanban view
2. Set page to full-width layout
3. Add groups until they go past the viewport
4. Confirm horizontal scrollbar shows up and works

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite validating Kanban view UI styling, containment behavior, and layout properties with mock setup and DOM assertions

* **Style**
  * Enhanced horizontal spacing implementation in Kanban view component with improved layout approach

* **Chores**
  * Cleaned up test comments from internal utility code

### DIFF
--- a/blocksuite/affine/data-view/src/__tests__/kanban.unit.spec.ts
+++ b/blocksuite/affine/data-view/src/__tests__/kanban.unit.spec.ts
@@ -25,6 +25,7 @@ import {
 import type { KanbanCard } from '../view-presets/kanban/pc/card.js';
 import { KanbanDragController } from '../view-presets/kanban/pc/controller/drag.js';
 import type { KanbanGroup } from '../view-presets/kanban/pc/group.js';
+import { KanbanViewUI } from '../view-presets/kanban/pc/kanban-view-ui-logic.js';
 
 type Column = {
   id: string;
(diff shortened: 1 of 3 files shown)


darkskygit (approved):

darkskygit:
the change broke some kanban tests

PR #14753, opened 2026-03-31
[codex] improve tablet drawing interactions

pipedexplumbing:
## What changed

This PR improves mobile and tablet drawing interactions in edgeless mode.

It adds touch-first undo/redo controls to the edgeless toolbar, moves the edgeless toolbar above the virtual keyboard on mobile, blurs active text inputs before switching to pen or eraser tools, and changes the mobile keyboard toolbar to scroll the current editor viewport instead of lurching the window.

It also tightens the turbo renderer canvas mount logic so the canvas is attached to the current editor's viewport instead of the first matching viewport in the document.

## Why

Tablet web editing had a cluster of interaction problems:

- the virtual keyboard bar could overlap the pen toolbar
- stylus input could end up typing into an active text field while the user was trying to draw
- opening keyboard tool panels could overscroll the canvas downward
- touch surfaces lacked obvious undo/redo affordances
- turbo renderer canvas mounting was too global, which made viewport behavior more fragile than it needed to be

## User impact

On mobile web and tablet surfaces, drawing tools should behave more like drawing tools and less like a haunted text editor. The pen toolbar should stay usable when the keyboard is up, pen/eraser selection should stop competing with active text focus, and mobile viewport scrolling should be calmer when selecting or editing blocks.

## Root cause

The issues were split across a few layers:

- mobile toolbar positioning did not account for `visualViewport` keyboard offsets
- tool switching left the currently focused text input alive
- keyboard-toolbar scroll recovery used window scrolling instead of the nearest editor viewport
- touch-first undo/redo controls were missing from the edgeless toolbar
- turbo renderer canvas mounting relied on a global viewport query instead of the active editor context

## Validation

- `yarn eslint --no-warn-ignored` on the touched mobile/BlockSuite files
- cherry-pick conflict resolved onto `canary` in a clean worktree
- isolated branch pushed as `codex/mobile-drawing-interactions`

## Notes

A narrow TypeScript check in the isolated worktree is still noisy because of an existing ambient `VirtualKeyboard` type issue in `blocksuite/framework/global/src/disposable/index.ts`, which is outside this change.

### DIFF
--- a/blocksuite/affine/gfx/brush/src/toolbar/components/eraser/eraser-tool-button.ts
+++ b/blocksuite/affine/gfx/brush/src/toolbar/components/eraser/eraser-tool-button.ts
@@ -66,7 +66,10 @@ export class EdgelessEraserToolButton extends EdgelessToolbarToolMixin(
         ></affine-tooltip-content-with-shortcut>`}
         .tooltipOffset=${4}
         .active=${type === EraserTool}
-        @click=${() => this.setEdgelessTool(EraserTool)}
+        @click=${() => {
+          (this.ownerDocument.activeElement as HTMLElement | null)?.blur?.();
+          this.setEdgelessTool(EraserTool);
+        }}
       >
         <div class="eraser-button">${icon}</div>
       </edgeless-toolbar-button>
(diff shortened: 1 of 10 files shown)


pipedexplumbing:
Closing this draft PR because it was opened from the wrong GitHub owner/fork by mistake.

PR #14831, opened 2026-04-15
fix(core): prevent navigation panel from reordering while typing

### LINKED ISSUE #14714: [Bug]: Sidebar constantly reordering while typing
### What happened?

While typing in a document the documents in the sidebar are constantly reordering. 

### Distribution version

macOS ARM 64 (Apple Silicon)

### App Version

0.26.3

_No response_

### Are you self-hosting?

- [x] Yes

### Self-hosting Version

0.26.4

### Relevant log output

```shell

```

### Anything else?

_No response_

davideconte:
## What's the problem?

Closes #14714

The navigation panel sidebar visibly reorders linked child documents on every keystroke while typing in a document.

**Root cause:** `watchRefsFrom()` in `DocsSearchService` has no stabilization on its Observable output. The indexer re-runs approximately every 1 second during editing (triggered by Yjs `afterTransaction`). Each cycle causes `search$` to re-emit results in arbitrary search-engine order → React re-renders the sidebar children in a different order → visible reordering on every keystroke.

## Fix

Added a `distinctUntilChanged` guard after the `switchMap` in `watchRefsFrom()` that compares the **set** of linked `docId`s rather than the array reference or order. Downstream consumers only receive a new emission when a link is actually added or removed — not when the search engine returns the same documents in a different order.

```ts
distinctUntilChanged((prev, curr) => {
  if (prev.length !== curr.length) return false;
  const currIds = new Set(curr.map(r => r.docId));
  return prev.every(r => currIds.has(r.docId));
})
```

## Known limitation

This fix prevents reordering but does not enforce a *deterministic* display order. A meaningful stable order (e.g. document-appearance order as in Notion) would require block position data from the indexer and is tracked separately.

## Testing

1. Open a document that has linked child documents in the navigation panel
2. Start typing — the sidebar children should remain in a stable order
3. Add or remove a link — the sidebar should update correctly

## Summary by CodeRabbit

* **Performance**
  * Optimized search to reduce redundant updates when results remain unchanged, improving overall efficiency.


davideconte:
The E2E BlockSuite Cross Browser Test failure is unrelated to this PR — it fails in `packages/frontend/blocksuite/src/__tests__/edgeless/connector-dom.spec.ts:198` (connector canvas test), while this PR only modifies `packages/frontend/core/src/modules/docs-search/services/docs-search.ts` (navigation panel). This appears to be a pre-existing flaky cross-browser test. The other 4 E2E BS Test jobs passed successfully.

darkskygit (approved):

Issue #14059, opened 2025-12-07
[Bug]: Incorrect key mapping: Cyrillic 'х' triggers bracket auto-closing in markdown code blocks


nomoreqwerty:
### What happened?

When typing the Cyrillic letter 'х' (ha) inside a markdown code block, the editor inserts square brackets [] instead of the character. This happens because the Russian 'х' is located on the same physical key as [ on the US QWERTY layout

It seems the editor triggers the "auto-close brackets" feature based on the physical keycode, ignoring the current keyboard layout (Russian).

This works the same for the letter 'э' which is located on the same key as '

This issue affects both the desktop application and the web version

**Steps to Reproduce:**
1. Create a code block in a document using triple brackets (```)
2. Switch the keyboard layout to Russian
3. Type the letter 'х' (located on the [ key) within the code block
4. Type the letter 'э' (located on the ' key) within the code block

**Expected Behavior:**
When typed 'х' - Cyrillic character 'х' should be inserted
When typed 'э' - Cyrillic character 'э' should be inserted

**Actual Behavior:**
When typed 'х' - the editor ignores the layout and inserts square brackets []
When typed 'э' - the editor ignores the layout and inserts quotes ''

**Keyboard Layout:**
Russian / English (US)﻿

### Distribution version

Windows x64

### App Version

0.25.5

### Are you self-hosting?

- [ ] Yes

### Relevant log output

```shell

```

americanexplorer13:
Same problem

PR #14866, opened 2026-04-22
fix(core): prevent Alt+Key shortcuts from hijacking macOS Option-key input

### LINKED ISSUE #14519: [Bug]: Keyboard shortcuts unable writing in Polish
### What happened?

Hi, I'm just starting my way with affine, but I've stumbled upon a bug which does not allow me to use it in my mother language.
In the web version on MacOS I'm unable to write the Polish letter "ś" nor "Ś" (option + s), because the keyboard shortcut for "switch view" kicks in.
Expected behaviour: keyboard shortcut gets hard-remapped to something else or there's ability to remap the shortcuts yourself.
Thanks in advance

### Distribution version

Web (https://app.affine.pro)

### App Version

_No response_

Chrome

### Are you self-hosting?

- [ ] Yes

### Self-hosting Version

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

Abdulrehman-PIAIC80387:
Fixes #14519

## Summary

On macOS, the Option key combined with a letter produces locale input characters (e.g. Polish layout: Option+S → `ś`, Option+L → `ł`). The AFFiNE command registry registers shortcuts like `Alt+KeyS` (used for Page ↔ Edgeless mode switch) via `tinykeys`, which matches on `event.code` (the physical key) — so it fires even when the user was actually typing a non-ASCII character.

Reported in #14519: Polish users cannot type `ś` inside AFFiNE because Option+S triggers the mode switch instead.

## Fix

In the command registry handler ([registry.ts](packages/frontend/core/src/commands/registry/registry.ts)), skip the command when Alt is the only modifier **and** the key produced a non-ASCII character — the user intends to type the character, not invoke the shortcut.

Matches the existing handling in blocksuite's `keymap.ts` (added for the same class of issue in #14059).

## Demo

https://github.com/user-attachments/assets/eb6d2e69-39bf-4236-a886-9e2bde425626

## Verified locally (macOS)

- Switched input source to Polish
- Typed `właśnie` in an AFFiNE doc — all characters including `ś` (Option+S), `ł` (Option+L) now produce the correct output
- Previously Option+S would toggle edgeless mode
- US layout (Option+S → `ß`) and other locale chars (ą, ń, ę) also now pass through correctly
- Regular Cmd-based shortcuts (Cmd+K, Cmd+S, etc.) unaffected because the guard excludes `metaKey`

## Test plan

- [x] On macOS, add Polish input source (System Settings → Keyboard → Input Sources → +)
- [x] Switch to Polish layout
- [x] In any AFFiNE doc, type Option+S → `ś` appears (not mode switch)
- [x] Confirm other shortcuts (Cmd+K, Cmd+Enter, etc.) still work
- [x] Confirm on US layout that Option+S produces `ß` (OS default) without firing the mode switch

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed keyboard event handling with Alt key and non-ASCII characters to prevent unintended command execution.

### DIFF
--- a/packages/frontend/core/src/commands/registry/registry.ts
+++ b/packages/frontend/core/src/commands/registry/registry.ts
@@ -68,6 +68,20 @@ export const AffineCommandRegistry = new (class {
         window,
         {
           [keybinding]: (e: Event) => {
+            // Skip when Alt produces a locale input character (e.g. macOS
+            // Polish layout: Option+S → "ś"). The user is typing the
+            // character, not invoking the shortcut bound to the physical
+            // key. Matches the handling in blocksuite's keymap.ts.
+            if (
+              e instanceof KeyboardEvent &&
+              e.altKey &&
+              !e.metaKey &&
+              !e.ctrlKey &&
+              e.key.length === 1 &&
+              e.key.charCodeAt(0) > 0x7e
+            ) {
+              return;
+            }
             e.preventDefault();
             command.run()?.catch(e => {
               console.error(`Failed to run command [${command.id}]`, e);


Abdulrehman-PIAIC80387:
Hi @darkskygit — ready for review whenever you have a moment. Any feedback is welcome. Thanks!

darkskygit (approved):

PR #14905, opened 2026-05-05
feat(blocksuite): add password encrypted blocks

ygcaicn:
## Summary
- add password-based AES-GCM helpers and encrypted block metadata support
- add Encrypt/Lock toolbar actions for content blocks without storing the password
- render locked blocks through a root widget overlay, and show unlocked encrypted blocks with an inline marker while edits are re-encrypted from an in-memory session key

## Verification
- yarn tsc -b blocksuite/affine/shared blocksuite/affine/blocks/root blocksuite/affine/all --pretty false
- yarn eslint blocksuite/affine/blocks/root/src/view.ts blocksuite/affine/blocks/root/src/effects.ts blocksuite/affine/blocks/root/src/encryption/encrypted-blocks-widget.ts blocksuite/affine/blocks/root/src/configs/toolbar.ts blocksuite/affine/shared/src/encryption/password-aes.ts blocksuite/affine/shared/src/encryption/block-encryption.ts blocksuite/affine/shared/src/__tests__/encryption/password-aes.unit.spec.ts
- yarn vitest run blocksuite/affine/shared/src/__tests__/encryption/password-aes.unit.spec.ts

## Summary by CodeRabbit

* **New Features**
  * Password-protected block encryption and decryption via new toolbar actions (Encrypt, Lock).
  * Block locking/unlocking for encrypted content with editable unlocked state and auto-persist of edits.
  * Visual encrypted-block overlays showing previews and a Decrypt action.

* **Tests**
  * Added unit tests covering password-AES encryption/decryption and helper utilities.

### DIFF
--- a/blocksuite/affine/blocks/root/src/configs/toolbar.ts
+++ b/blocksuite/affine/blocks/root/src/configs/toolbar.ts
@@ -43,6 +43,13 @@ import {
   getSelectedModelsCommand,
   getTextSelectionCommand,
 } from '@blocksuite/affine-shared/commands';
+import {
+  canEncryptBlock,
+  encryptBlockWithPassword,
+  isBlockEncrypted,
+  isBlockLocallyUnlocked,
+  lockBlockWithEncryptedEdits,
+} from '@blocksuite/affine-shared/encryption';
 import type {
   ToolbarAction,
   ToolbarActionGenerator,
@@ -52,6 +59,7 @@ import type {
 import {
   ActionPlacement,
   blockCommentToolbarButton,
+  NotificationProvider,
 } from '@blocksuite/affine-shared/services';
 import { getMostCommonValue } from '@blocksuite/affine-shared/utils';
 import { tableViewMeta } from '@blocksuite/data-view/view-presets';
@@ -61,6 +69,7 @@ import {
   DeleteIcon,
   DuplicateIcon,
   LinkedPageIcon,
+  LockIcon,
 } from '@blocksuite/icons/lit';
 import {
   type BlockComponent,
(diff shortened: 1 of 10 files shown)


ygcaicn:
### Encrypted block workflow screenshots

These screenshots show how password encrypted blocks are expected to work. The password is a per-block password, not the account password, and it is not persisted anywhere.

1. **Encrypt action from the block context menu**

Encrypt action

Any supported content block keeps its original block type. The context menu adds `Encrypt`, which encrypts that block props into the encryption payload instead of converting it into a new visible block type.

2. **Locked rendering hides the original content**

Locked encrypted block

When locked, the real block content is hidden and the user sees an encrypted-card UI with the ciphertext/base64 preview, the original block flavour, and the `Decrypt` action.

3. **Decrypt prompt**

Decrypt prompt

Clicking `Decrypt` on a locked encrypted block opens a modal prompt. The user enters the block password here; the password is used only for the decrypt attempt and is not stored.

4. **Unlocked rendering keeps the original block UI**

Unlocked encrypted block

After decrypting, the block renders normally as its original type. A small centered `Encrypted` marker is shown below the block to indicate that the block is still encrypted at rest.

5. **Lock action for unlocked encrypted blocks**

Lock action

While the block is unlocked, edits are re-encrypted into the stored payload using the in-memory session key so changes are not lost. The context menu switches to `Lock`; locking flushes the latest edits into encrypted storage and restores the locked placeholder UI.

darkskygit:
Thank you for your PR, but I think this is a poor interaction, and embedding encrypted data in CRDT data is also not conducive to collaborative performance, so we are not accepting this PR for now.

PR #15058, opened 2026-06-01
fix: handle empty results in MCP keyword_search tool

### LINKED ISSUE #15038: [Bug]: `keyword_search` MCP tool errors on no-result queries
### Summary

The AFFiNE MCP `keyword_search` tool appears to fail when a valid query returns no matching documents. Instead of returning an empty result set or a normal "no results" response, the tool call errors with `Unexpected response type`.

### Steps to Reproduce

Call the AFFiNE MCP `keyword_search` tool with a query that is valid but unlikely to match any documents:

```json
{
  "query": "zzzz_nonexistent_affine_probe_20260529"
}
```

### Actual Result

The tool call fails:

```text
tool call error: tool call failed for `affine/keyword_search`

Caused by:
    Unexpected response type
```

### Expected Result

The tool should return a valid MCP response, for example:

```json
[]
```

or a text response such as:

```text
No matching documents found.
```

The important part is that a no-results query should not throw a tool-call error.

### Additional Observations

Other related cases behave more gracefully:

- `keyword_search` with a matching query works and returns document metadata.
- `keyword_search` with an empty query returns: `Query is required for keyword search.`
- `semantic_search` with an empty query returns: `Query is required for semantic search.`
- `read_document` with an invalid doc ID returns a normal text response: `Doc with id ... not found.`

So the issue seems specific to `keyword_search` when the query is non-empty but produces zero matches.

### Environment

- AFFiNE MCP server
- Tool tested: `keyword_search`
- Client observed through Codex MCP tool integration

### Distribution version

macOS ARM 64 (Apple Silicon)

### App Version

0.26.3

_No response_

### Are you self-hosting?

- [ ] Yes

### Self-hosting Version

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

Marsssssssssssdsss:
## Description

Fixes: #15038 — MCP keyword_search tool errors with "Unexpected response type" when no results are found.

### Problem

When the MCP `keyword_search` tool returns no matching documents, the access control `.docs()` method may return `undefined`/`null` for an empty input array. Calling `.map()` on this value throws an error, and the MCP framework wraps it as "Unexpected response type".

### Solution

Added a guard check after the permission filtering step. If the result is empty or null, the tool now returns a proper informational response instead of throwing.

### Changes

- `packages/backend/server/src/plugins/copilot/mcp/provider.ts`: Added null/empty check before `docs.map()` in the keyword_search tool execute function.

### Testing

- **Before**: `keyword_search` with a non-existent keyword throws "Unexpected response type"
- **After**: `keyword_search` with a non-existent keyword returns `{ content: [{ type: 'text', text: 'No matching documents found.' }] }`

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented errors when document data is missing, improving search stability.
  * Improved search feedback by displaying a clear "No matching documents found." message instead of empty results.

### DIFF
--- a/packages/backend/server/src/plugins/copilot/mcp/provider.ts
+++ b/packages/backend/server/src/plugins/copilot/mcp/provider.ts
@@ -230,6 +230,17 @@ export class WorkspaceMcpProvider {
         const abortedAfterDocs = abortIfNeeded(options.signal);
         if (abortedAfterDocs) return abortedAfterDocs;
 
+        if (!docs || docs.length === 0) {
+          return {
+            content: [
+              {
+                type: 'text',
+                text: 'No matching documents found.',
+              },
+            ],
+          };
+        }
+
         return {
           content: docs.map(doc => ({
             type: 'text',
--- a/packages/backend/server/src/plugins/copilot/mcp/provider.ts
+++ b/packages/backend/server/src/plugins/copilot/mcp/provider.ts
@@ -231,14 +231,7 @@ export class WorkspaceMcpProvider {
         if (abortedAfterDocs) return abortedAfterDocs;
 
         if (!docs || docs.length === 0) {
-          return {
-            content: [
-              {
-                type: 'text',
-                text: 'No matching documents found.',
-              },
-            ],
-          };
+          return toolText('No matching documents found.');
         }
 
         return {
--- a/packages/backend/server/src/plugins/copilot/mcp/provider.ts
+++ b/packages/backend/server/src/plugins/copilot/mcp/provider.ts
@@ -191,10 +191,10 @@ export class WorkspaceMcpProvider {
         if (abortedAfterDocs) return abortedAfterDocs;
 
         return {
-          content: docs.map(doc => ({
+          content: (docs?.map(doc => ({
             type: 'text',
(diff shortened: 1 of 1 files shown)


packages/backend/server/src/plugins/copilot/mcp/provider.ts
darkskygit: i think the better way is `docs?.map( ... ) ?? []`

darkskygit:
Thank you for your PR; please sign the CLA

Marsssssssssssdsss:
Thanks for the review @darkskygit! I've applied both suggestions:
- Used `toolText()` helper for the empty check
- Changed `docs.map()` to `(docs?.map(...) ?? [])` for safety

Marsssssssssssdsss:
I have reviewed and agree to the AFFiNE Contributor License Agreement. Please recheck the CLA status.

Marsssssssssssdsss:
I have read and agree to the CLA. Please recheck.

darkskygit (approved):

Marsssssssssssdsss:
CLA has been signed. All review feedback has been applied (docs?.map() ?? [] and toolText() helper). Ready for re-review.

PR #15089, opened 2026-06-07
fix(callout): prevent backspace in icon picker search from deleting editor content

htmujahid:
## Problem
When the callout block's icon picker is open and the user types in the search input, pressing backspace deletes content in the main editor instead of the search text.

## Root Cause
The callout icon picker is mounted via `createPopup` inside `editor-host`. `PageKeyboardManager` registers a global `Backspace` handler on the editor host (`keyboard-manager.ts`) with `{ global: true }`, which fires on every backspace keydown regardless of what element is focused. Without `stopPropagation`, the backspace event from the search input bubbles up through the DOM and triggers block deletion.

Other keys are unaffected because the editor handles character input through `contenteditable` focus, those handlers only act when a contenteditable node is active.

## Fix
Add `onKeyDown` with `e.stopPropagation()` to the search inputs in both `EmojiPicker` and `AffineIconPicker`. This matches the existing pattern already used by `MenuComponent` (`menu-renderer.ts:107`) and all other interactive components (`date-picker`, `inline-edit`, `prompt-modal`).

## Why not affected elsewhere
`DocIconPicker` uses the same pickers but wraps them in a Radix UI `Menu` with `modal: true`, which portals outside `editor-host` — so backspace events never reach the editor's global handler there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard event handling in search inputs for icon and emoji pickers

### DIFF
--- a/packages/frontend/component/src/ui/icon-picker/picker/affine-icon/affine-icon-picker.tsx
+++ b/packages/frontend/component/src/ui/icon-picker/picker/affine-icon/affine-icon-picker.tsx
@@ -1,7 +1,13 @@
 import keywords from '@blocksuite/icons/keywords/en.json';
 import * as allIcons from '@blocksuite/icons/rc';
 import { cssVarV2 } from '@toeverything/theme/v2';
-import { startTransition, useCallback, useEffect, useState } from 'react';
+import {
+  type KeyboardEvent,
+  startTransition,
+  useCallback,
+  useEffect,
+  useState,
+} from 'react';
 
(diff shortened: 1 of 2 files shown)


htmujahid:
Before Fix

https://github.com/user-attachments/assets/27e6f7d6-8eea-4ca1-97c8-ddbb4ee48e52

After Fix

https://github.com/user-attachments/assets/ed71df78-c42b-4dba-91e1-f0435e115b09

darkskygit (approved):

PR #15129, opened 2026-06-18
fix(core): filters emojipicker on label in addition to tags

### LINKED ISSUE #15116: [Bug]: EmojiPicker keyword only searching in tags, not label
### What happened?

When searching for Emojis, the emoji list is only filtered on emoji tags, and not the label.
For example, when searching for the the "door" emoji ->🚪. One would naturally search "door", expecting the emoji to show up, but it does not. See below.
Image

This is due to the filtering function in [groups.tsx](https://github.com/toeverything/AFFiNE/blob/c51bdb74dee5f09878ebfa6fb469d5efbab6b75a/packages/frontend/component/src/ui/icon-picker/picker/emoji/groups.tsx#L127) only filtering on the emoji's `tags` list, and not the label.  as seen below
```javascript
// packages/frontend/component/src/ui/icon-picker/picker/emoji/groups.tsx
emojis: group.emojis.filter(emoji =>
              emoji.tags?.some(tag => tag.includes(keyword.toLowerCase()))
            ),
```
In the emoji data file [en.json](https://github.com/toeverything/AFFiNE/blob/c51bdb74dee5f09878ebfa6fb469d5efbab6b75a/packages/frontend/component/src/ui/icon-picker/picker/emoji/data/en.json#L30376), we can see that the door emoji does not have "door" as a tag, but "door" is the label.
```json
{
        "group": 7,
        "hexcode": "1F6AA",
        "label": "door",
        "order": 4424,
        "tags": ["back", "closet", "front"],
        "unicode": "🚪"
 },
```
The door emoji here is just an example. the problem persists in all emoji, but is mostly a problem for a bit more niche emojis.
Simply filtering based on label as well as tags would solve this.

### Distribution version

macOS x64 (Intel)

### App Version

0.26.3

_No response_

### Are you self-hosting?

- [x] Yes

### Self-hosting Version

stable

### Relevant log output

```shell

```

### Anything else?

_No response_

tines98:
Fixes #15116 
# Issue
Emojipicker keyword filtering only filtered on `tags`, and not `label`. So searching for an emoji's name would not result in said emoji ending up in the result. E.G. searching "sunflower" does not make 🌻 appear

# Solution
Adding an extra condition to the filter function to check if the keyword is a substring of an emoji's label

# Result
Search results now include emojis with that `label`

## Summary by CodeRabbit

* **New Features**
  * Improved emoji picker search to include matches on both emoji labels and tags (case-insensitive), enabling broader search results for better discoverability.

### DIFF
--- a/packages/frontend/component/src/ui/icon-picker/picker/emoji/groups.tsx
+++ b/packages/frontend/component/src/ui/icon-picker/picker/emoji/groups.tsx
@@ -123,8 +123,10 @@ export const EmojiGroups = memo(function EmojiGroups({
         emojiGroupList
           .map(group => ({
             ...group,
-            emojis: group.emojis.filter(emoji =>
-              emoji.tags?.some(tag => tag.includes(keyword.toLowerCase()))
+            emojis: group.emojis.filter(
+              (emoji) =>
+                emoji.label.toLowerCase().includes(keyword.toLowerCase()) ||
+                emoji.tags?.some((tag) => tag.includes(keyword.toLowerCase())),
             ),
           }))
           .filter(group => group.emojis.length > 0)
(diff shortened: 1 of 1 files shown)


darkskygit:
need to format file: `yarn prettier packages/frontend/component/src/ui/icon-picker/picker/emoji/groups.tsx --write`

darkskygit (approved):

PR #15149, opened 2026-06-24
fix: enforce quota for comment attachments

failsafesecurity:
## Summary

This change includes comment attachments in workspace storage usage and checks workspace storage quota before accepting a new comment attachment upload.

## Impact

Comment attachments already had a per-file size limit, but they were not counted in the same workspace storage usage path as other uploaded blobs. A user with comment permission could keep adding attachments without those bytes participating in workspace storage quota calculations.

## Fix

- Count comment attachment bytes in workspace storage usage reconciliation.
- Check the workspace quota before storing a new comment attachment.
- Return the existing comment attachment quota error when the upload would exceed limits.

## Validation

- `git diff --check`
- Full test/lint suite was not run locally because dependencies are not installed in this checkout.

## Summary by CodeRabbit

* **New Features**
  * Workspace attachment uploads now respect storage and file quota limits more accurately.
  * Workspace storage tracking now includes comment attachments, improving quota enforcement.

* **Bug Fixes**
  * Attachment uploads now fail with a clear quota error when a workspace is out of space or blob capacity.
  * Storage usage calculations now better reflect actual workspace content, including non-deleted files.

### DIFF
--- a/packages/backend/server/src/core/comment/index.ts
+++ b/packages/backend/server/src/core/comment/index.ts
@@ -2,6 +2,7 @@ import { Module } from '@nestjs/common';
 
 import { ServerConfigModule } from '../config';
 import { PermissionModule } from '../permission';
+import { QuotaServiceModule } from '../quota';
 import { StorageModule } from '../storage';
 import { CommentRealtimeModule } from './realtime.module';
 import { CommentResolver } from './resolver';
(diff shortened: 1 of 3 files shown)


darkskygit:
thank you for your pr, please sign the cla so we can proceed with merging this fix

failsafesecurity:
Hi maintainers 👋 — the CLA is now signed. All checks should be passing. Ready for review when you have a moment. Thanks!

darkskygit (approved):

PR #15150, opened 2026-06-24
fix: restrict email invite metadata lookup

failsafesecurity:
## Summary

This change restricts email invitation metadata lookup to the authenticated user that the invitation was issued to.

## Impact

The public invite-info resolver returns workspace, owner, invitee, and status details for a valid invitation identifier. For email invitations, that information should only be returned to the intended recipient after authentication.

## Fix

- Keep existing link-invitation behavior unchanged.
- For email invitations, require an authenticated user whose id matches the invitation recipient before returning invitation details.
- Return the existing invalid-invitation error for mismatched or unauthenticated access.

## Validation

- `git diff --check`
- Full test/lint suite was not run locally because dependencies are not installed in this checkout.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened invitation access checks so non-link invites are only readable by the intended recipient.
  * Invalid or missing user context now returns an error earlier, preventing access to invite details when the invitation doesn’t match.


failsafesecurity:
Hi maintainers 👋 — the CLA is now signed. CodeRabbit's nitpick about regression tests is noted — happy to add them if you'd like, but the fix is self-contained. Ready for review. Thanks!

darkskygit (approved):

failsafesecurity:
Friendly nudge — has anyone had a chance to review this security fix? Happy to address any feedback. Thanks!

PR #14624, opened 2026-03-10
feat(i18n): automatic RTL layout for Arabic, Persian, and Urdu + complete Arabic translations

imohad:
### RTL Support (automatic, locale-driven)
- Add `rtl?: boolean` metadata to locale definitions in `SUPPORTED_LANGUAGES`
- Set `rtl: true` for Arabic (`ar`), Persian (`fa`), and Urdu (`ur`)
- Automatically set `document.documentElement.dir` based on locale RTL metadata on language change
- Remove hardcoded `lang="en"` from HTML template — JS now controls both `lang` and `dir`

### Arabic Translations
- Add 100 missing keys to `ar.json` (Calendar integration, Doc Analytics, MCP Server, AI Chat, and more)
- Arabic locale now has 2,313/2,313 keys (100% coverage, matches `en.json` exactly)

## Testing
Switching to Arabic/Persian/Urdu now automatically flips the entire UI layout to RTL without any manual feature flag.

Fixes #7099

## Summary by CodeRabbit

* **New Features**
  * Added Right-to-Left (RTL) support for Arabic, Persian, and Urdu with automatic document direction and language attributes when a language is selected.

* **Refactor**
  * Centralized and reordered internal language handling so document language and direction are applied earlier and consistently.

* **Chore**
  * Set a default text direction attribute on the base HTML template.

Issue #8966, opened 2024-11-29
​RTL Text Direction Support


AbdulrahmanNahhas:
### What happened?

AFFINE does not currently support RTL text direction. This makes it difficult to use for applications that need to display RTL text, such as Arabic, and Persian.

### Distribution version

macOS x64 (Intel)

### Are you self-hosting?

- [ ] Yes

### Anything else?

I wanted to highlight that, according to W3C data, approximately **2.3 billion people** around the world use Right-to-Left (RTL) languages. This represents a significant portion of the global population, making RTL support essential for ensuring accessibility and inclusivity for millions of users.

zoh-li:
This project is only lacking RTL support.
I hope this issue gets resolved as soon as possible.

AsimNet:
upvote here: https://github.com/toeverything/blocksuite/issues/7894
🚀

ParsaUltimate:
Great to see the active work on this through PRs #15214, #15215, and #15216! As a Persian-speaking user, this feature is critically important to me and many others.

I'd like to share a few additional points and use-case considerations that I hope will help shape the final implementation:

## Persian/Farsi-Specific Concerns

Persian (Farsi) has some nuances beyond Arabic that are worth considering:
- **Persian numerals** (۰۱۲۳۴۵۶۷۸۹) are widely used alongside Eastern Arabic numerals — the bidirectional rendering of numeric content inside RTL paragraphs should handle both.
- **Mixed content** is extremely common in Persian tech writing: Persian prose with inline English terms, code snippets, URLs, and Latin abbreviations. The per-block `dir=auto` approach (as in #15215) should handle this well, but edge cases like inline code blocks inside RTL paragraphs deserve testing.
- **List items and nested blocks** (checklists, toggles, callouts) should inherit or independently set RTL direction.

## Feature Checklist I Hope Gets Covered

- [ ] Auto RTL detection per block (`dir=auto` or Unicode-first-strong heuristic)
- [ ] Manual direction override via toolbar or slash command (`/rtl`, `/ltr`)
- [ ] RTL support in Edgeless/Whiteboard text elements
- [ ] Correct cursor and Home/End key behavior in RTL blocks
- [ ] RTL-aware export to HTML (`dir="rtl"`) and PDF
- [ ] RTL in table cells
- [ ] RTL in the mobile app

## Impact

[...]

PR #15214, opened 2026-07-10
feat(core): add Radix DirectionProvider for RTL support

nirpow:
## What

Mounts a `DirectionProvider` in `AffineContext` that feeds the current language's text direction to all Radix UI primitives. Adds an `rtl` flag to `LanguageInfo` (from `SUPPORTED_LANGUAGES`), with a colocated unit test.

Part of #8966.

## Why

Radix primitives read direction from React context, not from the document `dir` attribute. Since #14624, `document.documentElement.dir` correctly flips to `rtl` for Arabic/Persian/Urdu — but every Radix dropdown, popover and menu still hard-renders LTR internally.

## Notes for reviewers

- `@radix-ui/react-direction` is pinned to the exact version the other Radix primitives resolve (`1.1.1`), not a caret range. A caret range resolves to a newer duplicate module copy, which means a different React context object — the primitives would silently ignore the provider. Verified at runtime both ways.
- Portaled Radix content preserves React context, so portaled popovers inherit direction too.

## Testing

- `yarn typecheck`, `yarn vitest run packages/frontend/core/src/modules/i18n` (3 tests) pass.
- Verified in the running web app (Playwright + manual): with language set to العربية, the workspace selector popover and menus render `dir="rtl"`; English UI unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added automatic right-to-left layout support for languages such as Arabic, Persian, and Urdu.
  * Interface direction now updates with the selected language across supported UI components.

* **Tests**
  * Added coverage confirming correct direction settings for right-to-left, left-to-right, and default English languages.

### DIFF
--- a/packages/frontend/core/package.json
+++ b/packages/frontend/core/package.json
@@ -40,6 +40,7 @@
     "@preact/signals-core": "^1.8.0",
     "@radix-ui/react-collapsible": "^1.1.2",
     "@radix-ui/react-dialog": "^1.1.3",
+    "@radix-ui/react-direction": "1.1.1",
     "@radix-ui/react-popover": "^1.1.3",
     "@radix-ui/react-scroll-area": "^1.2.2",
     "@radix-ui/react-slot": "^1.1.1",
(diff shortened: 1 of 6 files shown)


rellikivo-ai:
We need RTL support for the fantastic app AFFINE ❤️

PR #15243, opened 2026-07-14
fix(editor): make Android keyboard autocorrect work in the mobile editor

### LINKED ISSUE #14021: Samsung keyboard issue on Android app
### What happened?

I am experiencing an issue with the Android app on a Samsung s23 device. Typing on the keyboard triggers abnormal behavior. If I don’t specify that a block is text, I have to type a word and then a space, and the word gets deleted. When I edit a word in a text block, my new input deletes the text that follows. And same kind of other issues.

I found a workaround by switching from the Samsung keyboard to Google’s Gboard keyboard."

### Distribution version

Linux

### App Version

Android app

_No response_

### Are you self-hosting?

- [x] Yes

### Self-hosting Version

0.24.4

### Relevant log output

```shell

```

### Anything else?

_No response_

KingJulien1st:
## Problem

On Android, keyboard autocorrect (Samsung keyboard, Gboard, HeliBoard, AOSP)
rewrites the current word on space or enter through IME composition. Typing in
the editor was corrupted by this:

- a word followed by space deleted the word,
- editing inside a word deleted the text after the cursor,
- creating a new line was unreliable,
- the stray text was visible on screen but gone after closing and reopening the
  document (it lived in the DOM only and never reached the model).

It reproduced with several keyboards and in mobile browsers (Chrome, Brave,
Firefox), not only in the Android app, and disabling autocorrect/suggestions
worked around it.

Closes #14021

## Root causes and fixes

1. **Nested contenteditable on mobile.** The page root was `contenteditable`
   and competed with each block's own inline editor. On Android the page root
   captured the IME input while its native selection sat between blocks
   (whitespace and `<style>` text), so composed text never reached the block and
   never committed to the model. The page root is no longer contenteditable on
   mobile (`IS_MOBILE`), so the block inline editors keep focus and receive the
   IME input. This is the change that makes body editing usable again.

2. **Composition range was over-counted.** During composition the composed
   characters live in the DOM but not in the model, so the commit index captured
   from the caret pointed past the real position. The composition start is now
   taken from the composition target range and clamped to the model bounds. This
   also fixes the first word typed in an empty line disappearing.

3. **Enter on Android** arrives as an `insertParagraph` beforeinput rather than a
   keydown, so it was inserted as a soft line break inline instead of splitting
   the block. It is now routed to the block `Enter` binding.

4. **Stale native range recovery.** Some IMEs fire the input that follows a
   composition before the browser re-syncs its selection to the rebuilt DOM. That
   input is now applied at the editor's current inline range instead of letting
   the browser mutate the DOM natively.

## Impact of the page-root change

- Mobile only (`IS_MOBILE`); desktop is unchanged (covered by a regression test).
- Block inline editors stay `contenteditable`, so typing, caret and IME work.
- The two code paths that query `contenteditable` (range-manager `clear()` and
  range-binding selection sync) were reviewed and behave neutrally or better:
  selections that land between blocks are now correctly ignored.

## Testing

- Unit tests (framework/std): composition commit at the caret, autocorrect on
  space keeps the corrected word in place, first word in an empty line is kept,
  Android `insertParagraph` does not insert a soft line break, `insertParagraph`
  is routed to the `Enter` binding, and stale-range recovery.
- Regression test (integration-test): the page root stays contenteditable on
  desktop.
- Manual: verified on a Samsung device with the native Samsung keyboard (the
  reported case). Text input, autocorrect on space, editing mid-word and Enter
  (new block) all work, and the content stays consistent after reopening.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Android IME composition/commit handling, including autocorrect-on-commit and correct caret positioning.
  * Prevented IME `insertParagraph` from creating unintended soft inline line breaks.
  * Reduced risks of stale/“out of root” beforeinput selection causing editor/model desync by routing through editor ranges.
  * Routed Android `insertParagraph` to the Enter binding when available, with consistent patched keyboard state.
  * Clamped out-of-bounds composition ranges to preserve composed text.
* **Mobile**
  * Disabled editing on the outer page container on mobile to ensure IME targets inner editors.
* **Tests**
  * Added Android IME, keymap routing, and desktop page-root contenteditable regression coverage.

### DIFF
--- a/blocksuite/affine/blocks/root/src/page/page-root-block.ts
+++ b/blocksuite/affine/blocks/root/src/page/page-root-block.ts
@@ -19,6 +19,7 @@ import {
   getScrollContainer,
   matchModels,
 } from '@blocksuite/affine-shared/utils';
+import { IS_MOBILE } from '@blocksuite/global/env';
 import { Point } from '@blocksuite/global/gfx';
 import type { PointerEventState } from '@blocksuite/std';
 import { BlockComponent, BlockSelection, TextSelection } from '@blocksuite/std';
@@ -413,7 +414,13 @@ export class PageRootBlockComponent extends BlockComponent<RootBlockModel> {
       return !(isNote && displayOnEdgeless);
     });
 
-    this.contentEditable = String(!this.store.readonly$.value);
+    // On mobile the page root being contenteditable competes with each block's
+    // own inline editor: keyboards like the Samsung one focus the outer page
+    // root and their selection lands between blocks (whitespace / <style>
+    // text), so IME input never reaches the block and autocorrect on space or
+    // enter corrupts the text. Keep only the inner block editors editable on
+    // mobile. See https://github.com/toeverything/AFFiNE/issues/14021
+    this.contentEditable = String(!this.store.readonly$.value && !IS_MOBILE);
 
     return html`
       <div class="affine-page-root-block-container">${children} ${widgets}</div>
--- a/blocksuite/framework/std/src/__tests__/inline/event-composition.unit.spec.ts
+++ b/blocksuite/framework/std/src/__tests__/inline/event-composition.unit.spec.ts
@@ -1,0 +1,368 @@
+import { expect, test, vi } from 'vitest';
+import * as Y from 'yjs';
+
+// Force the Android IME code path, which is where autocorrect-on-space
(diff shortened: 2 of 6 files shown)


KingJulien1st:
Closing this pull request because obviously no one cares

PR #15247, opened 2026-07-15
fix(server): mcp api visibility

### LINKED ISSUE #15246: [Bug]: Valid MCP credentials are always rejected after recent PR
### What happened?

AFFiNE Cloud generates MCP credentials with this structure:

aff_mcp_v1.<uuid>.<secret>

The generated configuration correctly sends the credential as:

Authorization: Bearer aff_mcp_v1.<uuid>.<secret>

However, every MCP request using a newly generated credential returns HTTP 401 with `ACCESS_TOKEN_INVALID`. This reproduces with a direct HTTP request, independently of Codex or any other MCP client.

Expected behavior:

A newly generated, active MCP credential should successfully initialize the workspace MCP endpoint.

Actual behavior:

The request is rejected by AFFiNE's global authentication layer before it reaches the MCP credential validator.

I also performed a controlled comparison:

- The original three-segment MCP token returns the generic `ACCESS_TOKEN_INVALID` response.
- Appending a fourth diagnostic segment changes the response to the MCP controller's JSON-RPC `Authentication failed` response.

This suggests that the valid three-segment token is being intercepted as a session JWT.

### Distribution version

Web (https://app.affine.pro)

### App Version

0.27.0

_No response_

### Are you self-hosting?

- [ ] Yes

### Self-hosting Version

_No response_

### Relevant log output

```shell
WORKSPACE_ID="<workspace-id>"
MCP_TOKEN="<redacted aff_mcp_v1 token>"

curl -i \
  --request POST \
  --header "Authorization: Bearer ${MCP_TOKEN}" \
  --header "Content-Type: application/json" \
  --header "Accept: application/json, text/event-stream" \
  --data '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "initialize",
    "params": {
      "protocolVersion": "2025-03-26",
      "capabilities": {},
      "clientInfo": {
        "name": "mcp-diagnostic",
        "version": "1.0"
      }
    }
  }' \
  "https://app.affine.pro/api/workspaces/${WORKSPACE_ID}/mcp"

HTTP/2 401

{
  "status": 401,
  "code": "ACCESS_TOKEN_INVALID",
  "type": "AUTHENTICATION_REQUIRED",
  "name": "ACCESS_TOKEN_INVALID",
  "message": "The access token is invalid."
}
```

### Anything else?

The likely cause is a collision between the new MCP credential format and the session-JWT detection heuristic introduced around PR #15221.

1. MCP credentials are deliberately generated with three dot-separated segments:

https://github.com/toeverything/AFFiNE/blob/c61cc6a86f5f8364732296f0bb8393b37e0f70b3/packages/backend/server/src/plugins/copilot/mcp/credential.ts#L221-L235

2. `isLikelyJwt()` classifies every token containing exactly three segments as a JWT:

https://github.com/toeverything/AFFiNE/blob/c61cc6a86f5f8364732296f0bb8393b37e0f70b3/packages/backend/server/src/core/auth/access-token.ts#L119-L121

3. The global guard attempts optional authentication before returning early for a `@Public()` endpoint:

https://github.com/toeverything/AFFiNE/blob/c61cc6a86f5f8364732296f0bb8393b37e0f70b3/packages/backend/server/src/core/auth/guard.ts#L105-L150

As a result, `aff_mcp_v1.<uuid>.<secret>` is sent to the session JWT verifier. It fails there with `ACCESS_TOKEN_INVALID`, so the MCP controller never receives the valid credential.

The current MCP test appears to invoke `WorkspaceMcpController.mcp()` directly, bypassing the global AuthGuard:

https://github.com/toeverything/AFFiNE/blob/c61cc6a86f5f8364732296f0bb8393b37e0f70b3/packages/backend/server/src/__tests__/copilot/copilot.spec.ts#L281-L304

Suggested fix:

- Make JWT detection validate the encoded JOSE header instead of relying only on `token.split('.').length === 3`; or
- Ensure `aff_mcp_v1.*` credentials bypass session-JWT authentication and are handled by `McpCredentialService`.

Suggested regression test:

Issue a real MCP credential and use Supertest against the complete HTTP route, including the global AuthGuard. Assert that `initialize` and `tools/list` succeed. A direct controller invocation does not cover this failure.

Related change:

https://github.com/toeverything/AFFiNE/pull/15221

darkskygit:
fix #15246

#### PR Dependency Tree

* **PR #15247** 👈

This tree was auto-generated by [Charcoal](https://github.com/danerwilliams/charcoal)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency when issuing and validating MCP credential tokens by using a shared token prefix across issuance and parsing.
  * Preserved correct recognition of standard JWT-based authentication tokens.

* **Tests**
  * Updated MCP credentials coverage to validate behavior through the HTTP API response (instead of direct controller invocation).
  * Adjusted workspace quota e2e setup to derive restricted limits via entitlements before reconciling quota state.

### DIFF
--- a/packages/backend/server/src/__tests__/copilot/copilot.spec.ts
+++ b/packages/backend/server/src/__tests__/copilot/copilot.spec.ts
@@ -5,7 +5,6 @@ import { ProjectRoot } from '@affine-tools/utils/path';
 import { McpAccessMode, PrismaClient } from '@prisma/client';
 import type { TestFn } from 'ava';
 import ava from 'ava';
-import type { Request, Response } from 'express';
 import { nanoid } from 'nanoid';
 import Sinon from 'sinon';
 import { z } from 'zod';
@@ -43,7 +42,6 @@ import {
   CopilotEmbeddingJob,
   MockEmbeddingClient,
 } from '../../plugins/copilot/embedding';
-import { WorkspaceMcpController } from '../../plugins/copilot/mcp/controller';
 import { McpCredentialService } from '../../plugins/copilot/mcp/credential';
 import { WorkspaceMcpProvider } from '../../plugins/copilot/mcp/provider';
 import { PromptService } from '../../plugins/copilot/prompt';
@@ -84,12 +82,12 @@ import { SubscriptionService } from '../../plugins/payment/service';
 import { SubscriptionStatus } from '../../plugins/payment/types';
 import { installMockCopilotRuntime, MockCopilotProvider } from '../mocks';
 import { TestingPromptService } from '../mocks/prompt-service.mock';
-import { createTestingModule, TestingModule } from '../utils';
+import { createTestingApp, TestingApp } from '../utils';
 import { singleUserPromptMessages, systemPrompt } from './prompt-test-helper';
 
 type Context = {
   auth: AuthService;
-  module: TestingModule;
+  module: TestingApp;
   db: PrismaClient;
   event: EventBus;
   models: Models;
@@ -143,7 +141,7 @@ let restoreMockCopilotNativeRuntime: (() => void) | undefined;
 
 test.before(async t => {
   restoreMockCopilotNativeRuntime = installMockCopilotRuntime();
-  const module = await createTestingModule({
+  const module = await createTestingApp({
     imports: [
       ConfigModule.override({
         copilot: {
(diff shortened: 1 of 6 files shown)


oersted:
I created issue https://github.com/toeverything/AFFiNE/issues/15246, I was writing my own PR but I reviewed this one and it looks great. Hopefully it can be merged and deployed quickly, the MCP feature is completely broken because of this and the GraphQL API is not really appropriate as an alternative.

PR #15369, opened 2026-07-28
feat(mcp): add constrained database row writes

imxy0101:
## Summary

Adds a dev/canary-only workspace MCP tool, `append_database_row`, for appending a constrained row into an existing AFFiNE database block through the existing server-side document update path.

This is intentionally scoped as an upstreamable implementation seam, not a production-write rollout.

## What changed

- Add `packages/backend/server/src/core/doc/database-row.ts`
  - Applies existing document update data into `Y.Doc`
  - Validates the target block is an AFFiNE database
  - Appends a paragraph row under database `sys:children`
  - Writes validated primitive cells into database `prop:cells`
- Add focused AVA coverage for the generated Yjs shape and validation boundaries
- Add `DocWriter.appendDatabaseRow()` using existing storage/event paths
- Expose `append_database_row` in the MCP provider only for `READ_WRITE` credentials and dev/canary environments

## Safety boundaries

Current patch:

- Does not enable production writes
- Does not create database schemas
- Does not delete anything
- Requires existing document update permission (`Doc.Update`)
- Requires `READ_WRITE` MCP credential
- Validates actual Yjs column metadata
- Validates select/status options against option IDs
- Accepts only constrained primitive cell values
- Rejects non-finite number/date values (`Infinity`, `NaN`)

Before production enablement, this should still get follow-up policy work:

- explicit feature flag
- credential allowlists for tools/docs/database blocks/property IDs
- persistent idempotency
- persistent audit log with source/confirmation references
- rate limits / workspace allowlist if needed

## Verification

Local verification completed:

```text
corepack yarn workspace @affine/server build
→ packages/backend/server/src/index.ts compiled successfully
```

Focused helper tests:

```text
7 tests passed
```

Disposable E2E was also run against an isolated local AFFiNE stack:

```text
patched server build
→ isolated Postgres / Redis / storage
→ real Workspace
→ real READ_WRITE MCP credential
→ real MCP create_document
→ real MCP append_database_row
→ persisted CRDT stream verification
→ server restart
→ persisted row still present
```

Production AFFiNE / real workspace data was not touched.

## Notes for maintainers

The main question here is whether this server-side Yjs row-builder seam is acceptable, or whether BlockSuite should expose a shared server-safe row builder before MCP database writes are added.

If this should live somewhere other than the current copilot MCP provider, I can split/rework the provider layer while keeping the pure doc helper and tests intact.

### DIFF
--- a/packages/backend/server/src/core/doc/__tests__/database-row.spec.ts
+++ b/packages/backend/server/src/core/doc/__tests__/database-row.spec.ts
@@ -1,0 +1,211 @@
+import test from 'ava';
+import * as Y from 'yjs';
+
+import { appendDatabaseRow } from '../database-row';
+
+type FixtureColumn = {
+  data?: Record<string, unknown>;
+  id: string;
+  name: string;
+  type: string;
+};
+
+function toYMap(input: Record<string, unknown>): Y.Map<unknown> {
+  const map = new Y.Map<unknown>();
+  for (const [key, value] of Object.entries(input)) {
+    if (Array.isArray(value)) {
+      const array = new Y.Array<unknown>();
+      array.push(
+        value.map(item =>
+          typeof item === 'object' && item
+            ? toYMap(item as Record<string, unknown>)
+            : item
+        )
+      );
+      map.set(key, array);
+    } else if (typeof value === 'object' && value) {
+      map.set(key, toYMap(value as Record<string, unknown>));
+    } else {
+      map.set(key, value);
+    }
+  }
+  return map;
+}
+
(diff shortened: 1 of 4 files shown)


darkskygit:
due to performance and memory consumption, yjs is not suitable for use on the server, affine uses y-octo to implement server-side document processing

I have currently implemented full blocksuite support in a local branch, but because y-octo’s write functionality has not undergone complete compatibility testing, the related features have not yet been merged

this feature is planned to be merged in 0.28, so server-side document feature prs based on yjs are not currently accepted

Issue #15361, opened 2026-07-27
Proposal: affine-cli - a headless Rust CLI over the local-first store (+ latex/math support for affine_doc_loader)


wongkang01:
## Summary

I built `affine-cli`, a headless Rust CLI over the local-first AFFiNE store, and would like to contribute it (PR ready on my fork). This issue is to gauge interest and to ask where the accompanying doc-parser improvements should land now that the Rust doc parser ships as the published `affine_doc_loader` crate.

## What it does

A single static binary (`tools/affine-cli`, workspace member) that lets agents and scripts create and edit AFFiNE content in the local nbstore without the app running, emitting JSON:

- `workspace create/list`, `doc create/list/read/update/set-title/set-mode/delete`
- Markdown round-trip including `$…$` / `$$…$$` math and `affine:latex` blocks
- `blob put/get/list` (SHA-256 keys, app convention), `search` (nbstore BM25 indexer)
- `diagram add-shape/add-text/add-connector/create --spec` - byte-compatible edgeless surface elements with fractional-index ordering and spec-driven layout
- Ships an agent skill (`skills/affine`) documenting every command and the JSON contract

Safety guards baked in: refuses to write while the app holds the workspace DB open (SQLite WAL `F_GETLK` probe, `--force` override), rejects reserved doc ids (`db$…`, `userdata$…`, root doc), validates geometry before it can poison the edgeless surface, and `diagram create` is atomic (one delta or nothing).

## The interesting CI bit

y-octo's own reader normalizes encodings that real yjs decodes differently (a bare top-level `Any::Array` decodes to its last element in the browser - we hit this in production with connector `labelXYWH`, full postmortem included). The branch adds a workflow that decodes CLI-written binaries with the app-pinned yjs (13.6.21) in Node and shape-asserts them - the only trustworthy guard for the cross-library seam.

## The doc-parser question

The CLI originally consumed `affine_common::doc_parser`. PR #15197 removed that module in favour of the published `affine_doc_loader` crate, whose source doesn't appear to be developed in this repo. My branch therefore vendors the parser under `tools/affine-cli/src/doc_parser`, carrying three improvements that `affine_doc_loader` 0.1.3 doesn't have yet:

[...]

PR #15374, opened 2026-07-29
feat(tools): add affine-cli, a headless CLI over the local-first store

wongkang01:
## What this PR adds

`tools/affine-cli` is a headless Rust CLI for the local-first AFFiNE store. Agents and scripts can create and edit AFFiNE content in the local nbstore while the app is closed. Every command emits JSON. Discussed in #15361.

### Command surface

| Area | Commands |
|---|---|
| Workspace | `workspace create`, `workspace list` (a corrupt DB gives a per-entry error, not a full failure) |
| Docs | `doc create/list/read/update/set-title/set-mode/delete`, `doc add-latex` |
| Markdown | full round-trip, with `$…$` / `$$…$$` math and `affine:latex` blocks |
| Blobs | `blob put/get/list` (SHA-256 content keys, the app convention) |
| Search | `search` on the nbstore BM25 indexer (a CLI-private `cli:doc` index) |
| Diagrams | `diagram add-shape/add-text/add-connector`, `diagram create --spec` (atomic), `diagram repair-labels` |

### Safety guards

- Write lock: commands that write refuse with `"error":"locked"` while another process holds the workspace DB open. The check is an `F_GETLK` probe of the SQLite WAL lock. `search` also takes the lock because it writes index rows. The global `--force` flag overrides the lock.
- The CLI rejects reserved doc ids: `db$…`, `userdata$…`, and the root doc (its id equals the workspace id).
- The CLI validates geometry before it writes. One malformed `xywh` string breaks the full edgeless surface in the app.
- `diagram create` is atomic. The CLI validates the full spec first, then writes one delta or nothing.

### Real-yjs decode check (new workflow)

The y-octo reader normalizes some encodings that real yjs decodes differently. A bare top-level `Any::Array` decodes to its last element in the browser. This caused a real bug: connector `labelXYWH` broke rendering and selection (see `docs/affine-cli-edgeless-render-postmortem.md`). The new `affine-cli-yjs-compat` workflow emits fixtures with the CLI's own writers. It then decodes the fixtures in Node with the app-pinned yjs (13.6.21) and asserts every field. No pure-Rust test can cover this seam.

### Vendored doc parser

`src/doc_parser` is the former `affine_common::doc_parser`, which PR #15197 removed in favor of the published `affine_doc_loader` crate. The CLI needs three things that `affine_doc_loader` 0.1.3 does not have:

1. latex/math markdown parsing and `affine:latex` read-back
2. a fix for standalone `$$…$$` blocks that dropped the text after them in the same paragraph
3. escapes for literal `$` on render, so that prose does not re-parse as math

I am happy to move these three changes into `affine_doc_loader` and delete the vendored copy. See #15361.

### Blast radius

The PR touches only `tools/`, the root workspace `members` list, `Cargo.lock`, `docs/`, and one new workflow file. It does not modify an existing crate or package.

### Verification

- 129 lib tests (this includes the vendored parser round-trip suite) and 4 e2e suites, all green
- `cargo clippy --all-targets -- -D warnings` is clean, `cargo fmt --check` is clean
- The real-yjs chain passes locally: `emit_yjs_fixtures`, then `yjs-compat/check.mjs`
- I verified diagram output against the desktop app (v0.26.3) over CDP. Selection, the connector tool, and rendering all work.

## Summary by CodeRabbit

* **New Features**
  * Added a command-line interface for managing workspaces, documents, diagrams, search, blobs, and LaTeX content.
  * Added Markdown import/export with support for tables, media, embeds, lists, and math.
  * Added diagram creation and editing with shapes, text, connectors, layouts, and label repair.
  * Added structured JSON output, locking safeguards, validation, and resilient error handling.

* **Bug Fixes**
  * Fixed connector label compatibility issues that could corrupt diagrams and cause stale canvas pixels.

* **Documentation**
  * Added CLI guides, command references, design documentation, and a rendering postmortem.

* **Tests**
  * Added end-to-end coverage for CLI workflows and compatibility with Yjs.


wongkang01:
I pushed 1b95e0e5f to address all findings from the automated review.

- **Workflow token surface**: fixed. The job now sets `permissions: contents: read` and checks out with `persist-credentials: false`.
- **Cycle protection in parent-chain walks**: fixed. `get_list_depth`, `nearest_by_flavour`, `has_skipped_markdown_ancestor`, and `block_level` now carry a visited set. A regression test builds a cyclic parent chain and asserts termination.
- **`escape_math_dollars` op-boundary under-escape**: fixed. The escape decision now sees the first character that the next op renders, or the close marker of the current style. A styled neighbor reports a marker char, which can only over-escape (a `\$` still round-trips as a literal `$`), never under-escape. A new unit test covers a trailing `$` before a bold run.
- **Table id regeneration**: fixed. `apply_table_block_props` now reuses row and column ids by position, mints ids only for added rows and columns, removes stale keys, and writes only values that changed. A regression test asserts that a one-cell edit keeps every row and column id.
- **Silent `if let` on the page-stub read-back**: fixed. All three sites now propagate an error instead of leaving an id-less entry in `meta.pages`.
- **Duplicated page-stub insert**: fixed. `add_doc_to_root_doc` and `update_root_doc_meta_title` now call the shared `insert_page_stub`.
- **Malformed fraction digits panic**: fixed. `validate_order_key` rejects non-base-62 fraction digits, and the `midpoint` digit lookups plus the `b_chars[i]` index are now fallible or bounded. A new test covers keys like `a0$`.
- **`open` creates the DB for read paths**: fixed. The new `LocalBackend::open_existing` errors when the database file is absent, and every command except `workspace create` uses it. An e2e test asserts that a wrong `--workspace` id does not create a workspace.

The local gate is green: 133 lib tests plus the e2e suites, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and the real-yjs (13.6.21) decode chain.

donqu1xotevincent (commented):
First of all — thank you for this PR. This is one of the highest-quality external contributions we've received: the storage layer reuses our own code instead of reimagining it, the e2e tests drive the real binary against a real store, the docs are honest about their own limitations, and the yjs-compat harness shows you understood exactly where the risk in this design lives. We did a deep review of the full branch; here is what we found, the good and the bad, and where we'd like to take it from here.

## Why this is moving to draft

1. **The maintainer position from #15361 stands**: @darkskygit won't move forward with CLI/MCP editing until the y-octo ↔ yjs behavioral differences are resolved, since they can make merging of edit operations inaccurate. That work is in progress on our side.
2. **The branch head currently doesn't compile.** Canary bumped y-octo 0.0.3 → 0.1.0 in #15363 (`Any::Object` is now boxed), and the merge commit `dae0c0ea58` pulled that in without a rebuild — `cargo check -p affine-cli` fails with 7 lib + 3 test errors (e.g. `doc_parser/markdown/delta.rs:189`, `doc_parser/value.rs:78`, `write/root_doc.rs:30`, `engine.rs:307/395/618`). CI hasn't surfaced this because the workflow runs were still pending approval for a first-time contributor.

## What's verified good

We checked the PR's claims against the app code rather than taking them on faith. These hold up:

- **Storage goes through `affine_nbstore` itself** (`use-as-lib`), not a reimplementation: `push_update`/`delete_doc`/timestamp collision-retry are the app's own logic, `snapshots` is never written directly, and the path layout + universal-id format (trailing semicolon included) match the electron helper exactly.
- **The `F_GETLK` probe targets the right byte** (SQLite's WAL dead-man-switch at offset 128 of `-shm`), the `unsafe` block is minimal and sound, and the private `cli:doc` FTS index instead of touching `doc:title` is the right call.
- **`fractional_index.rs` is a faithful port of `fractional-indexing@3.2.0`**, and you correctly identified that surface elements use the raw `generateKeyBetween` (via `LayerManager.generateIndex()`), not the jittered V2 wrapper used elsewhere.
- **The e2e suites are real round-trips**: decoding stored SQLite bytes with y-octo *and* re-reading through the production reader, plus a lock test that holds a live WAL connection. The locked-DB + `--force` behavior is genuinely exercised.
- The workflow security posture (`pull_request` trigger, `contents: read`, `persist-credentials: false`) is above our current repo baseline.

## Confirmed defects

Each of these was reproduced against the branch (we wrote throwaway tests where needed), not just eyeballed:

1. **`escape_math_dollars` under-escapes `$` before digits → silent content corruption.** `doc_parser/markdown/delta.rs` (~L340) assumes `$`+digit can't open math, but pulldown-cmark 0.13 disagrees: `a $10-$20 range` exports unescaped, and re-ingest turns it into an inline equation (`latex="10-"`). Insidiously, the markdown string round-trips byte-identical while the doc content changes, which is why the existing currency test (whitespace-blocked closers) passes. Fix: drop the `is_ascii_digit` exemption — escape any `$` followed by non-whitespace. Note the earlier op-boundary fix from the CodeRabbit round is itself correct; this is a different hole.
2. **Table updates wipe app-authored row/column props.** The stale-key sweep in `write/builder.rs:317-341` whitelists only `.rowId`/`.columnId`/`.order`/`.text`, so `prop:columns.<id>.width` and `.backgroundColor` set by the app get deleted from rows/columns that were *retained*. The sweep should key on retained id segments, not exact-key membership. (The id-reuse part of your fix is otherwise correct and its test passes.)
3. **Slice panic on a hostile table key.** `write/builder.rs:374` computes `key[prefix.len()..key.len()-suffix.len()]`; the key `"prop:rows.order"` (prefix+suffix, no id segment) makes that range `10..9` and panics. Same corrupt-doc class the PR hardens elsewhere — guard with `strip_prefix`/`strip_suffix`.
4. **One recursion is still unguarded.** The four visited-set guards you added are real and tested, but `write/update.rs:112-134` (`build_stored_tree`) recurses over `sys:children` with no visited set — a corrupt doc with a child cycle overflows the stack (abort, not `Err`) in `doc update`.
5. **`next_index` breaks on grouped elements.** `engine.rs:471-482` takes the string-max of existing `index` values and passes it raw to `generate_key_between`. Compound indexes like `a1-a0V` (any doc where the user ever grouped canvas elements) win the string-max, then `validate_order_key` rejects `'-'` — every `diagram add-*` fails on such docs. The app strips group suffixes first (`ungroupIndex`, `blocksuite/framework/std/src/utils/layer.ts`); a one-line truncate-at-`'-'` fixes it.

## Where the docs/claims oversell the implementation

- **"diagram create is atomic … one delta or nothing" is only true for the surface.** Every diagram command performs a second, separate write (`ensure_edgeless` → `db$docProperties`, `commands.rs:649-654`, after the element push at e.g. `:678-679`). If the second write fails, the CLI returns `ok:false`/exit 1 **but the element is already persisted** — a caller treating non-zero exit as "nothing happened" is wrong.
- **"Every command emits JSON" fails for the whole clap error class.** `Cli::parse()` (`main.rs:14`) runs before the JSON envelope: missing required flags, conflicting flags, unknown subcommands print plain text to stderr and exit 2 with empty stdout — exactly what SKILL.md tells agents to parse as JSON.
- **The "write lock" is a one-shot pre-flight probe, not a lock.** There's a real TOCTOU window between the probe and the write (SQLite integrity is never at risk, but the description promises more than the mechanism delivers). And on Windows the probe unconditionally returns `false` (`store.rs:146-149`) — the documented `error:locked` contract silently doesn't exist there.
- **Every command — including `doc read` — silently runs sqlx migrations on the user's DB** (`store.rs:87-88` → nbstore `migrate()`). A CLI built from a newer canary than the installed app upgrades `_sqlx_migrations` in place and the older app can then fail to open its own database. This needs a check-and-refuse instead of an implicit migrate.
- Minor: connector endpoints are never checked for existence (disclosed in REFERENCE.md, but trivially guardable since the surface map is already loaded); `diagram create --spec` has no node/edge cap and `separate()` is O(n²·200) with silent non-convergence; every parser invocation mints a fresh y-octo client id, which will permanently bloat state vectors under agent-driven usage.

## The structural concerns (why we can't merge even after the fixes)

1. **The yjs-compat harness doesn't cover the risk that gates this PR.** It asserts full-state updates applied to fresh `Y.Doc`s — but what the app decodes row-by-row from the DB are *raw deltas*, and the deletion-bearing paths (`doc update` structural diff, `--replace`, `remove_doc_from_root`) produce deltas carrying *delete sets*. Not one deletion is ever decoded by real yjs in CI, and delete-set/skip encoding is precisely the historic y-octo↔yjs divergence area. Merge semantics against concurrent app edits (the exact blocker from #15361) are untested and untestable in this harness shape. The harness idea itself is genuinely valuable — see the path forward below.
2. **The drift surface is large and invisible.** The CLI hardcodes ~14 categories of app conventions: 31 block flavour strings, `sys:*`/`prop:*` keys, schema version integers (page v2, surface v5), the `$blocksuite:internal:native$` marker, the xywh wire format, root-doc meta shape, blob SHA-256 keys, etc. The compat workflow only triggers on `tools/affine-cli/**` — a BlockSuite schema change would never run it. Merging means every future schema migration acquires a second consumer with no compile-time or CI-time link.
3. **The vendored `doc_parser` is a 9.2k-line fork of code we deliberately removed in #15197.** (Also: the semantic delta vs the pre-removal code is six changes, not three — the cycle guards, table id reuse, and page-stub hardening from your review-fix round are in there too. Worth knowing for extraction.)

## Path forward

- **Please split the latex/math parser work into a PR against `affine_doc_loader`** once @darkskygit publishes its repo (as discussed in #15361). Our analysis says it extracts cleanly: the math support + `$$…$$` trailing-text fix are naturally one commit, the `$`-escaping a second, and the table/cycle/page-stub fixes a separate PR — **but the digit-escape fix (defect 1) must ship with it**, or the feature lands upstream carrying the same corruption bug. Note the extraction can't be produced by diffing files (the vendored copy was reformatted 2→4-space); regenerate function-scoped hunks against upstream formatting.
- **We'd love to fold the real-yjs fixture harness into the y-octo interop testing effort** — extended to decode raw per-row deltas and delete-set-heavy sequences, it becomes exactly the guard that work needs. Your offer to contribute it there is very welcome.
- **The CLI itself stays on hold** until the y-octo/yjs interop work lands; at that point we'll decide the CLI-vs-MCP surface question together rather than running two parallel tracks.

Keeping this open as a draft so the discussion and the harness work have a home. Thanks again — the postmortem alone (`docs/affine-cli-edgeless-render-postmortem.md`) is the kind of contribution that makes a codebase healthier, and we'd genuinely like to keep working with you on the pieces above.

wongkang01:
Thanks for the deep review and apologies for the delayed response was busy with other commitments.  I have addressed the findings from the review in 12 commits on top of `dae0c0ea58` (head `aee8fdacb2`).

- Build: ported to y-octo 0.1.0, Cargo.lock refreshed, `--locked` passes.
- The five defects: digit escape, retained-id table sweep, `prop:rows.order` guard, `build_stored_tree` visited set, `ungroupIndex` in `next_index`. Each has a regression test.
- Docs vs mechanism: clap errors now go through the JSON envelope (`error:usage`, exit 2); `ensure_edgeless` runs before the element push; "lock" is now documented as a pre-flight check with a Windows warning; and the CLI refuses to open a DB whose `_sqlx_migrations` differ from `affine_schema` (`migration_required` / `db_newer`, opt-in `--allow-migrate`). Connector endpoints are checked, `--spec` is capped, and the CLI keeps one y-octo client id per workspace behind a flock lease.
- Harness: it now emits the exact per-row bytes the CLI pushes, including delete sets, and replays them row by row in real yjs 13.6.21, plus interleaved app edits. No encoding divergence in 302 checks. The interleaving did catch two merge bugs in the CLI's own writer (it replaced `sys:children` and `prop:text` wholesale, so concurrent app edits were dropped). Fixed in `689fbff0b3` by editing the existing `Y.Array`/`Y.Text` in place.
- Drift: the workflow now triggers on the BlockSuite and app paths the CLI hardcodes, and a `schema_drift` test compares the CLI's flavour/version list against `defineBlockSchema(...)`.

On the path forward:

`affine_doc_loader`: the public repo has not been created yet (0.1.7 on crates.io is built from `crates/affine_doc_loader` at `4e4a2c19`, which is not in any public repository), so there is nothing to open a PR against. I prepared the series locally against the 0.1.7 tarball, re-derived per function in upstream formatting. PR A is math + the `$$` trailing-text fix in one commit, then `$` escaping with the digit fix in a second. PR B is the table/cycle/page-stub fixes. Every commit builds and passes tests on y-octo 0.1.0. @darkskygit, could you share the repo link when it is up, or a rough timeline? I will file both the same day.

y-octo harness: I also built a CLI-independent version for y-crdt/y-octo. One `scenarios.json` drives both y-octo and real yjs through the same ops in both directions, over nine delete-set-heavy scenarios. It found three real divergences: a bare `Any::Array` map value reads back as its last element in yjs; y-octo has no `cleanupFormattingGap`, so deleted formatted runs leave `ContentFormat` items live; and splitting a tombstoned run errors, which y-octo #60 fixes (all steps pass with it merged). Happy to open that PR on y-octo if you want it there.

wongkang01:
Pushed `e6ccd832f7` and `351e783cc7`. The canary merge pulled in #15504, which removed the named FTS index the CLI used, so `search` is now index-free and read-only (in-process `memory-indexer`, no writes to the app's tables). The five findings from the last CodeRabbit run are fixed with tests: path-segment validation for `--peer`/workspace ids, `std::fs::File::try_lock` so the lease holds on Windows, a node budget while building the stored tree, stable code-block round trips, and the postmortem now cites the pinned 13.6.21 harness.

wongkang01:
Both fixed in `9d14bc06e1`.

Issue #14033, opened 2025-11-30
Android app Chinese IME broken when using sync cloud


NightFurySL2001:
I am using self host 0.25.2 docker version (latest stable) and 0.24.4 Android app from official Google Play Store. The following bugs are when using a self-hosted instance with cloud sync on Android app.

When inputting Chinese characters using GBoard Chinese language on Android 14 (Pocophone), the original input (usually the pinyin or bopomofo or alphabet) will input into the text area, however selecting any actual character on the keyboard to swap the input string to the actual character will cause all typed text to disappear.

Also, trying to backspace any English text that are already synced to server, when the backspace deleted the all letters, the same word will come back. Sometimes trying to drag and reposition the cursor will duplicate the current suggested word too at the point of cursor. E.g. "Hel∣lo" → click cursor → "HelHello∣lo".

These input issues make the Android app quite literally unusable.

imasaru:
Same issue with Japanese, both local and self-hosted (using Gboard on Android).
Thanks for this software, hope you guys can take a look at this issue!

Issue #12806, opened 2025-06-12
[Bug]: Mobile enters all words twice


itniels:
### What happened?

When I open my AFFiNE on android (using edge) and I start typing, then when I press space it will enter the last word again so `Hi` becomes `HiHi ` and `word` becomes `wordword `

https://github.com/user-attachments/assets/d9d3079a-e0f5-43b9-846b-9b44337afbfe

### Distribution version

Web (https://app.affine.pro)

### What browsers are you seeing the problem on if you're using web version?

Microsoft Edge

### Are you self-hosting?

- [x] Yes

### Self-hosting Version

0.22.3 + 0.22.4

### Relevant log output

```shell

```

itniels:
I have tested this on latest version 0.22.4 and it still happens.

NorkzYT:
@itniels 

For you to read and for reference:
https://github.com/toeverything/AFFiNE/issues/12674#issuecomment-2952578398

itniels:
@NorkzYT Thank you 🙂

NorkzYT:
@itniels 

Please close this issue as it is a duplicate of a different one.

itniels:
Closing this as a duplicate

shanetuohy:
I actually believe this is a different bug to the linked bug (which is to do with the / menu). 

As referenced here -> https://github.com/toeverything/AFFiNE/discussions/11378

there seems to be an issue specifically with Microsoft Switfkey keyboard (a third party keyboard for android). Using the google keyboard with Affine on mobile doesn't seem to have the same issue. 

I can reproduce this issue with the Swiftkey keyboard in each of Firefox, Chrome and the Affine android app.

itniels:
Indeed, I just followed the advice to use FF and same issue, and without swiftkey (using gboard) no issue!
So I will reopen this. Thanks @shanetuohy 🙂

laetha:
For what it's worth I had this issue with Microsoft SwiftKey Keyboard, Heliboard, and FUTO Keyboard. Gboard works fine.

YummYume:
Same for me, on the latest version of the app I have the issue with FUTO keyboard and HeliBoard. I can confirm Gboard works fine, which sucks, because I normally have it disabled.

eddylackmann:
The Issue is still there, also on Android Tablets 😞😞

brunobertoldi:
#13590 fixed this issue for me! ✨🙏✨

Astnoa:
[...]

Issue #7479, opened 2024-07-12
Make columns like Notion's columns


ghost:
### Description

I'm new in the app, i loved the app! it's really good, i suffered to find an app that it's a Notion open-source alternative, and Affine was the perfect sollution for me.

I have a feature idea for the app, can you guys make a column-like display, like what the Notion has, with 2, 3, 4 and 5 columns. I like this feature from Notion, but i want to do a collaborative project with a friend and Notion it's very limited and we can't afford the payment plans. 

We would like to put an image aside of a text, like a lot of websites do.  Me and my friend would appreciate a lot if that feature was added in the app!

### Anything else?

![image](https://github.com/user-attachments/assets/121940d7-2086-448a-85b4-2b73eb89e945)
This is an example that my friend do in his private Notion workspace

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

xmflswood:
I'm looking forward to having this feature.

marcopadillag:
It would be a really good feature.

Jarnooow:
This is the only reason why I can't switch yet. I use columns a lot. Would be amazing if it could be added.

AndreasWidmer:
+1 really looking for this!

walero-moon:
+1

icarus-00:
i was about to write an issue for this, especially that there're no inline images, which annoys me somewhat, but having multi-column support would hopefully fix that

icarus-00:
i'd put that as a feature request, this is something that we hopefully could get before 0.20

Woisol:
Still cant switch now on 02-28. is there any update?

hisuperwen:
+1

zag13:
+1

apertso:
+1

Magasul:
+1

blacklightpy:
+1, one of the reasons I'm not switching yet. I like my Notion Dashboards that utilize the space well.

YoshoTosho:
+1 . One of the reasons why I'm not ready yet to switch. Most pages I have are well divided into columns, and it doesn't fit at all with AFFiNE.

justiceserv:
+1

paufer:
+1

Pascal-Flores:
+1

lovelycbm:
+1

thompsonpat:
The only thing preventing me from switching over. Columns are too important to go without.

[...]

PR #15413, opened 2026-08-01
Fix issue 12806

S-A-Adit:
The PR addresses an issue on Android (particularly Edge/Chromium browsers) where pressing the spacebar to commit a predictive text word causes the word to duplicate (e.g., 'Hi' becomes 'HiHi ').

The root cause was an asynchronous race condition within the event service:
1. `_onCompositionEnd` eagerly set `_isComposing = false` before yielding to `await this.editor.waitForUpdate()`.
2. The browser immediately dispatched a `beforeinput` event for the space key.
3. Because `_isComposing` was already false, the `_onBeforeInput` handler incorrectly processed this event and inserted the text into the CRDT.
4. When `waitForUpdate()` resolved, `_onCompositionEnd` resumed and inserted the text *again*, resulting in the duplication.

Fixes:
- Hardened `_onBeforeInput` to strictly ignore events that are inherently part of the composition cycle by checking `(this._isComposing || event.isComposing)`.
- Deferred toggling `this._isComposing = false` until the absolute end of the `_onCompositionEnd` handler (safely wrapped in a `finally` block), closing the window where rogue `beforeinput` events could be misprocessed during the CRDT reconciliation.

## Summary by CodeRabbit

* **New Features**
  * Added flexible row and column layouts for arranging content.
  * Added slash-menu actions for creating two- to five-column layouts.
  * Added support for adding and editing content within layout columns.
  * Added consistent spacing, sizing, and directional styling for layout containers.

* **Bug Fixes**
  * Improved text composition handling during input and editing.
  * Reduced unexpected changes when composition ends during selection or editor state transitions.

### DIFF
--- a/blocksuite/framework/std/src/__tests__/inline/event-composition.unit.spec.ts
+++ b/blocksuite/framework/std/src/__tests__/inline/event-composition.unit.spec.ts
@@ -1,0 +1,121 @@
+import type * as BlocksuiteGlobalEnv from '@blocksuite/global/env';
+import { expect, test, vi } from 'vitest';
+import * as Y from 'yjs';
+
+vi.mock('@blocksuite/global/env', async importOriginal => {
+  const actual = await importOriginal<typeof BlocksuiteGlobalEnv>();
+  return { ...actual, IS_ANDROID: true };
+});
+
+import { effects } from '../../effects.js';
+import { InlineEditor } from '../../inline/index.js';
+
+effects();
+
+async function setupInlineEditor(text: string) {
+  const yDoc = new Y.Doc();
+  const yText = yDoc.getText('text');
+  yText.insert(0, text);
+
+  const editor = new InlineEditor(yText);
+  const root = document.createElement('div');
+  const outside = document.createElement('div');
+  outside.textContent = 'outside';
+
+  document.body.append(root, outside);
+  editor.mount(root);
+  await editor.waitForUpdate();
+
+  return { editor, root, outside };
+}
+
+function setNativeSelection(range: Range) {
+  const selection = document.getSelection();
+  if (!selection) {
+    throw new Error('Selection is not available');
+  }
+  selection.removeAllRanges();
+  selection.addRange(range);
+}
+
+function clearNativeSelection() {
+  const selection = document.getSelection();
+  selection?.removeAllRanges();
+}
+
+async function teardownInlineEditor(
+  ctx: Awaited<ReturnType<typeof setupInlineEditor>>
+) {
+  clearNativeSelection();
+  ctx.editor.unmount();
+  ctx.root.remove();
+  ctx.outside.remove();
+}
+
+// Regression test for https://github.com/toeverything/AFFiNE/issues/12806:
+// on Android, committing an IME composition with space used to duplicate the
+// composed word because `_isComposing` was cleared before awaiting the
+// post-composition rerender, letting the space-commit `beforeinput` race in
+// and get processed as ordinary (non-composing) input.
+test('compositionend race: a beforeinput arriving mid-await does not duplicate the composed word', async () => {
+  const ctx = await setupInlineEditor('');
+  try {
+    const es = ctx.editor.eventService as any;
+
+    const collapsedRange = ctx.editor.toDomRange({ index: 0, length: 0 });
+    expect(collapsedRange).not.toBeNull();
+    setNativeSelection(collapsedRange!);
+
+    es._onCompositionStart({ data: '' } as CompositionEvent);
+    expect(es.isComposing).toBe(true);
+
+    // A composing keystroke, as the IME builds up "Hi".
+    const composingEvent = {
+      inputType: 'insertCompositionText',
+      data: 'H',
+      dataTransfer: null,
+      isComposing: true,
+      preventDefault: vi.fn(),
+      stopPropagation: vi.fn(),
+      getTargetRanges: () => [],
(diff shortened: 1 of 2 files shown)


donqu1xotevincent:
Thanks for digging into the root cause of #12806 — the analysis of the `_isComposing` race (flag cleared before `await waitForUpdate()`, letting the space-commit `beforeinput` through, then compositionend inserting again) is sound, and the `try/finally` reset plus capturing `_compositionInlineRange` before the await are the right structural fixes.

However, this PR cannot be merged in its current shape. Requesting changes:

### 1. Split this into three PRs (blocking)

This PR bundles three unrelated changes:

- The IME composition fix (#12806) — `framework/std/src/inline/services/event.ts`
- A new multi-column layout feature (#7479) — the whole `blocks/layout` package, models, slash menu, `workspace.gen.ts`, `yarn.lock`
- CI/release workflow hardening — `build-images.yml`, `release-cloud.yml`, `release-desktop-platform.yml`

Please rebase this branch onto `canary` keeping only the `event.ts` change, retitle it as a conventional commit (e.g. `fix(std): prevent duplicate text insertion on Android IME compositionend`), and link #12806. The title currently says "Fix issue 12806" while the branch is `fix-issue-7479` and the description never mentions the layout feature.

### 2. The layout feature is not functional and needs its own design discussion

If you want to pursue #7479, please open a separate draft PR — but note:

- The block components are never registered as view extensions, so `affine:layout-row` / `affine:layout-column` created by the slash menu cannot render.
- `package.json` exports `"./view"` but no `src/view.ts` exists.
- The slash menu action uses `any`, has a commented-out icon placeholder, no i18n, and `store.deleteBlock(model)` destroys the block the user was typing in without preserving focus.
- New block flavours also need clipboard/markdown/HTML adapters and edgeless handling, otherwise copy/paste and export silently drop content.
- No tests. A feature of this size should be aligned with maintainers first.

### 3. The workflow changes don't belong in a contributor bug-fix PR

The least-privilege and codesign-guard changes may have merit, but touching release/signing workflows is out of scope here. Drop them, or propose them separately as a `ci:` PR where they can be reviewed by someone with release context.

### 4. On the fix itself

- The `inlineRange` null-check moved *before* `rerenderWholeEditor()`. Previously the editor still rerendered when `_compositionInlineRange` was null, purging composition DOM artifacts; now it early-returns and may leave the DOM out of sync with the model. Either keep the rerender before the null-check or explain why it's safe.
- With `_isComposing` held true through the `await` window, a `beforeinput` arriving in that window is ignored *without* `preventDefault()` — the native DOM mutation is later wiped by the rerender, so that keystroke is silently dropped rather than duplicated. Please confirm on-device that the committed space isn't lost (it should arrive via `event.data` on compositionend).
- `event.ts` is shared across all platforms. Please run a manual IME regression pass on desktop (Pinyin / Japanese / Korean, on Chrome + Safari + Firefox — Korean on Safari/Firefox is historically the most fragile) and note the results in the PR description.
- Add a regression test simulating the event sequence (compositionstart → beforeinput insertCompositionText → compositionend → beforeinput insertText) so this race can't silently come back.

S-A-Adit:
Thanks for the review — I've pushed these fixes:

- Rebased onto `canary`, dropped the layout feature and CI workflow changes — this branch now only touches `event.ts` (+ a new regression test).
- Fixed the null-check/rerender ordering you flagged, and added `preventDefault()` to the ignored-while-composing path so a dropped keystroke can't happen either.
- `_onBeforeInput` now also honors the native `event.isComposing` flag.
- Added `event-composition.unit.spec.ts`, which reproduces the race and fails without the fix.

Still open: I haven't been able to run a manual IME pass on real devices/browsers (Pinyin/Japanese/Korean × Chrome/Safari/Firefox) — will update once that's done, or happy for a maintainer to run it in the meantime.

Will retitle/update the description to match.

NightFurySL2001:
Related: #14033
does this fix the backspace causing more text appearing bug?

S-A-Adit:
> Related: #14033 does this fix the backspace causing more text appearing bug?

Good catch, and yes — this should fix it, for a specific reason worth spelling out.

#14033's backspace symptom ("delete the word, and it comes back") has the same root cause as the space-commit duplication this PR targets, just manifesting the other direction. Before this fix, `_onBeforeInput`'s composing guard returned early *without* calling `preventDefault()`:

```js
if (this._isComposing) {
  ...
  return; // no preventDefault()
}
```

That let the browser's native `contenteditable` still delete the character in the DOM, while the model (Y.Text) never learned about it — the handler bailed before reaching `transformInput`. DOM and model are now out of sync. The next model-driven rerender (e.g. `_onCompositionEnd`'s `rerenderWholeEditor()`) repaints the DOM from the still-unchanged model, and the "deleted" text visibly comes back.

This PR adds `event.preventDefault()` to that exact branch, so a `beforeinput` arriving while composing is fully suppressed instead of half-applied — no more DOM/model split. Combined with deferring `_isComposing = false` to the `finally` block (so the guard covers the whole async rerender window), a racing backspace during that window is cleanly ignored rather than corrupting sync.

I've added a regression test (`event-composition.unit.spec.ts`) covering this delete-race case specifically, alongside the existing space-commit one. Still haven't been able to do a manual on-device pass, so I'd treat this as "fixed by mechanism, not yet device-confirmed" until that happens.

NightFurySL2001:
Better to rename the pr title to the feature it fixed instead of referencing another issue ID as there are many issues related to IME input.

PR #15444, opened 2026-08-07
Fix/issue 15435

S-A-Adit:
# feat: Add Per-Page Read-Only Mode Toggle

## Description
This pull request introduces a new **Read-Only Mode** toggle that is scoped per-page. This feature is particularly useful for preventing accidental edits and preventing the software keyboard from unexpectedly popping up, which is a common annoyance for users on mobile devices when they only intend to read a document.

The toggle is positioned in the top-right header alongside the Favorite and Info buttons (or Share/Menu buttons on mobile) and is represented by the standard `ViewIcon` (Eye) and `EditIcon` (Pencil) from `@blocksuite/icons/rc`.

## Related Issues
Fixes #15423 

## Type of Change
- [x] ✨ New feature (non-breaking change which adds functionality)

## What was changed
* **Per-Page State Management**: Introduced a local React state (`isReadOnlyMode`) within `DetailPageImpl` (Desktop) and `MobileDetailPageContent` (Mobile) to track the mode independently for each opened page without persisting it globally.
* **Header UI Toggle**: Updated `NormalPageHeader` (Desktop) and `MobileDetailPageHeader` (Mobile) to include an `IconButton` that seamlessly switches between the Pencil and Eye icons depending on the current mode.
* **Editor Integration**: Connected the local `isReadOnlyMode` state directly to `PageDetailEditor`'s `readonly` prop, enforcing the constraint across both Desktop and Mobile environments. 

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] TypeScript typechecks pass locally without errors

Issue #15423, opened 2026-08-03
[Feature Request]:


ToastIsADev:
### Description

Add "**Read & Edit Only**" to AFFiNE. 

### Use case

Everyone can benefit from this ⭐ _"Feature"_. It prevents accidental clicks/inputs especially for mobile users. 

(Where the auto pop up of the keyboard gives mobile user some headaches.)

### Anything else?

Sadly no links and references.

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

### Is your content generated by AI?

- [x] I confirm that the content I submitted was **not** generated by AI / **merely contained minimal** AI edits.

S-A-Adit:
@ToastIsADev and @darkskygit  I would like to start working on this issue

ToastIsADev:
> @ToastIsADev and @darkskygit  I would like to start working on this issue

Please do, I would appreciate it very much. Thanks⭐

Issue #15435, opened 2026-08-06
[Bug]: Date in property impossible to pick before 1970


eddielang:
### What happened?

Hi, 
I am currently testing AFFiNE for my needs (genealogy / history). When creating a new page related to an event that occurred in the 19th century, I wanted to add a new date property containing the actual date of event. Unfortunately, your app does not allow to pick any date before 1970 (which seems to be the limit of Unix timestamp). Moreover, the only option to pick a date is to go through a calendar. No option to fill the full date manually is available.

### Distribution version

Windows x64

### Are you self-hosting?

- [ ] Yes

### Relevant log output

```shell

```

### Is your content generated by AI?

- [x] I confirm that the content I submitted was **not** generated by AI / **merely contained minimal** AI edits.

S-A-Adit:
@eddielang I would like to work on this issue.

PR #15453, opened 2026-08-09
feat(editor): add temporary read-only button

S-A-Adit:
## Description

This PR introduces a per-page read/edit toggle to the document header on mobile, allowing users to explicitly switch between reading mode and editing mode for the current page. 

Previously, it could be difficult to smoothly scroll and read documents on mobile without accidentally triggering the editor. This explicit toggle ensures users have a clean, read-only experience when they want to consume content, while keeping editing easily accessible when they need to make changes.

## Changes Made
- Added a new `isReadOnlyMode` state to the mobile detail page.
- Added a toggle button (`IconButton`) to the `MobileDetailPageHeader` to switch between `ViewIconBlocksuite` and `EditIcon`.
- Passed the `isReadOnlyMode` prop to the editor container to enforce read-only status dynamically.

## Related Issues
- Resolves #15444

## Summary by CodeRabbit

* **New Features**
  * Added a read-only mode toggle for workspace detail pages on desktop and mobile.
  * Users can switch between viewing and editing when permitted, with clear view/edit icons and tooltips.
  * Editing is automatically disabled when users lack permission or when a document is in the trash.


packages/frontend/core/src/desktop/pages/workspace/detail-page/detail-page-header.tsx
```diff
@@ -159,6 +165,14 @@ export function NormalPageHeader({ page, workspace }: PageHeaderProps) {
       <BlocksuiteHeaderTitle inputHandleRef={titleInputHandleRef} />
       <TemplateMark />
       <div className={styles.iconButtonContainer}>
+        {onToggleReadOnly && (
+          <IconButton
+            onClick={onToggleReadOnly}
+            icon={isReadOnlyMode ? <ViewIconBlocksuite /> : <EditIcon />}
+            tooltip={isReadOnlyMode ? t['Read Only']() : t['Editing']()}
```
donqu1xotevincent, line 172: These i18n keys don't exist: neither `Read Only` nor `Editing` is defined in `packages/frontend/i18n/src/resources/en.json`. This compiles only because the i18n wrapper's return type has a `{ [unknownKey: string]: () => string }` index signature — at runtime `i18n.exists(key)` is false and the raw key string is returned, so the tooltip will show hardcoded English in every locale and will never enter the translation pipeline.

Please add proper namespaced keys (e.g. `com.affine.header.toggle-readonly.on` / `.off`) to `en.json` and use the generated typed accessors.

While you're at it, consider action-oriented wording ("Switch to edit mode" / "Switch to read-only mode") — a tooltip on a toggle should describe what clicking does, not the current state.

packages/frontend/core/src/desktop/pages/workspace/detail-page/detail-page-header.tsx
```diff
@@ -159,6 +165,14 @@ export function NormalPageHeader({ page, workspace }: PageHeaderProps) {
...
+            style={{ marginRight: 8 }}
```
donqu1xotevincent, line 173: Inline styles aren't used in this codebase — styling goes through vanilla-extract (`styles.ts`). Also note `iconButtonContainer` already manages spacing between its children, so this `marginRight` likely double-spaces. Please move this into the existing stylesheet (same applies to the mobile counterpart).

packages/frontend/core/src/desktop/pages/workspace/detail-page/detail-page-header.tsx
```diff
@@ -80,6 +82,8 @@ const TemplateMark = memo(function TemplateMark({
 interface PageHeaderProps {
   page: Store;
   workspace: Workspace;
+  isReadOnlyMode?: boolean;
+  onToggleReadOnly?: () => void;
```
donqu1xotevincent, line 86: These props are declared on the shared `PageHeaderProps` but `JournalPageHeader` never consumes them, so journal pages on desktop can't enter read-only mode — while on mobile the toggle renders for journal pages too (the mobile header is shared). Please make the two platforms consistent: either wire the toggle into `JournalPageHeader` or explicitly exclude journals on both.

packages/frontend/core/src/desktop/pages/workspace/detail-page/detail-page-header.tsx
```diff
@@ -27,6 +28,7 @@ import type { AffineDNDData } from '@affine/core/types/dnd';
 import { useI18n } from '@affine/i18n';
 import { track } from '@affine/track';
 import type { Store } from '@blocksuite/affine/store';
+import { EditIcon, ViewIcon as ViewIconBlocksuite } from '@blocksuite/icons/rc';
```
donqu1xotevincent, line 31: Nit: the `ViewIcon as ViewIconBlocksuite` alias is unnecessary — there's no naming conflict in this file (nor in the mobile one). Just import `ViewIcon` directly.

packages/frontend/core/src/desktop/pages/workspace/detail-page/detail-page.tsx
```diff
@@ -314,10 +314,11 @@ const DetailPageImpl = memo(function DetailPageImpl() {
   }, []);
 
   const [dragging, setDragging] = useState(false);
+  const [isReadOnlyMode, setIsReadOnlyMode] = useState(false);
```
donqu1xotevincent, line 317: Confirmed bug (echoing CodeRabbit's finding after checking `detail-page-wrapper.tsx`): `DetailPageWrapper` renders its children without keying by `pageId`, so `DetailPageImpl` is *not* remounted when navigating between docs in the same view. Toggle read-only on doc A, click a link to doc B → B silently opens read-only with no indication why. That's a silent loss of editability, which I'd consider blocking.

Please reset the state when the doc changes, e.g.:
```tsx
useEffect(() => {
  setIsReadOnlyMode(false);
}, [doc.id]);
```
The mobile page has the same issue.

packages/frontend/core/src/desktop/pages/workspace/detail-page/detail-page.tsx
```diff
@@ -314,10 +314,11 @@ const DetailPageImpl = memo(function DetailPageImpl() {
...
 
   const canEdit = useGuard('Doc_Update', doc.id);
 
-  const readonly = !canEdit || isInTrash;
+  const readonly = !canEdit || isInTrash || isReadOnlyMode;
```
donqu1xotevincent, line 321: This `readonly` only reaches `PageDetailEditor`, but the doc title is still editable in read-only mode: `BlocksuiteHeaderTitle` gates editing solely on `useGuard('Doc_Update')` and knows nothing about this local state. So with the toggle on, tapping the title still opens the inline editor — which partially defeats the feature.

Either propagate the state to the title component, or (preferably, see my PR-level comment) implement this via the doc-level readonly channel that the trash flow already uses, which covers all editing surfaces at once.

packages/frontend/core/src/desktop/pages/workspace/detail-page/detail-page.tsx
```diff
@@ -326,6 +327,8 @@ const DetailPageImpl = memo(function DetailPageImpl() {
           page={doc.blockSuiteDoc}
           workspace={workspace}
           onDragging={setDragging}
+          isReadOnlyMode={isReadOnlyMode}
+          onToggleReadOnly={() => setIsReadOnlyMode(prev => !prev)}
```
donqu1xotevincent, line 331: The callback is passed unconditionally. When `canEdit` is false or the doc is in trash, the button still flips its icon/tooltip while `readonly` stays true — the UI is lying. Worse: if a user toggles it in that state and later regains edit permission, the doc stays unexpectedly read-only.

```tsx
onToggleReadOnly={
  canEdit && !isInTrash
    ? () => setIsReadOnlyMode(prev => !prev)
    : undefined
}
```

packages/frontend/core/src/mobile/pages/workspace/detail/mobile-detail-page.tsx
```diff
@@ -381,6 +397,7 @@ const MobileDetailPageContent = ({
   const mode = useLiveData(editor.mode$);
   const [isLandscape, setIsLandscape] = useState(getIsLandscape);
   const [chromeVisible, setChromeVisible] = useState(true);
+  const [isReadOnlyMode, setIsReadOnlyMode] = useState(false);
```
donqu1xotevincent, line 400: Same state-leak issue as on desktop: this component isn't remounted per doc, so the read-only state carries over when navigating to another doc. Please reset on doc change here as well.

packages/frontend/core/src/mobile/pages/workspace/detail/mobile-detail-page.tsx
```diff
@@ -339,6 +347,14 @@ const MobileDetailPageHeader = ({
       contentClassName={styles.headerContent}
       suffix={
         <>
+          {onToggleReadOnly && (
+            <IconButton
```
donqu1xotevincent, line 351: This button has no `tooltip`, no `aria-label`, and no `aria-pressed` — the only signal is the icon swap. Please add a localized `aria-label` for each mode plus `aria-pressed={isReadOnlyMode}` (using the same new i18n keys as the desktop tooltip). Also the inline `style={{ marginRight: 8 }}` should move to the vanilla-extract stylesheet, as on desktop.

packages/frontend/core/src/mobile/pages/workspace/detail/mobile-detail-page.tsx
```diff
@@ -511,13 +528,16 @@ const MobileDetailPageContent = ({
           allJournalDates={allJournalDates}
           handleDateChange={handleDateChange}
           trackScrollTitle={trackScrollTitle}
+          isReadOnlyMode={isReadOnlyMode}
+          onToggleReadOnly={() => setIsReadOnlyMode(prev => !prev)}
```
donqu1xotevincent, line 532: On mobile the effective-editability predicate is wider than desktop: `readonly` is also forced when `!enableKeyboardToolbar` or (`mode === 'edgeless' && !enableEdgelessEditing`). The toggle should be hidden/disabled in *all* of those states, not just `canEdit && !isInTrash` — otherwise it flips the icon without doing anything. Consider hoisting the same predicate `DetailPageImpl` uses and passing the callback only when the doc is actually editable.

donqu1xotevincent (commented):
Thanks for working on this — the underlying request (#15423) is a real pain point, especially the mobile keyboard popping up while scrolling. The overall approach (local state → `readonly` prop) is reasonable, but there are a couple of confirmed bugs and some housekeeping items to address before this is mergeable. Detailed comments inline; PR-level items below.

**PR metadata**
- The linked issue is wrong: #15444 is your previous (closed) PR, not an issue, and the branch name references #15435 which is an unrelated date-picker bug. Please change the body to `Resolves #15423` so the actual feature request gets closed on merge.

**CI**
- The main build/typecheck/e2e workflows haven't run on the latest commit yet (only triage/CLA/CodeRabbit). Once the review items are addressed we can approve the workflow runs.

**Missing pieces**
- No analytics: other header buttons go through `track.$.header.$.*`; this new toggle should emit a typed tracking event as well.
- No tests: an e2e covering "toggle on → body not editable → toggle off → editable again → navigate to another doc → state reset" would lock the behavior in (see the state-leak comment inline).

**Design question worth settling early**
- This adds a permanent button to the desktop header, while the original issue is mobile-focused. Before iterating further, let's confirm the interaction design: mobile-only? Desktop too, but inside the doc menu instead of the top-level header? Persisted per-doc vs. ephemeral? Happy to loop in design on this.

**Architecture alternative to consider**
- `DetailPageWrapper` already implements read-only for trashed docs by setting `doc.blockSuiteDoc.readonly = true` at the doc level. Routing this feature through the same doc-level channel (instead of local state + prop drilling duplicated across desktop and mobile) could be cleaner and would also cover surfaces the current approach misses, e.g. the title editor.

steffenrapp:
> mobile-only? Desktop too, but inside the doc menu instead of the top-level header? Persisted per-doc vs. ephemeral?

I think persisted on mobile and desktop would make sense. Many similar products contain a lock mode per doc. That can also be used once a doc is considered as "complete" to avoid accidental changes.

S-A-Adit:
@donqu1xotevincent 
Superseded by #15522.

This PR was opened from `fix/issue-15435` (a branch name that actually references an unrelated date-picker bug) and linked #15444 (a previous closed PR, not an issue) instead of the real feature request, #15423. Per review feedback, I've renamed/re-based the branch as `fix/issue-15423`, fixed the confirmed bugs (mobile toggle button had no canEdit/trash guard, missing i18n keys, redundant inline styles), added the missing analytics event and e2e test, and opened #15522 targeting the correct issue.

Closing this one in favor of #15522 — please continue the review there.

PR #12679, opened 2025-06-02
feat(editor): add date grouping configurations

NorkzYT:
https://github.com/user-attachments/assets/d5578060-2c8c-47a5-ba65-ef2e9430518b

This PR adds the ability to group-by date with configuration which an example is shown in the image below:

image

## Summary by CodeRabbit

* **New Features**
  * Date-based grouping modes (relative, day, week Sun/Mon, month, year), a date group renderer, and quick lookup for group-by configs by name.

* **Improvements**
  * Enhanced group settings: date sub‑modes, week‑start, per‑group visibility, Hide All/Show All, date sort order, improved drag/drop and reorder.
  * Consistent popup placement/middleware, nested popup positioning, per‑item close-on-select, and enforced minimum menu heights.
  * UI: empty groups now display "No <property>"; views defensively handle null/hidden groups.

* **Tests**
  * Added unit tests for date-key sorting and comparison.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #15469, opened 2026-08-11
fix(data-view): remove duplicate addToGroup assignment logic

MonsterMarian:
duplicit code bug fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed grouping behavior when the selected group has a `null` value.
  * Added validation to ensure items are only added to existing groups.


donqu1xotevincent:
Please sign CLA first so we can get the Merge process going.

donqu1xotevincent:
Thanks for the fix! I dug into the history of this duplication and the change is correct — a few notes and requests below.

## Root cause (worth adding to the PR description)

The duplicated block was left behind by #12679 (f832b28d): the `addToGroup` rewrite appended the new logic at the end of the method but forgot to delete the first half of the old body. Since then `addFn` has run **twice** on every drag. For the idempotent configs (select / number / boolean / date, all effectively `(v, _old) => v`) this was invisible, but multi-select's `addToGroup` has append semantics (`[...old, value]`) — the second run reads the cell value already written by the first run and appends the tag again, which is exactly the kanban tag-duplication bug this PR fixes.

Keeping the **second** block is the right call: the removed block had a `v != null` guard that skipped the Ungroups case, while the retained block handles null group values correctly (e.g. dragging a select card into Ungroups clears the value). Since the old block was a strict subset that ran first, the only behavioral change from this deletion is that `addFn` now runs once — no hidden regressions. I also checked `removeFromGroup`, `moveCardTo`, and `moveGroupTo` for similar leftovers from that refactor; they're clean.

## Requests

1. **Please add a regression test.** This bug was itself introduced by a refactor, and `blocksuite/affine/data-view/src/__tests__/kanban.unit.spec.ts` already has the `GroupTrait` test harness set up, so the cost is low. Suggested cases:
   - group by multi-select, call `addToGroup` for a row that already has tags → assert the target tag appears exactly once in the cell value;
   - (optional) dragging into Ungroups leaves a multi-select value unchanged (`addFn(null, old)` returns `old`).
2. **Flesh out the PR description** — current body is just "duplicit code bug fix". Suggest stating the user-visible symptom (duplicated tags when dragging kanban cards under multi-select grouping), the root cause commit (#12679), and why the second block is the one kept. Link the issue with `Fixes #xxx` if one exists.
3. **CLA** — the `license/cla` check is still pending; the PR can't merge until it's signed.

## Nits (optional)

- The retained comment `// Handle both null and non-null values to ensure proper group assignment` was explaining the difference between the two blocks; now that only one path remains it carries little information and could be dropped.
- Missing blank line between `addToGroup` and `changeGroupMode`.

PR #15486, opened 2026-08-15
feat: custom icons for documents, collections, and folders

eth0-re:
## What this PR does

This PR adds a "Custom" tab to the icon picker. A user can upload an image and use it as the icon of a document, collection, folder, tag, or callout. The existing emoji and built-in icon options do not change.

ezgif-4712e8f80ca9b5fc

### How it works

- **`@affine/component`** — the shared `IconPicker` gets a new `CustomIconPicker` tab.
  - The tab accepts PNG, JPEG, SVG, WebP, and GIF files.
  - Raster files have a 5 MB input cap. The picker resizes each raster to 128 px in the client. It encodes images with transparency as PNG, and opaque images as JPEG. Thus the stored blob stays small for all raster inputs.
  - AFFiNE stores SVG files with no changes, so SVG files have a tighter 500 KB cap.
  - The picker keeps only the first frame of an animated GIF.
  - The tab emits a raw `Blob`. This package cannot access the blob engine, so the caller stores the blob.
- **`@affine/core`** — the app stores and shows the icons.
  - `ExplorerIconService.setIcon` uploads the `Blob` to the workspace blob engine. It writes only the content-addressed blob id to the synced `explorerIcon` table.
  - The workspace DB gets no binary data and no base64 data. A repeated upload of one image maps to one blob, because the id is a content address.
  - New `ExplorerIcon` and `BlobIcon` components and a `useBlobIconUrl` hook show blob icons. They cover the navigation tree, the doc icon picker, and `doc-display-meta` (React and Lit paths).
  - While a blob loads, and when a blob is not available, the components show the document's default icon.
- **blocksuite** — the callout block gets the same feature.
  - `IconData` now stores a `blobId`. This matches the component package.
  - `IconPickerService.onSelect` now also accepts a `Blob`.
  - The callout block uploads the blob through the store's `blobSync` and stores only the blob id.
  - All render paths resolve a blob id to an object URL through one shared cache (`getBlobIconUrl` in `@blocksuite/affine-shared`). The cache keeps only successful results. If a blob from another device did not sync yet, the next render tries again. The icon does not stay blank for the session.
- **server** — the reference projection for blob cleanup now includes icon blobs. See the GC section below.

### Notes for reviewers

Review the PR commit by commit. Each commit covers one area, and each commit passes typecheck and lint alone:

1. `feat(core): store custom icons in the workspace blob engine` — the data model and the storage API: type changes, the `setIcon` upload, and the async call-site changes.
2. `feat(server): track custom icon blobs in doc blob refs` — makes icon blobs visible to blob cleanup. See the GC section below.
3. `feat(component): add custom image tab to the icon picker` — the upload UI, the image processing utility, and its tests.
4. `feat(editor): render custom image icons in callout blocks` — blocksuite rendering and the shared object-URL cache.
5. `feat(core): render custom blob icons across the app` — the navigation tree, the doc icon picker, and the doc-display-meta fallback.

More notes:

- **Why a new resize utility, and not `image-blob-reduce`** — the avatar code in `@affine/core` uses that library. The icon resize lives in `@affine/component`, and that package does not depend on the library. The icon path also selects PNG or JPEG from the transparency of the image, at one small target size. A canvas utility of about 100 lines, with unit tests, was simpler than a new dependency.
- The picker tab labels "Emoji", "Icons", and "Remove" were hardcoded. They are i18n keys now.

### Blob GC: icon blobs are tracked server-side

Blob cleanup deletes a blob when it finds no reference to the blob. It builds its reference table, `doc_blob_refs`, from document content alone (image and attachment `prop:sourceId`, through `affine_doc_loader`). Custom icon blobs get their references from the synced `explorerIcon` table doc and from callout `prop:icon` props. The cleanup job cannot see these references. Without a server change, the job would delete every custom icon blob about 30 days after upload, and a content-addressed blob cannot be restored.

The `feat(server)` commit adds both reference sources to the projection:

- The refs scan set now includes the workspace's `db$<workspace>$explorerIcon` doc. The server projects its rows with `affine_doc_loader::project_orm_records`. The `scope_compiler` already uses this pattern for `docProperties`. Each blob icon row emits one `affine:explorer-icon` reference, with the row id as its key. A missing table doc counts as "no refs", not as a parse failure. Thus a workspace with no custom icons cannot block cleanup.
- Content docs get a supplemental y-octo pass. The pass reads each callout block's `prop:icon` (a nested `Y.Map` or a plain object). The server merges these references with the loader's refs, under the `affine:callout` flavour.

`PARSER_VERSION` increases from 1 to 2. All workspaces then crawl again, so existing icon blobs get references before the next cleanup pass. When a user removes or replaces an icon, the next crawl drops its reference, and cleanup collects the unused blob as normal. Icons get a real lifecycle, not an allowlist exemption.

> We found two pre-existing issues while we traced this code. Both are independent of this PR, and we can file issues for them:
> 1. `document_cleanup` has no guard for `db$` or `userdata$` docs. Synced table docs are never in `meta.pages`, so the job marks them as cleanup candidates.
> 2. Before this change, `db$…` docs reached the refs crawl only as a side effect of that candidate retention.

### Test plan

Unit tests:

- `image.spec.ts` — file validation with the tighter SVG cap, resize dimensions, alpha detection, and the unchanged SVG path.
- `blob-icon-url.unit.spec.ts` — the cache keeps only successful results, and it retries after a null result or a failure.
- `explorer-icon.spec.ts` — a Blob upload persists a blob id, structured icon data is not changed, and removal works.
- `commit-service.spec.ts` — updated for the now-async `setIcon`.
- Server extractor tests — explorer-icon rows (blob, emoji, deleted), callout icons (nested map and plain object, merged with loader refs), and docs without a `blocks` root.

## Summary by CodeRabbit

* **New Features**
  * Added custom image icons for documents, callouts, and navigation items.
  * Added drag-and-drop and file-based uploads with previews, resizing, validation, and localized guidance.
  * Custom icons are stored and displayed consistently across the app.
* **Bug Fixes**
  * Improved fallbacks and retry behavior for unavailable or failed icon images.
  * Ensured asynchronous icon uploads do not overwrite newer selections.
* **Documentation**
  * Added localized labels, status messages, and validation errors for custom icon selection.


donqu1xotevincent:
Thanks for the PR — overall this is high quality work. Content-addressed blob storage instead of base64-in-doc, the server-side GC reference projection shipping in the same PR, decode caps against decompression bombs, and solid unit coverage are all the right calls. I also verified the earlier round of bot feedback (callout selection race, picker requestId, decode limits, lit fallback, `setUrl` reset) is already addressed in the current revision.

That said, I found a few issues that were either missed or only half-fixed. Grouped by priority:

## Must-fix before merge

### 1. The selection race is fixed on the BlockSuite side but not in `@affine/core`

The callout picker got the generation guard (`icon-selection.ts`), but `ExplorerIconService.setIcon` (`packages/frontend/core/src/modules/explorer-icon/services/explorer-icon.ts`) has the exact same race with zero protection:

- User picks a custom image → `blobSync.set` is in flight.
- User quickly picks an emoji (or hits Remove) → written immediately.
- The upload resolves → the stale blob icon **overwrites the newer selection**.

Both call sites (`DocIconPicker`, `node.tsx`) are fire-and-forget, so nothing sequences these writes. This is the same bug class you already acknowledged and fixed for callouts — a per-`(where, id)` generation guard in the service (mirroring your callout fix) would close it.

### 2. A not-yet-synced blob leaves the icon blank indefinitely — the PR description overstates the retry behavior

On a second device, the `explorerIcon` table row typically syncs before the blob does. `useBlobIconUrl` fetches once on mount / `blobId` change; after a `null` result **nothing re-triggers the fetch when the blob later arrives**. The Lit `until()` paths render once as well. A mounted sidebar node may not re-render for a long time, so the icon stays on the fallback — "the next render tries again" only helps if a next render actually happens, and nothing guarantees one.

Suggested fix: subscribe to blobSync state (or retry with backoff) so pending icons resolve when the blob lands. At minimum, please adjust the PR description ("The icon does not stay blank for the session") and file a follow-up issue.

## Questions to resolve before merging the server commit

### 3. Double parse per doc + full re-crawl — has infra signed off?

`extract_refs` in `doc_blob_refs.rs` now does a full y-octo parse of **every content doc** (for callout icons) on top of the `affine_doc_loader` parse — roughly doubling the CPU cost of the refs crawl. Combined with the `PARSER_VERSION` 1→2 bump re-crawling every workspace, rollout will produce a real load spike.

Would it be feasible to fold the callout `prop:icon` extraction into `affine_doc_loader` so each doc is parsed once? If the two-pass approach stays, it would be good to coordinate the re-crawl timing with ops.

### 4. Coupling with the pre-existing `document_cleanup` bug on `db$` docs

You noted that `document_cleanup` has no guard for `db$`/`userdata$` docs and marks them as cleanup candidates. This PR makes `db$<ws>$explorerIcon` the single store for all custom icons, so if cleanup ever progresses from marking to deleting those docs, that's user data loss. Two asks:

- Please confirm the current cleanup job only marks and never deletes `db$` docs, and file the two pre-existing issues you found so they land before this feature ships.
- Minor: because of that same bug, the explorerIcon doc id can appear twice in `load_workspace_doc_ids` (explicit push + the candidates query), so it gets processed twice per crawl. Harmless, but worth a dedupe.

### 5. CodeQL alert #171 needs disposition

`github-advanced-security` flagged "DOM text reinterpreted as HTML" at `custom-icon-picker.tsx:124`. The current code has no `innerHTML`/`dangerouslySetInnerHTML` and React escapes text, so this looks like a stale finding from an earlier revision — but it should be verified and dismissed with a justification so it doesn't block the checks.

## Minor / fine as follow-ups

- Upload failures are swallowed by `.catch(console.error)` everywhere (quota exceeded, offline). The user gets no feedback and the picker silently stays open — a toast would help.
- `IconData` is now duplicated between `@blocksuite/affine-shared` and `@affine/component` with the `blobId` semantics needing to stay in sync manually; worth consolidating eventually.
- With `closeAfterSelect`, the "Choose different image" button is unreachable (the picker closes on the first processed file).
- The file input's `value` isn't reset after a successful selection, so re-picking the same file is a no-op.

## Verified non-issues (for the record)

- **SVG stored verbatim is OK**: the server treats `image/svg+xml` as a dangerous inline MIME (`applyAttachHeaders` forces `Content-Disposition: attachment` + `nosniff`, and the workspaces controller uses it), and the client only renders via ``, which doesn't execute scripts. Client-side SVG sanitization on upload would be nice defense-in-depth but isn't required.
- **Data-model compat is safe**: the old `IconType.Blob { blob: Blob }` was never persistable, and old clients render `null` for the new `{ type: 'blob', blobId }` shape — graceful degradation, no migration needed.
- **`getBlobIconUrl` cache design is sound**: stable promise identity means no `until()` flicker, and never revoking object URLs is bounded and documented. I don't think the heavier eviction machinery suggested earlier is warranted.

PR #15507, opened 2026-08-21
fix(nbstore): support synchronous doConnect and fix BroadcastChannel race condition

aryankumar06:
### Summary
Fixes an unhandled runtime exception (`Error: Connection BroadcastChannelConnection has not been established`) that occurs when initializing the storage worker with `BroadcastChannelAwarenessStorage`.

### Root Cause
1. `AutoReconnectConnection` assumed `doConnect()` must always return a `Promise<T>`.
2. `BroadcastChannelConnection.doConnect()` was declared `async`, which deferred channel initialization across asynchronous microtasks.
3. Synchronous callers during startup (such as `Sync.start()` -> `awarenessStorage.subscribeUpdate()`) accessed `connection.inner` before the microtask settled, causing an unestablished connection error and crashing page load.

### Changes
- **`packages/common/nbstore/src/connection/connection.ts`**: Updated `doConnect()` signature to `Promise<T> | T` and updated `innerConnect()` to synchronously assign `this._inner` and set status to `connected` when `doConnect()` returns synchronously.
- **`packages/common/nbstore/src/impls/broadcast-channel/channel.ts`**: Made `BroadcastChannelConnection.doConnect()` synchronous.
- **`packages/common/nbstore/src/impls/broadcast-channel/awareness.ts`**: Added safeguard in `channel` getter to connect if `idle`.

### Testing
- `vitest run packages/common/nbstore`: All 69 unit tests passed.
- `oxlint` and `oxfmt`: 0 errors, 0 warnings across all files.
- Verified `@affine/web` initializes and mounts without runtime errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection handling for synchronous and asynchronous results, including cross-environment promises.
  * Ensured connection timeouts, aborts, status updates, and errors are handled consistently.
  * Improved immediate subscription and message handling for broadcast-channel awareness storage.

* **Tests**
  * Added coverage for synchronous connections and failures, cross-environment promises, and immediate broadcast-channel interactions.


packages/common/nbstore/src/impls/broadcast-channel/awareness.ts
```diff
@@ -34,6 +34,9 @@ export class BroadcastChannelAwarenessStorage extends AwarenessStorageBase {
     id: this.options.id,
   });
   get channel() {
+    if (this.connection.status === 'idle') {
+      this.connection.connect();
```
donqu1xotevincent: **[P2] Avoid acquiring an unowned connection reference from this getter.**

`connect()` increments `AutoReconnectConnection`'s private ref count, but this implicit acquisition has no matching `disconnect()`. I reproduced the sequence `subscribeUpdate()` from `idle` -> normal lifecycle `connect()` -> lifecycle `disconnect()`, and the connection remains `connected`, leaving the `BroadcastChannel` open. The normal worker path already calls `SpaceStorage.connect()` before using this storage, and `BroadcastChannelConnection.doConnect()` is now synchronous, so the safest fix is to remove this getter-side `connect()`. If lazy connection is required, it needs an explicit symmetric acquire/release owner.

packages/common/nbstore/src/connection/connection.ts
```diff
@@ -86,7 +94,7 @@ export abstract class AutoReconnectConnection<
     }
   }
 
-  protected abstract doConnect(signal?: AbortSignal): Promise<T>;
+  protected abstract doConnect(signal?: AbortSignal): Promise<T> | T;
```
donqu1xotevincent: **[P2] Make the declared contract match the thenables accepted at runtime.**

`isPromiseLike()` and `Promise.resolve(result)` deliberately accept `PromiseLike<T>`, but the abstract method only permits `Promise<T> | T`. A subclass returning a valid thenable fails with TS2416, and the new cross-realm test has to cast the value to `Promise` to hide that mismatch. Please change this to `PromiseLike<T> | T` and remove the cast in the test.

packages/common/nbstore/src/connection/__tests__/auto-reconnection.spec.ts
```diff
@@ -316,3 +318,45 @@ test('connecting timeout', async () => {
   expect(connection.connectCount).toBe(5);
   expect(connection.status).toBe('closed');
 });
+
+test('synchronous connection establishes immediately', () => {
```
donqu1xotevincent, line 324: **[P2] Add a regression test for the actual BroadcastChannel startup path.**

This generic synchronous subclass test does not prevent `BroadcastChannelConnection.doConnect()` from accidentally becoming `async` again; all tests added here would still pass while the original startup crash returned. Please add a test using `BroadcastChannelAwarenessStorage`: call `connection.connect()`, immediately call `subscribeUpdate()` without awaiting a microtask, then unsubscribe/disconnect and assert the connection is `closed`. A synchronous-throw test for the new `try/catch` branch would also cover the new negative path.

donqu1xotevincent (changes requested):
No P0/P1 findings. I found three P2 issues: an unowned connection reference in the awareness getter, a PromiseLike contract mismatch, and missing regression coverage for the actual BroadcastChannel startup path. I verified that the managed synchronous startup path passes, and separately reproduced the implicit reference surviving a normal connect/disconnect lifecycle. Please address these before landing.

aryankumar06 (commented):
Removed the unowned connect() from channel getter — callers must call SpaceStorage.connect() before use

aryankumar06 (commented):
Made test async, used await vitest.waitFor(...) before asserting

PR #15508, opened 2026-08-21
feat(editor): add callout to "Turn into" menu

### LINKED ISSUE #13954: [Bug]: Turn into callout is missing
### What happened?

When you select a text block and open the "Turn into" menu the new callout block is missing. Would be great if that could be added, so existing text or quote blocks could be turned into callouts where useful.
Image

### Distribution version

macOS ARM 64 (Apple Silicon)

### App Version

0.25.5

Safari

### Are you self-hosting?

- [ ] Yes

### Self-hosting Version

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

ywchenwy:
## Summary

Closes #13954

The "Turn into" menu was missing the Callout option. Users had no way to convert an existing paragraph/list/code block into a Callout directly from the slash menu or context toolbar.

## Root Cause

`textConversionConfigs` in `rich-text/src/conversion.ts` had no entry for `affine:callout`, so the option never appeared in the menu.

Additionally, Callout is a **hub block** — its text lives in a child paragraph, not in its own `text` prop. This means the generic `transformModel` path would silently fail, requiring a dedicated conversion handler.

## Changes

- `blocksuite/affine/rich-text/src/conversion.ts` — add callout entry to `textConversionConfigs`
- `blocksuite/affine/blocks/note/src/commands/block-type.ts` — add `transformToCallout` command that:
  1. Creates a new `affine:callout` block at the original position
  2. Moves the original text into a child `affine:paragraph` inside the callout
  3. Deletes the original block

## Before / After

| Before | After |
|--------|-------|
| "Turn into" menu had no Callout option | Callout appears in "Turn into" menu |
| Selecting any block → Turn into showed: Heading, Text, Quote, Divider… | Now also shows: **Callout** |

## Summary by CodeRabbit

* **New Features**
  * Convert selected paragraphs, lists, and code blocks into callouts while preserving text and nested content.
  * Added callout recognition and icon display in rich-text conversion.
  * Conversion results now distinguish between selected blocks and text.
* **Bug Fixes**
  * Prevented invalid, nested, or incompatible callout conversions.
  * The conversion menu now hides Callout when unavailable.
  * Preserved original blocks when conversion fails.
  * Improved undo and redo behavior for callout conversions.

### DIFF
--- a/blocksuite/affine/blocks/note/src/commands/block-type.ts
+++ b/blocksuite/affine/blocks/note/src/commands/block-type.ts
@@ -121,6 +121,84 @@ export const updateBlockType: Command<
     }
     return next({ updatedBlocks: [newModel] });
   };
+  const transformToCallout: Command<{}, { updatedBlocks: BlockModel[] }> = (
+    _,
+    next
+  ) => {
+    if (flavour !== 'affine:callout') return;
+    // When flavour IS 'affine:callout', this command MUST always be terminal
+    // (always call next, never fall through). The generic transformModel branch
+    // that follows unconditionally deletes the source block even when addBlock
+    // fails a schema check — falling through would cause data loss.
+
+    const calloutModels: BlockModel[] = [];
+    const paragraphModels: BlockModel[] = [];
+
+    blockModels.forEach(model => {
+      if (
+        !matchModels(model, [
+          ParagraphBlockModel,
+          ListBlockModel,
+          CodeBlockModel,
+        ])
+      ) {
+        return;
+      }
+
+      const parent = doc.getParent(model);
+      if (!parent) return;
+
+      // [P1a] Callout cannot be nested inside another Callout — skip silently.
+      // We still call next() at the end so the chain does NOT fall through to
+      // the generic transformModel branch (which would delete the source block).
+      if (parent.flavour === 'affine:callout') return;
+
+      const index = parent.children.indexOf(model);
+      const textContent = model.text?.clone();
+
+      // Callout is a hub block — its text lives in a child paragraph,
+      // not in its own text prop, so we create the callout first,
+      // then add a paragraph inside it carrying the original text.
+      const calloutId = doc.addBlock('affine:callout', {}, parent, index);
+      if (!calloutId) return;
+
+      const calloutModel = doc.getModelById(calloutId);
+      if (!calloutModel) return;
+
+      const paragraphId = doc.addBlock(
+        'affine:paragraph',
+        { text: textContent },
+        calloutModel
+      );
+
+      const paragraphModel = paragraphId
+        ? doc.getModelById(paragraphId)
+        : null;
+
+      // [P1b] Re-parent nested children (e.g. indented list items) to the new
+      // paragraph so they remain reachable after the source block is removed.
+      if (paragraphModel && model.children.length > 0) {
+        doc.deleteBlock(model, { bringChildrenTo: paragraphModel });
+      } else {
+        doc.deleteBlock(model, { deleteChildren: false });
+      }
+
+      calloutModels.push(calloutModel);
+      if (paragraphModel) {
+        paragraphModels.push(paragraphModel);
+      }
+    });
+
+    // Always terminal for 'affine:callout' — never fall through to
+    // genericTransform. If no blocks were converted (e.g. all were inside an
+    // existing Callout and skipped), we still call next with empty arrays.
+    const updatedBlocks =
+      paragraphModels.length > 0 ? paragraphModels : calloutModels;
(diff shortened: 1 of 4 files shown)


donqu1xotevincent:
## Review verdict: Request changes

I reviewed the current head (`5358662`) against `canary` and traced the conversion through the toolbar, command chain, block schemas, and store CRUD implementation. The menu wiring is consistent with the existing Callout UI, but the conversion is not safe to merge yet.

### Blocking findings

1. **[P1] Converting content already inside a Callout can delete the source block**

   The Turn into menu renders the new Callout action for the paragraph/list inside an existing Callout, but `CalloutBlockSchema` does not allow a Callout to be a child of another Callout. On that path, `Store.addBlock` catches the schema exception inside `transact` and still returns a generated ID. `transformToCallout` then cannot resolve `calloutModel` and exits without calling `next`, so `CommandManager.try` falls through to the generic `transformModel` branch. That helper repeats the failed add and unconditionally deletes the original model.

   Please:
   - hide or disable Callout conversion when the selected models' parents cannot accept `affine:callout`;
   - preflight every selected source/parent before mutating anything;
   - make the Callout branch terminal on all failure paths so it cannot fall through to generic conversion;
   - harden `transformModel` so the source is deleted only after the replacement model can actually be resolved.

2. **[P1] Nested/indented descendants become unreachable after conversion**

   `doc.deleteBlock(model, { deleteChildren: false })` prevents recursive deletion, but it does not reparent `model.children`. Store CRUD removes the source from its parent and deletes the source record; without `bringChildrenTo`, nested list items and other descendants remain in CRDT storage but disappear from the reachable document tree, rendering, and exports.

   The existing CodeRabbit thread at `block-type.ts:162` is valid and should remain blocking. Please resolve the generated paragraph first, move the source children to it through the schema-validated `bringChildrenTo` path, and only then delete the source.

3. **[P2] Block selection targets the generated paragraph instead of the Callout container**

   Returning `paragraphModel` in `updatedBlocks` fixes text focus, but `updatedBlocks` is also the public result of `updateBlockType` and the input to `focusBlock`. For an original BlockSelection, the editor therefore selects the inner paragraph rather than the new Callout. Pressing Delete immediately afterward removes the paragraph and leaves an empty Callout.

   Please keep Callout container models in `updatedBlocks` and carry or derive separate paragraph targets for `focusText`.

### Required regression coverage

Please add focused E2E coverage for:

- paragraph/list conversion with nested descendants, including reachability and exact tree shape;
- Turn into → Callout from content already inside a Callout, proving no source data is removed;
- TextSelection and BlockSelection target IDs after conversion;
- mixed/multi-block selection with all-or-nothing behavior;
- immediate Delete after conversion;
- undo/redo restoring the exact block tree and selection;
- preservation of rich-text deltas/formatting.

### CI/CD

After the fixes and tests are pushed, please approve and run both **Build & Test** and **PR Title Lint**. The required `3, 2, 1 Launch` and `Check pull request title` gates have not run on the current head (`action_required`, no jobs). No manual deploy or release workflow is needed for this editor-only PR; normal post-merge `canary` automation should handle CD.

Until the two data-loss paths and the selection contract are fixed and the required checks pass, I do not think this PR is merge-ready.

ywchenwy:
> Thanks for the detailed review! All blocking findings have been addressed:
> 
> **[P1a] Nested Callout fallthrough / data loss**
> Added a `parent.flavour === 'affine:callout'` preflight check in `transformToCallout`. The command is now always terminal when `flavour === 'affine:callout'` — it always calls `next()` regardless of outcome, so it can never fall through to the generic `transformModel` branch.
> 
> **[P1b] Nested descendants becoming unreachable**
> Replaced `deleteBlock(model, { deleteChildren: false })` with `deleteBlock(model, { bringChildrenTo: paragraphModel })` when the source block has children, so nested list items are properly reparented to the new paragraph before the source is deleted.
> 
> **[P2] BlockSelection targeting inner paragraph instead of Callout container**
> `transformToCallout` now passes both `updatedBlocks` (paragraph models, for `focusText`) and `calloutModels` (Callout containers) to the context. `focusBlock` reads `calloutModels` when present, so `Delete` after a `BlockSelection` conversion removes the whole Callout, not just the inner paragraph.
> 
> **UI — hide Callout option inside existing Callout**
> Added a filter in `toolbar.ts` `conversionsActionGroup`: when all selected models' parents are `affine:callout`, the Callout entry is hidden from the Turn-into menu.
> 
> **E2E regression coverage**
> Added 6 focused tests in `callout.spec.ts` covering: plain paragraph conversion, nested list reachability after conversion, no-data-loss when already inside a Callout, Delete after conversion, undo/redo, and rich-text delta preservation.

PR #15518, opened 2026-08-24
fix(copilot): cap embedding batch size to 64 to avoid provider limits

SCys:
### Summary
Google Gemini embedding API (`batchEmbedContents`) enforces a maximum limit of 100 requests per batch (`* BatchEmbedContentsRequest.requests: at most 100 requests can be in one batch`).

Previously, `PROVIDER_BATCH` in the native embedding worker was set to `128`, causing batch embed calls to fail on documents with > 100 chunks when using Gemini (or other providers with < 128 limits such as Cohere).

This PR:
1. Reduces `PROVIDER_BATCH` in the native embedding worker from 128 to 64.
2. Adds chunk batching in `CapabilityRuntime.embed` to prevent large inputs from exceeding single-batch limits.

## Summary by CodeRabbit

* **Improvements**
  * Embedding requests are now processed in smaller batches of up to 64 items.
  * Larger embedding workloads are automatically split into multiple requests and combined into a single result.
  * This improves handling of large inputs and helps provide more consistent embedding operations.

### DIFF
--- a/packages/backend/native/src/runtime/backend_runtime/embedding/worker.rs
+++ b/packages/backend/native/src/runtime/backend_runtime/embedding/worker.rs
@@ -16,7 +16,7 @@ const MAX_INPUT_BYTES: usize = 50 * 1024 * 1024;
 const MAX_TEXT_BYTES: usize = 64 * 1024 * 1024;
 const MAX_TOKENS: usize = 1_000_000;
 const MAX_CHUNKS: usize = 2048;
-const PROVIDER_BATCH: usize = 128;
+const PROVIDER_BATCH: usize = 64;
 
 pub(super) struct WorkerHandle {
   stop: watch::Sender<bool>,
--- a/packages/backend/server/src/plugins/copilot/runtime/capability-runtime.ts
+++ b/packages/backend/server/src/plugins/copilot/runtime/capability-runtime.ts
@@ -365,17 +365,38 @@ export class CapabilityRuntime {
     input: string | string[],
     options: CopilotEmbeddingOptions = {}
   ) {
-    const result = (await this.execute(
-      'index.embedding',
-      buildLlmEmbeddingRequest({
-        model: 'route-selected',
-        inputs: Array.isArray(input) ? input : [input],
-        dimensions: options.dimensions,
-      }),
-      {},
-      options
-    )) as { embeddings: number[][] };
-    return result.embeddings;
+    const inputs = Array.isArray(input) ? input : [input];
+    const BATCH_SIZE = 64;
+    if (inputs.length <= BATCH_SIZE) {
+      const result = (await this.execute(
+        'index.embedding',
+        buildLlmEmbeddingRequest({
+          model: 'route-selected',
+          inputs,
+          dimensions: options.dimensions,
+        }),
+        {},
+        options
+      )) as { embeddings: number[][] };
+      return result.embeddings;
+    }
+
+    const embeddings: number[][] = [];
+    for (let i = 0; i < inputs.length; i += BATCH_SIZE) {
+      const batch = inputs.slice(i, i + BATCH_SIZE);
+      const result = (await this.execute(
+        'index.embedding',
+        buildLlmEmbeddingRequest({
+          model: 'route-selected',
+          inputs: batch,
+          dimensions: options.dimensions,
+        }),
+        {},
+        options
+      )) as { embeddings: number[][] };
+      embeddings.push(...result.embeddings);
+    }
+    return embeddings;
   }
 
   async rerankConfigured(_modelId: string) {


donqu1xotevincent:
Sign CLA first so we can get it into the review process.

packages/backend/native/src/runtime/backend_runtime/embedding/worker.rs
```diff
@@ -16,7 +16,7 @@ const MAX_INPUT_BYTES: usize = 50 * 1024 * 1024;
 const MAX_TEXT_BYTES: usize = 64 * 1024 * 1024;
 const MAX_TOKENS: usize = 1_000_000;
 const MAX_CHUNKS: usize = 2048;
-const PROVIDER_BATCH: usize = 128;
+const PROVIDER_BATCH: usize = 64;
```
donqu1xotevincent, line 19: Reducing the batch size makes a 2048-chunk projection perform 32 sequential provider calls under a fixed, non-renewed five-minute lease. The operation also permits 120 seconds of extraction and a 30-minute timeout per provider target, while commit rejects any expired lease and eventually fails after repeated `commit_failed` retries. Please add lease-token-fenced heartbeat renewal or a lease-aware total deadline, and cover a delayed 2048-chunk projection in a test.

packages/backend/server/src/plugins/copilot/runtime/capability-runtime.ts
```diff
@@ -365,17 +365,38 @@ export class CapabilityRuntime {
...
+        {},
+        options
+      )) as { embeddings: number[][] };
+      return result.embeddings;
+    }
+
+    const embeddings: number[][] = [];
+    for (let i = 0; i < inputs.length; i += BATCH_SIZE) {
+      const batch = inputs.slice(i, i + BATCH_SIZE);
+      const result = (await this.execute(
```
donqu1xotevincent, line 387: Each batch starts a new unpinned route decision. With multiple compatible BYOK embedding targets, the runtime may fail over to profile/model B for one batch and use A for the next, then concatenate vectors from different embedding spaces. The Rust worker prevents this with an expected fingerprint check. Please resolve and pin one target for the entire logical embed call, or move batching behind one native route decision and reject any route/model change.

packages/backend/server/src/plugins/copilot/runtime/capability-runtime.ts
```diff
@@ -365,17 +365,38 @@ export class CapabilityRuntime {
...
+        'index.embedding',
+        buildLlmEmbeddingRequest({
+          model: 'route-selected',
+          inputs: batch,
+          dimensions: options.dimensions,
+        }),
+        {},
+        options
+      )) as { embeddings: number[][] };
+      embeddings.push(...result.embeddings);
```
donqu1xotevincent, line 397: Please validate that every batch returns exactly `batch.length` embeddings, with one consistent finite vector dimension, before appending. `llm_adapter` 0.2.21 preserves the provider array as returned and does not enforce response count; OpenAI `data[].index` is also ignored. A short, extra, or reordered response can silently break input-to-vector alignment. The Rust worker already rejects an invalid embedding count.

packages/backend/server/src/plugins/copilot/runtime/capability-runtime.ts
```diff
@@ -365,17 +365,38 @@ export class CapabilityRuntime {
...
+          dimensions: options.dimensions,
+        }),
+        {},
+        options
+      )) as { embeddings: number[][] };
+      return result.embeddings;
+    }
+
+    const embeddings: number[][] = [];
+    for (let i = 0; i < inputs.length; i += BATCH_SIZE) {
```
donqu1xotevincent, line 385: `CopilotEmbeddingOptions` exposes an `AbortSignal`, but this loop never checks it and the non-stream native execute path cannot cancel. If the caller aborts after batch 1, all remaining batches are still dispatched. At minimum call `signal.throwIfAborted()` before each batch and after each await; ideally propagate cancellation into native execution.

packages/backend/server/src/plugins/copilot/runtime/capability-runtime.ts
```diff
@@ -365,17 +365,38 @@ export class CapabilityRuntime {
...
-        model: 'route-selected',
-        inputs: Array.isArray(input) ? input : [input],
-        dimensions: options.dimensions,
-      }),
-      {},
-      options
-    )) as { embeddings: number[][] };
-    return result.embeddings;
+    const inputs = Array.isArray(input) ? input : [input];
+    const BATCH_SIZE = 64;
```
donqu1xotevincent, line 369: This batching loop has no overall input, byte, token, batch-count, or execution-time limit. Very large input arrays are converted into an effectively unbounded sequence of provider calls instead of being rejected early. Please align this path with the Rust worker's resource limits before dispatching the first batch.

donqu1xotevincent (changes requested):
The provider-limit rationale and the choice of 64 are reasonable, but this implementation crosses several correctness boundaries that are not currently handled. The native change doubles the worst-case number of sequential provider calls without renewing the fixed five-minute projection lease, while the TypeScript batching path re-routes every batch independently, does not validate response cardinality or vector invariants, and continues dispatching batches after cancellation. Please add lease-token-fenced renewal or a lease-aware total deadline, pin one embedding route for the entire logical request, validate every batch before concatenation, propagate cancellation, and add boundary/failure tests before merging.

PR #15520, opened 2026-08-24
fix(editor): correct markdown export of database blocks

tikimo:
## What

Exporting a doc that holds a database block produces markdown that is wrong in three ways. This fixes all three, plus two smaller output defects that come from the same path.

### 1. The header row can sit above the wrong columns

A row is built as `[row title, ...non-title columns]`, but `processTable` passed the raw `columns` array as the headers. When the title column is not first, every name sits above the wrong column. A database with no title column produces a header row one cell short of every data row.

The headers are now built in the same order the rows are, with the title first.

### 2. An empty cell aborts the export of the whole document

`property.config.rawValue.toString(...)` can return `undefined`. That value went through unchanged, the adapters then read `value.delta` on it and threw, and the export of the entire document failed — not just the table.

The existing `try`/`catch` does not cover this, because a plain `undefined` return is not a throw. A `?? ''` inside the `try` handles it.

### 3. One long cell makes every row that wide

`gfmTableToMarkdown()` pads every cell out to the widest value in its column. A single long cell, such as a URL, turns a small table into hundreds of columns of trailing spaces. `tablePipeAlign: false` stops the padding; the output still parses as the same table.

### 4 and 5. Serializer output in `_astToMarkdown`

- A table cell cannot hold a newline, so the serializer emits `&#xA;`. That entity now becomes a space.
- The serializer escapes `&` in a link destination because it can start a character reference. That puts `\&` in every query string, and the link breaks when a reader copies it out. A bare `&` in a destination parses back as itself, so unescaping is safe. Only destinations are touched, so a literal `\&` in body text stays escaped.

## Tests

`blocksuite/affine/all/src/__tests__/adapters/markdown.unit.spec.ts` — the `snapshot to markdown > table` fixture now puts the title column third instead of first. The expected markdown does not change, because a row always starts with the title. Without the header fix that test fails:

```
- | Title | Status | Date | Number | ...
+ | Status | Date | Title | Number | ...
```

The expected table and the expected reference links were updated for the padding and escaping changes.

Run locally on this branch:

```
vitest --run blocksuite/affine/all/src/__tests__/adapters/markdown.unit.spec.ts   75 passed
vitest --run blocksuite/affine/all/src/__tests__/adapters/plain-text.unit.spec.ts 14 passed
vitest --run blocksuite/affine/all/src/__tests__/adapters/html.unit.spec.ts       40 passed
vitest --run blocksuite/affine/all/src/__tests__/adapters/notion-html.unit.spec.ts 17 passed
tsc -b blocksuite/affine/blocks/database/tsconfig.json   ok
tsc -b blocksuite/affine/shared/tsconfig.json            ok
oxfmt --check, oxlint --deny-warnings                    clean
```

## How it was found

A self-hosted 0.26.2 instance could not export any doc containing a database block. The three fixes were applied to a 0.26.2 build first and confirmed against the real document, then rebased onto `canary`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Markdown table import and export by placing title columns first and providing a default title when needed.
  - Prevented missing table values from producing invalid serialized content.
  - Corrected handling of newlines in table cells and ampersands in linked-page URLs.
  - Preserved literal entities and escaping consistently across links, images, and code content.
- **Improvements**
  - Simplified exported table formatting by removing automatic column padding.

### DIFF
--- a/blocksuite/affine/all/src/__tests__/adapters/markdown.unit.spec.ts
+++ b/blocksuite/affine/all/src/__tests__/adapters/markdown.unit.spec.ts
@@ -2483,12 +2483,6 @@ hhh
         },
         columns: [
           {
-            type: 'title',
-            name: 'Title',
-            data: {},
-            id: 'block:2VfUaitjf9',
-          },
-          {
             type: 'select',
             name: 'Status',
             data: {
@@ -2517,6 +2511,12 @@ hhh
             name: 'Date',
             data: {},
             id: 'block:5cglrBmAr3',
+          },
+          {
+            type: 'title',
+            name: 'Title',
+            data: {},
+            id: 'block:2VfUaitjf9',
           },
           {
             type: 'number',
@@ -2610,10 +2610,10 @@ hhh
     };
 
     const md = `\
-| Title  | Status      | Date       | Number | Progress | MultiSelect | RichText                    | Link               | Checkbox |
-| ------ | ----------- | ---------- | ------ | -------- | ----------- | --------------------------- | ------------------ | -------- |
-| Task 1 | TODO        | 2023-12-15 | 1      | 65       | test1,test2 | [test2](https://google.com) | https://google.com | True     |
-| Task 2 | In Progress | 2023-12-20 |        |          |             | test1                       |                    |          |
+| Title | Status | Date | Number | Progress | MultiSelect | RichText | Link | Checkbox |
+| - | - | - | - | - | - | - | - | - |
+| Task 1 | TODO | 2023-12-15 | 1 | 65 | test1,test2 | [test2](https://google.com) | https://google.com | True |
+| Task 2 | In Progress | 2023-12-20 | | | | test1 | | |
 `;
     const mdAdapter = new MarkdownAdapter(createJob(), provider);
     const target = await mdAdapter.fromBlockSnapshot({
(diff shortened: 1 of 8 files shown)


donqu1xotevincent:
I reviewed the current head (`f69ed177`) against the latest `canary` (`8e8a781`). The database header ordering and null-value handling are directionally good, but I don't think this version is ready to merge because the Markdown post-processing can silently change user content.

### 1. [P1] Scope Markdown normalization to AST nodes

`_astToMarkdown()` runs both replacements over the complete serialized document, rather than only over database table cells and link destinations. Using the repository's locked remark dependencies, I could reproduce these round-trip failures:

- Literal `&#xA;` in normal text, inline code, or fenced code is changed instead of preserved.
- `https://example.com/?x=1&copy;=2` is stringified safely as `...?x=1\&copy;=2`; the new replacement removes the backslash, and AFFiNE's parser then reads the URL back as `...?x=1©=2`.
- Link-shaped text inside code is also rewritten.
- A destination containing escaped parentheses, such as `/(foo)?a=1&b=2`, is not handled consistently because the regex stops at the first parenthesis.

This is semantic data loss, not just a formatting difference. I suggest removing the whole-document `.replace()` calls. Normalize real newlines only in text nodes under `tableCell` before serialization, excluding `code`, `inlineCode`, and HTML. Keep remark's destination escaping unless a node-scoped link/image serializer can preserve parse/stringify round trips for character-reference-looking ampersands and parenthesized URLs.

### 2. [P2] Preserve an intentionally blank title-column name

In `processTable()`, `titleColumn?.name || 'Title'` converts a persisted empty name into `Title`. AFFiNE currently allows `nameSet('')` and stores that value, so this changes user-authored data across Markdown, HTML, and plain-text export.

Please use:

```ts
name: titleColumn?.name ?? 'Title',
```

so the fallback applies only when the title column/name is actually missing.

### Regression coverage needed before merge

- Body text, inline code, and fenced code containing literal `&#xA;`.
- Link and image destinations containing `&copy;`, numeric character-reference shapes, and escaped parentheses, asserting parse/stringify URL equality.
- A table with no title column, a title column with `name: ''`, and a real table-cell newline.
- A row whose cell exists but has `value: undefined`; the current changed fixture does not export that row, so the new `toString(...) ?? ''` path is not exercised.

A practical patch split would be to keep the safe database/header/null-value and GFM table-padding changes, remove the global Markdown rewrites from this PR, and handle destination formatting separately with AST-scoped logic and round-trip tests.

tikimo:
hanks, you were right on all three. Pushed 72e93e89.

Both whole-document replaces are gone. Newlines are now flattened on text nodes under tableCell, skipping code, inlineCode and html. Destinations go through a node-scoped link/image serializer that only drops the backslash when the & does not complete a character reference, so &copy; and &amp; keep theirs and escaped parens are handled. processTable uses ??.

Tests added for every case you listed, including a link cell whose value is undefined — that path was unreachable before, since select has its own ?? ''. Shared 147, markdown 77, plain-text 14, html 40, notion-html 17, all passing; tsc, oxfmt and oxlint clean.

Kept it as a second commit rather than splitting, since the unsafe code is removed rather than deferred. Happy to split if you still prefer.

darkskygit:
Thank you for your PR. Please sign the CLA so we can proceed with the merge.

PR #15522, opened 2026-08-25
fix(core): add per-page read/edit toggle in header

### LINKED ISSUE #15423: [Feature Request]:

S-A-Adit:
Resolves #15423

## Summary
Adds a per-doc read/edit toggle button to both the desktop and mobile document headers, so a doc can be locked into a read-only view without navigating away — primarily to avoid the mobile on-screen keyboard popping up while scrolling.

## Relation to #15453
This replaces #15453, which was opened from a branch named `fix/issue-15435` (an unrelated date-picker bug) and linked the wrong issue (#15444, which is actually a previous closed PR, not an issue). This PR carries forward all of that PR's code, rebased on latest `canary`, plus fixes for the review feedback left on #15453:

- **Mobile toggle guard**: the mobile toggle button had no `canEdit`/trash guard at all (desktop already had one) — both platforms now hide/disable the toggle when the user can't edit or the doc is trashed.
- **i18n**: added the missing `com.affine.header.readonly-toggle.read-only` / `...editing` translation keys (previously referenced keys that did not exist in `en.json`).
- **Styling**: removed redundant inline `style={{ marginRight: 8 }}` on both header buttons — the surrounding containers already apply the correct `gap`.
- **Analytics**: added a typed `toggleReadOnlyMode` event under `track.$.header.actions`, following the existing `toggleFavorite: { on: boolean }` convention.
- **Tests**: added `tests/affine-local/e2e/readonly-toggle.spec.ts` covering toggle-on → body not editable → toggle off → editable again → navigate to a new doc → no state leak.

## Known follow-up (not in this PR)
As raised in review, routing this through `doc.blockSuiteDoc.readonly` (the same channel `DetailPageWrapper` already uses for trashed docs) instead of local `isReadOnlyMode` state duplicated across desktop/mobile would also fix the title editor staying editable while the body is read-only. Keeping this PR scoped to the existing local-state approach; open to a follow-up if the team wants the architecture change. Interaction design (mobile-only vs. desktop+mobile, persisted vs. ephemeral) is also still open for design feedback.

## Test plan
- [x] `tsc --noEmit` clean on all touched files
- [x] New e2e test covering the toggle/reset flow
- [ ] CI (build/typecheck/e2e workflows)

## Summary by CodeRabbit

* **New Features**
  * Added a read-only mode toggle for desktop and mobile document editors.
  * Users can switch between viewing and editing modes from the page header.
  * Read-only mode resets when navigating to a different document.
  * Added localized tooltips and icons for the toggle action.

* **Bug Fixes**
  * Prevented editing while read-only mode is enabled.
  * Restricted the toggle for documents that cannot be edited or are in the trash.


packages/frontend/core/src/mobile/pages/workspace/detail/mobile-detail-page.tsx
```diff
@@ -511,13 +543,18 @@ const MobileDetailPageContent = ({
           allJournalDates={allJournalDates}
           handleDateChange={handleDateChange}
           trackScrollTitle={trackScrollTitle}
+          isReadOnlyMode={isReadOnlyMode}
+          onToggleReadOnly={
+            canEdit && !isInTrash ? onToggleReadOnly : undefined
```
darkskygit, line 548: Use the same editability check as `DetailPageImpl` here. Include the keyboard toolbar and edgeless editing flags. Otherwise, the button can show “Editing” while the editor remains read-only.

tests/affine-local/e2e/readonly-toggle.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+  await expect(paragraph).toContainText('before toggle after toggle');
+
+  // navigating to a different doc must not leak the read-only state
+  await clickNewPageButton(page, 'Readonly Toggle Doc B');
+  await waitForEditorLoad(page);
+  await pressEnter(page);
+  await type(page, 'doc b content');
+
+  const paragraphB = page.locator('affine-paragraph rich-text').first();
+  await expect(paragraphB).toContainText('doc b content');
```
darkskygit, line 47: This test exits read-only mode before navigation, so it does not test the reset. Enter read-only mode again before opening Doc B. Please also cover the mobile path.

packages/frontend/core/src/desktop/pages/workspace/detail-page/detail-page.tsx
```diff
@@ -314,10 +314,20 @@ const DetailPageImpl = memo(function DetailPageImpl() {
...
+
+  const onToggleReadOnly = useCallback(() => {
+    setIsReadOnlyMode(prev => !prev);
+    track.$.header.actions.toggleReadOnlyMode({ on: !isReadOnlyMode });
+  }, [isReadOnlyMode]);
 
   const canEdit = useGuard('Doc_Update', doc.id);
 
-  const readonly = !canEdit || isInTrash;
+  const readonly = !canEdit || isInTrash || isReadOnlyMode;
```
darkskygit, line 330: This mode only makes the editor body read-only. The title still accepts edits because `BlocksuiteHeaderTitle` only checks `Doc_Update`. Please use the same read-only state for the body and title.

PR #15524, opened 2026-08-25
fix(server): send User-Agent on GitHub OAuth fetches

### LINKED ISSUE #15521: [Bug]: GitHub OAuth login fails with 403 "missing User-Agent header" due to allowedHeaders restriction in safeFetch
## 📋 Summary
Self-hosted GitHub OAuth login fails during the token-exchange step. The server's outbound request to `https://github.com/login/oauth/access_token` gets rejected by GitHub with `403 Request forbidden by administrative rules` because the request has no `User-Agent` header. Root cause traced to `allowedHeaders` in the OAuth provider's `fetchOptions()`, which does not include `user-agent`, combined with the native `safeFetch` (SSRF-protection) client stripping any header not in that list.

## Environment
- AFFiNE version: 0.27.4 (self-hosted, latest at time of report)
- Deployment: Docker, behind nginx reverse proxy
- Domain: custom domain over HTTPS on non-standard port (e.g. `https://example.com:3010`)
- Provider: GitHub OAuth App

## Steps to Reproduce
1. Configure a GitHub OAuth App with callback URL `https://<host>/oauth/callback`.
2. Set `OAUTH_GITHUB_CLIENT_ID` / `OAUTH_GITHUB_CLIENT_SECRET` on the self-hosted server.
3. From the AFFiNE login page, click "Sign in with GitHub".
4. Authorize the app on GitHub's consent screen.
5. GitHub redirects back to `https://<host>/oauth/callback?code=...`.
6. AFFiNE's frontend POSTs to `/api/oauth/callback` with the code.

## Expected Result
Server exchanges `code` for an access token and completes GitHub sign-in.

## Actual Result
`POST /api/oauth/callback` returns:

```json
{
  "status": 400,
  "code": "Bad Request",
  "type": "BAD_REQUEST",
  "name": "INVALID_OAUTH_CALLBACK_CODE",
  "data": {
    "status": 403,
    "body": "\r\nRequest forbidden by administrative rules. Please make sure your request has a User-Agent header (https://docs.github.com/en/rest/overview/resources-in-the-rest-api#user-agent-required). Check https://developer.github.com for other possible causes.\r\n"
  }
}
```

The `redirect_uri` and OAuth App configuration were verified correct — the callback reaches the server fine; the failure happens on the server's own outbound call to GitHub.

## Root Cause Analysis
Traced through the source:

1. `packages/backend/server/src/plugins/oauth/providers/github.ts` → `getToken()` calls `postFormJson('https://github.com/login/oauth/access_token', ...)`.
2. `postFormJson` / `fetchJson` in `packages/backend/server/src/plugins/oauth/providers/def.ts` call `safeFetch(url, init, this.fetchOptions(url))`, where:
   ```ts
   protected fetchOptions(_url: string | URL): SafeFetchOptions {
     return {
       timeoutMs: 10_000,
       maxRedirects: 3,
       maxBytes: 1024 * 1024,
       allowedHeaders: ['authorization', 'content-type', 'accept'],
     };
   }
   ```
   `user-agent` is not in `allowedHeaders`.
3. The actual network call goes through the native Rust module `packages/backend/native/src/safe_fetch.rs` (an SSRF-protection fetch wrapper). Headers not present in `allowed_headers` appear to be dropped when the request is constructed, so no `User-Agent` header is sent — and the underlying HTTP client does not set a default one either.
4. GitHub's edge/WAF rejects any request without a `User-Agent` header with `403 Request forbidden by administrative rules`, which AFFiNE then surfaces as `INVALID_OAUTH_CALLBACK_CODE` / 400 to the client.

This affects every self-hosted deployment doing GitHub OAuth login — it is not specific to any particular domain, proxy, or network setup.

## Suggested Fix
Add `user-agent` to the `allowedHeaders` list in `fetchOptions()` (or set it unconditionally inside `safeFetch`/`safe_fetch.rs` for all outbound requests), and always send a static, descriptive User-Agent string, e.g.:

```ts
headers: { ...init?.headers, Accept: 'application/json', 'User-Agent': 'AFFiNE-Server' }
```

and correspondingly allow it:

```ts
allowedHeaders: ['authorization', 'content-type', 'accept', 'user-agent'],
```

Since GitHub's API requires `User-Agent` on all requests (https://docs.github.com/en/rest/overview/resources-in-the-rest-api#user-agent-required), this should be treated as a baseline requirement for any provider calling GitHub, not an opt-in per provider.

## 📎 References
- GitHub REST API docs — User-Agent required: https://docs.github.com/en/rest/overview/resources-in-the-rest-api#user-agent-required
- `packages/backend/server/src/plugins/oauth/providers/def.ts`
- `packages/backend/server/src/plugins/oauth/providers/github.ts`
- `packages/backend/native/src/safe_fetch.rs`

BetterAndBetterII:
## Description

Self-hosted GitHub OAuth login fails at token exchange because the outbound request to `https://github.com/login/oauth/access_token` has no `User-Agent` header. GitHub then returns 403 ("Request forbidden by administrative rules"), which is surfaced as `INVALID_OAUTH_CALLBACK_CODE`.

OAuth `safeFetch` only forwarded `authorization`, `content-type`, and `accept`, so even a User-Agent on the request would be stripped. This change:

- allows `user-agent` in OAuth `fetchOptions()`
- always sends `User-Agent: AFFiNE-Server` from `fetchJson()` (covers token exchange and `api.github.com` user/email fetches)

Fixes #15521

## Checklist

- [x] The PR targets the `canary` branch and its title follows Conventional Commits

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OAuth request compatibility by including a standard `User-Agent` header.
  * Ensured the header is permitted consistently during OAuth token exchanges.

* **Tests**
  * Added coverage to verify case-insensitive handling of the `User-Agent` header in GitHub OAuth requests.

### DIFF
--- a/packages/backend/server/src/__tests__/oauth/github-user-agent.spec.ts
+++ b/packages/backend/server/src/__tests__/oauth/github-user-agent.spec.ts
@@ -1,0 +1,96 @@
+import serverNativeModule from '@affine/server-native';
+import { TestFn } from 'ava';
+import ava from 'ava';
+import Sinon from 'sinon';
+
+type NativeSafeFetchRequest = {
+  url: string;
+  headers?: Record<string, string>;
+  allowedHeaders?: string[];
+};
+
+const test = ava.serial as TestFn<{
+  requests: NativeSafeFetchRequest[];
+}>;
+
+function headerValue(
+  headers: Record<string, string> | undefined,
+  name: string
+) {
+  const target = name.toLowerCase();
+  for (const [key, value] of Object.entries(headers ?? {})) {
+    if (key.toLowerCase() === target) {
+      return value;
+    }
+  }
+  return undefined;
+}
+
+test.before(t => {
+  const requests: NativeSafeFetchRequest[] = [];
+
+  Sinon.stub(serverNativeModule, 'safeFetch').callsFake(async request => {
+    const nativeRequest = request as NativeSafeFetchRequest;
+    requests.push(nativeRequest);
+    return {
+      status: 200,
+      finalUrl: nativeRequest.url,
+      headers: { 'content-type': 'application/json' },
+      body: Buffer.from(
+        JSON.stringify({
+          access_token: 'github-access-token',
+          scope: 'read:user user:email',
+          token_type: 'bearer',
+        })
+      ),
+    };
+  });
+
+  t.context.requests = requests;
+});
+
+test.after.always(() => {
+  Sinon.restore();
+});
+
+test('github oauth token exchange should send a User-Agent header', async t => {
+  const { OAuthProviderName } = await import('../../plugins/oauth/config');
+  const { OAuthProvider } = await import('../../plugins/oauth/providers/def');
+
+  class Probe extends OAuthProvider {
+    override provider = OAuthProviderName.GitHub;
+
+    getAuthUrl() {
+      return '';
+    }
+
+    async getToken() {
+      return { accessToken: 'token' };
+    }
+
+    async getUser() {
+      return { id: 'id', email: 'user@example.com' };
+    }
+
+    exchange() {
+      return this.postFormJson(
+        'https://github.com/login/oauth/access_token',
+        'code=oauth-code'
+      );
+    }
+  }
+
+  const { requests } = t.context;
+  requests.length = 0;
+
+  await new Probe().exchange();
+
+  t.is(requests.length, 1);
+  t.is(requests[0].url, 'https://github.com/login/oauth/access_token');
+  t.truthy(headerValue(requests[0].headers, 'user-agent'));
+  t.true(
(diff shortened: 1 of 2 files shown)


darkskygit (dismissed):

darkskygit:
Thank you for your PR. Please sign the CLA so we can proceed with the merge.

BetterAndBetterII:
CLA is signed. Thanks for the reminder and the review.

darkskygit:
lint error should fix
```
Run yarn lint:ox
Error: packages/backend/server/src/__tests__/oauth/github-user-agent.spec.ts:2:24: Module 'ava' is imported more than once in this file

Found 0 warnings and 1 error.
```

darkskygit (dismissed):

darkskygit (approved):

PR #12269, opened 2025-05-14
feat(core): add pinned collections to all docs

EYHN:
## Summary by CodeRabbit

- **New Features**
  - Introduced the ability to pin and manage collections, allowing users to quickly access and switch between saved filter sets.
  - Added a UI for managing pinned collections, including adding, removing, and reordering pinned collections.
  - Enabled saving temporary filters as new collections directly from the workspace page.
  - Added support for a "back" button in the filter addition menu for easier navigation.

- **Enhancements**
  - Improved filter management by separating saved collection filters from temporary filters, providing clearer control and UI feedback.
  - Updated filter and collection UI components for better styling and responsiveness.
  - Added support for custom class names in the Filters component for easier styling.

- **Localization**
  - Added a new English localization string for "All" in the pinned collections menu.

- **Style**
  - Introduced new and updated CSS styles for pinned collections, filter areas, and responsive layouts.

- **Chores**
  - Integrated new services and database schema to support pinned collections functionality.

PR #10726, opened 2025-03-10
feat(core): intilize integration module and basic readwise impl

CatsJuice:
close AF-2257, AF-2261, AF-2260, AF-2259

### Feat

- New `Integration` Module
- Basic Readwise integration
  - connect
  - import
  - disconnect
- Common Integration UI
- Common Integration Writer  (Transform markdown to AFFiNE Doc)

### Not Implemented

>  will be implemented in down-stack
- delete docs when disconnect
- readwise settiing UI
- integration property rendering

PR #9566, opened 2025-01-07
feat(core): template-doc settings for user-worksapce scope

CatsJuice:

PR #15528, opened 2026-08-26
fix(server): index & gc queue & llm compatibility

### LINKED ISSUE #15523: [Bug]: BYOK OpenAI-compatible endpoint fails when response contains tool_calls: null
AFFiNE 0.27.4's BYOK OpenAI-compatible provider rejects an otherwise valid Chat Completions response when choices[].message.tool_calls is explicitly null.

The BYOK connection test returns:

invalid_response

The endpoint itself is reachable and returns HTTP 200 with a valid non-streaming Chat Completions response.

I traced the failure to llm_adapter 0.2.20, which is bundled with AFFiNE 0.27.4. Its message parser treats the presence of the tool_calls property as a value to deserialize, even when that value is JSON null.

Removing only "tool_calls": null from the response causes the exact same AFFiNE BYOK probe to change from invalid_response to verified.

Environment
AFFiNE: 0.27.4
Deployment: self-hosted / Docker
Provider: OpenAI-compatible BYOK endpoint
API dialect: Chat Completions
Backend: llama-swap → SGLang
AFFiNE dependency: llm_adapter 0.2.20

The BYOK policy is configured to permit the endpoint:

enabled: true
customEndpointMode: enabled
privateEndpointSupported: true
allowedProviders includes: openai

This is therefore not a private-endpoint or BYOK-policy rejection.

Reproduction

Configure a BYOK OpenAI-compatible provider and return a response containing an explicit null tool_calls.

A response reproducing the issue:

{
  "id": "30d8838a0f3540d1934f7dd43b22de74",
  "object": "chat.completion",
  "created": 1787639897,
  "model": "qwen3.8-27b-nvfp4-inferact",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "OK",
        "reasoning_content": null,
        "tool_calls": null
      },
      "logprobs": null,
      "finish_reason": "stop",
      "matched_stop": 248046
    }
  ],
  "usage": {
    "prompt_tokens": 16,
    "total_tokens": 18,
    "completion_tokens": 2,
    "prompt_tokens_details": null,
    "reasoning_tokens": 0
  },
  "metadata": {
    "weight_version": "default"
  }
}

The server returns:

HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: 459

The required OpenAI-style response fields are all present:

id: string
model: string
choices: non-empty array
choices[0].message: object
choices[0].message.content: "OK"
finish_reason: "stop"

AFFiNE nevertheless reports:

{
  "connection": {
    "kind": "failed",
    "errorKind": "invalid_response"
  }
}
Root cause

In llm_adapter 0.2.20, the message parsing path effectively does:

let raw_tool_calls = message.get("tool_calls").cloned();

if let Some(raw_tool_calls) = raw_tool_calls {
    parse_tool_calls(&raw_tool_calls)?;
}

When the response contains:

"tool_calls": null

message.get("tool_calls") returns:

Some(Value::Null)

rather than None.

The parser therefore calls parse_tool_calls(Value::Null).

That path attempts to deserialize the value into:

Vec<OpenaiToolCall>

and Serde cannot deserialize JSON null into a non-optional Vec, causing a protocol/response parsing failure.

That eventually surfaces through the AFFiNE BYOK probe as:

invalid_response
Proof that tool_calls: null is the failing field

I placed a temporary response-normalizing proxy between AFFiNE and the OpenAI-compatible endpoint.

With the original response:

"tool_calls": null

AFFiNE reports:

invalid_response

The proxy then removed only that field when its value was null:

{
  "role": "assistant",
  "content": "OK",
  "reasoning_content": null
}

No other fields were modified.

With that single transformation, the same AFFiNE BYOK tests succeeded:

connection: verified
chat model check: verified

I separately tested removing the other null-valued fields such as:

reasoning_content
logprobs
prompt_tokens_details

and those were not responsible for the failure.

Expected behavior

An explicit:

"tool_calls": null

should be treated equivalently to an omitted tool_calls property / no tool calls.

The response should successfully deserialize and the BYOK probe should be marked verified.

Actual behavior

tool_calls: null is passed to the tool-call deserializer as though it contained an array, causing the response to be classified as invalid.

The UI ultimately exposes only:

invalid_response

which makes this particularly difficult to diagnose.

Suggested fix

Handle JSON null as no tool calls before invoking parse_tool_calls.

For example:

match message.get("tool_calls") {
    None | Some(serde_json::Value::Null) => {
        // no tool calls
    }
    Some(raw_tool_calls) => {
        parse_tool_calls(raw_tool_calls)?;
    }
}

Or equivalently make the deserialization path explicitly tolerate null.

A regression test would be useful for:

{
  "role": "assistant",
  "content": "OK",
  "tool_calls": null
}

and should confirm that it parses identically to:

{
  "role": "assistant",
  "content": "OK"
}
Workaround

A response-normalizing proxy that removes message.tool_calls only when its value is JSON null makes AFFiNE 0.27.4 work correctly with the same endpoint.

That workaround is proven, but ideally shouldn't be necessary for an OpenAI-compatible BYOK endpoint.

Additional note

It would also be helpful if the BYOK probe surfaced the underlying protocol/deserialization error instead of collapsing this class of failures into the generic:

invalid_response

That would make compatibility problems with local/OpenAI-compatible providers much easier to diagnose.

### LINKED ISSUE #15526: [Bug]: doc gc deletes AFFiNE's own internal docs (db$…$docProperties/folders, userdata$…$settings) — they are never workspace pages
> **Disclosure:** this bug was found and this report was written by **Claude Code** (Anthropic's CLI coding agent), while diagnosing a real production failure on my self-hosted instance. I am submitting it after reviewing it. I have deliberately **left the "not generated by AI" checkbox unchecked** rather than tick it inaccurately. Every claim below is backed by rows from my own database and by the 0.27.4 source; please judge it on that evidence.

### What happened?

The `doc gc` / document-cleanup feature deletes AFFiNE's **own internal workspace docs**, because they are never listed as pages in the workspace root doc and are therefore treated as orphans.

On **2026-08-24 at 04:00:00–04:00:02 UTC**, the first-ever `document_cleanup_execute` sweep on my instance deleted **19 docs across all 5 workspaces**. 17 of the 19 were internal docs:

- `db$<ws>$docProperties`
- `db$<ws>$folders`
- `db$<ws>$explorerIcon`
- `db$<ws>$docCustomPropertyInfo`
- `userdata$<user>$<ws>$settings`
- `userdata$<user>$<ws>$favorite`

(The other 2 were genuinely orphaned page docs — the feature's legitimate target.)

**User-visible consequences:**

1. **Sidebar folder trees were lost** in 4 workspaces (`db$<ws>$folders` deleted).
2. **Journal template setting was lost** (`userdata$…$settings` holds `templateDoc.value.journalTemplateId`).
3. Doc properties were lost in 4 workspaces.
4. `snapshot_histories` had **0 rows** for the deleted docs, so there is **no version history to restore from**.

**Expected:** internal/reserved docs (`db$…`, `userdata$…`) should never be cleanup candidates. They are storage AFFiNE itself creates and depends on, and by design they never appear in the root doc's page list.

### Root cause (0.27.4 source)

In `packages/backend/native/src/runtime/storage_runtime/document_cleanup.rs`:

`load_stored_doc_activity()` scans **every** stored guid, sparing only the workspace root:

```sql
SELECT guid AS doc_id, updated_at FROM snapshots WHERE workspace_id = $1
UNION ALL SELECT guid, created_at FROM updates WHERE workspace_id = $1
UNION ALL SELECT guid, timestamp  FROM snapshot_histories WHERE workspace_id = $1
WHERE doc_id <> $1
```

Liveness is then decided purely by page membership in the root doc:

```rust
fn root_contains(root: CurrentDoc, doc_id: &str) -> RuntimeResult<bool> {
  let projection = affine_doc_loader::project_workspace_root(root.blob, true)...;
  Ok(projection.doc_ids.iter().any(|id| id == doc_id))
}
```

`reconcile_workspace()` marks everything stored-but-not-a-live-page, and `execute_document_cleanup_candidates` deletes it after `gracePeriodDays` (default **30**).

Internal docs are never in `projection.doc_ids`, so they are marked on **every** nightly run. Grepping that file for `userdata`, `db$`, `prefix`, `starts_with`, `is_internal`, `exclude`, `reserved` returns **no matches** — there is no allowlist or prefix filter at all.

### Why it fired when it did

`document_cleanup` (mark_only) began running on my instance **2026-07-25** — consistent with #15282 (`feat(server): impl doc gc`, merged 2026-07-19). 07-25 + 30 days grace = **2026-08-24**, the date of the first execute sweep. So instances that adopted doc gc around then are reaching the deletion boundary now.

### It recurs

`missing_since` only resets when the doc sees activity, and internal docs are re-marked nightly. So **any internal doc that goes 30 days without a write is deleted again.** After I recreated a folder, its clock simply restarted.

### The client does not heal it

After `db$<ws>$folders` was deleted, my browser tab was still connected, so it pushed only incremental ops against a parent that no longer existed server-side. Decoding the resulting snapshot shows the folder's name as a **live but orphaned item with an unresolvable parent**:

```
4522778557396753:0-6  parent=iKaPhGpcOnekUzNW4gyfo  ...  $$DELETED=true
2985492552941628:0    parent=-  del=false  ["Network Redesign_"]   <- orphan, never materialises
```

A page reload did not re-create the missing parent node. Net effect: the folder still renders from browser-local state, but the server copy is permanently unusable — so this is silent, and users will only notice on a new browser or device.

### Suggested fix

Exclude reserved doc-id prefixes (`db$`, `userdata$`) from the candidate scan in `load_stored_doc_activity()` / `reconcile_workspace()`, or make the live set include workspace-scoped internal docs rather than pages only.

### Not fixed on canary

I diffed `document_cleanup.rs` between `v0.27.4` and `canary` (2026-08-25): the changes are refactoring (function signatures, `merge_current_doc` arity, `doc_blob_ref_projections` added to the delete list). The same grep for internal-doc filtering still returns nothing, so `v0.27.5-beta.0` does not address this.

### No way to disable or tune it

In `packages/backend/server/src/core/storage/blob-job.ts`, `gracePeriodDays = 30` is a hardcoded handler default and the cron enqueues with an empty payload:

```ts
@Cron(CronExpression.EVERY_DAY_AT_4AM)
… queue.add('backendRuntime.executeDocumentCleanupCandidates', {})

async executeDocumentCleanupCandidates({ workspaceId, gracePeriodDays = 30, limit = 100 })
```

There is no config binding, and nothing in `app_configs` for it — so self-hosters cannot opt out while waiting for a fix.

### Environment

- **Distribution version:** Linux (self-hosted, Docker, `ghcr.io/toeverything/affine:0.27.4`)
- **Self-hosting:** Yes
- **Self-hosting version:** 0.27.4
- **Database:** external PostgreSQL

### Relevant log output

`storage_reconciliation_runs` — the first `document_cleanup_execute` rows ever recorded on this instance:

```
started_at                 kind                      workspace_id  doc_id
2026-08-24T04:00:02.514Z   document_cleanup_execute   fe07bcd8…     userdata$ae5e65c2…$fe07bcd8…$settings
2026-08-24T04:00:02.475Z   document_cleanup_execute   fe07bcd8…     db$fe07bcd8…$folders
2026-08-24T04:00:02.348Z   document_cleanup_execute   2cf4d0c0…     userdata$ae5e65c2…$2cf4d0c0…$favorite
2026-08-24T04:00:02.268Z   document_cleanup_execute   2cf4d0c0…     db$2cf4d0c0…$folders
2026-08-24T04:00:02.169Z   document_cleanup_execute   2cf4d0c0…     db$2cf4d0c0…$explorerIcon
2026-08-24T04:00:02.082Z   document_cleanup_execute   2cf4d0c0…     db$2cf4d0c0…$docProperties
2026-08-24T04:00:01.715Z   document_cleanup_execute   edecd35c…     userdata$ae5e65c2…$edecd35c…$settings
2026-08-24T04:00:01.391Z   document_cleanup_execute   edecd35c…     db$edecd35c…$docProperties
2026-08-24T04:00:00.702Z   document_cleanup_execute   6bc3499a…     db$6bc3499a…$docCustomPropertyInfo
```

Row counts for the deleted settings doc — note `snapshot_histories: 0`, i.e. unrecoverable:

```json
{"snapshots":1,"doc_grants":1,"updates":0,"snapshot_histories":0,"workspace_pages":0,
 "doc_access_policies":0,"doc_blob_refs":0,"comments":0,"replies":0}
```

Aggregate confirming this was the first execute sweep, after a month of mark-only runs:

```
kind                      mode        n     first                     last
document_cleanup          mark_only   155   2026-07-25T02:00:00.247Z  2026-08-24T02:00:06.499Z
document_cleanup_execute  execute      19   2026-08-24T04:00:00.201Z  2026-08-24T04:00:02.514Z
```

darkskygit:
fix #15523
fix #15526

#### PR Dependency Tree

* **PR #15528** 👈

This tree was auto-generated by [Charcoal](https://github.com/danerwilliams/charcoal)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved search generation cleanup, provider error reporting, and reconciliation reliability.
  - Retired search resources are cleaned up safely, including after credential changes.
  - Prompt size checks now ignore tool parameters and provide clearer errors.
  - Reserved documents are protected from accidental cleanup, and malformed identifiers are rejected.

- **Configuration**
  - Managed Copilot profiles require explicit, non-duplicated model assignments.
  - Improved managed provider profile migration.

- **Performance & Reliability**
  - Reduced unnecessary search-history cleanup and adjusted consistency-check intervals.
  - Failed reconciliation jobs stop after one attempt and are removed automatically.

- **Data Updates**
  - Updated legacy AI session prompt names to current labels.
  - Improved cloud load-balancer health-check configuration.


MdeRhoter:
Thank you for the very quick turnaround on this — reported yesterday and fixed today is genuinely impressive, and much appreciated.

The fix looks right to me. I especially like that `execute_one` doesn't just skip reserved docs but actively *recovers* an already-marked candidate (deleting the candidate row and returning the existing `-1` sentinel rather than the doc). That detail matters a lot for instances like mine that are already mid-flight: I have a `db$<ws>$docProperties` sitting at `status='marked'` with its grace period expiring in late September, and as I read this it will now be recovered instead of executed. Thanks for handling the existing-damage case rather than only the fresh path.

One small, entirely optional suggestion for hardening.

I checked the allowlist in `reserved_doc.rs` against `AFFiNE_WORKSPACE_DB_SCHEMA` / `AFFiNE_WORKSPACE_USERDATA_DB_SCHEMA` in `packages/frontend/core/src/modules/db/schema/schema.ts`, and it is currently an **exact** match — all 5 workspace tables and all 3 userdata tables are covered, nothing missing. So this is not a live gap, only a drift risk.

The thing is that `classify()` fails *open*: an unrecognised table name returns `None`, which means the doc is treated as a normal orphan and deleted after the grace period. Since nothing links the TypeScript schema to the Rust array — no shared constant, no codegen, no parity test, and it crosses both a language and a package boundary — a future table added to `schema.ts` without a matching entry here would reintroduce this bug for that table, silently, 30 days later.

For what it's worth, the historical rate is low: `settings` (#9566), `docIntegrationRef` (#10726), `pinnedCollections` (#12269) and `explorerIcon` (#13595) were added across 2025, and the schema has been unchanged since. So this is infrequent — but it is silent when it does happen, and the failure mode is data loss rather than an error.

Worth noting which docs are actually at risk, since it shapes how easy this is to miss: `missing_since` resets whenever the doc sees activity, so a table that is written regularly effectively protects itself. The exposure is concentrated in **write-once-then-idle** tables — which is exactly how this bit me, since `userdata$…$settings` held my journal template pointer, was written once, and then sat untouched past the grace period.

Two possible options, either cheap:

1. A parity test asserting the Rust arrays cover every key of both schema objects, so an omission fails CI instead of deleting data.
2. Or invert the rule to be prefix-based — treat any *well-formed* `db$<ws>$<table>` / `userdata$<owner>$<ws>$<table>` as reserved regardless of table name — which fails safe. The strict parsing and workspace-id matching you've already implemented would still reject the malformed and `__local__` cases you have tests for.

Not a blocker at all, and I'm happy either way given the current list is complete. Thanks again for the fast fix.

<sub>Disclosure, as in the original issue: the investigation and this comment were produced with Claude Code, reviewed by me before posting.</sub>

PR #15541, opened 2026-08-28
feat(ios): restore share extension enrichment

keepClamDown:
## Description

Restores the iOS share extension enrichment path on top of the latest canary branch.

- Capture Safari page title, URL, selected text, meta description, main page text, X post text, and YouTube metadata/transcript/thumbnail in the share extension preprocessing script.
- Preserve native link previews in the pending share manifest so AFFiNE mobile import can use the captured YouTube/X/page data without falling back to generic "Shared content".
- Extract text from shared PDF files and bounded remote .pdf URLs, with graceful URL fallback on unsupported/slow PDFs.
- Add regression coverage for Safari captured transcript/content, PDF file extraction, remote PDF timeout fallback, pending preview import, and share extension activation rules.

## Checklist

- [ ] I have signed the [AFFiNE Contributor License Agreement](https://cla-assistant.io/toeverything/AFFiNE) — required before merge; the `license/cla` check must be green ([how it works](https://github.com/toeverything/AFFiNE/blob/canary/docs/BUILDING.md#sign-the-cla-first))
- [x] The PR targets the `canary` branch and its title follows [Conventional Commits](https://www.conventionalcommits.org/)
- [ ] `yarn lint` and `yarn typecheck` pass locally

Local verification performed:

- `xcodebuild -workspace packages/frontend/apps/ios/App/App.xcworkspace -scheme AFFiNETests -configuration Debug -destination 'platform=iOS Simulator,id=8064D133-9BCC-4484-A936-36689730CF4C' -derivedDataPath /private/tmp/affine-dd-share-restore ONLY_ACTIVE_ARCH=YES -disableAutomaticPackageResolution -onlyUsePackageVersionsFromResolvedFile -only-testing:AFFiNETests/ShareInboxSafetyTests test` — 12 tests passed.
- `yarn vitest run packages/frontend/core/src/mobile/components/share-import-controller/share-link-preview.spec.tsx` — 41 tests passed.
- `git diff --check` — passed.
- `xcodebuild -workspace packages/frontend/apps/ios/App/App.xcworkspace -scheme App -configuration Debug -destination 'platform=iOS,id=00008101-000175E43442001E' -derivedDataPath /private/tmp/affine-dd-share-restore ONLY_ACTIVE_ARCH=YES -disableAutomaticPackageResolution -onlyUsePackageVersionsFromResolvedFile build` — build succeeded and the app was installed to device `00008101-000175E43442001E`.

## Summary by CodeRabbit

* **New Features**
  * Enhanced iOS sharing with richer Safari page captures, including titles, content, metadata, and transcripts.
  * Added PDF sharing support for local files and remote PDF links, including text extraction.
  * Preserved native link previews throughout the import flow for faster, more accurate previews.
  * Expanded share support to recognize PDF files.

* **Bug Fixes**
  * Added size and timeout safeguards for PDF and transcript processing.
  * Unsupported or unavailable remote PDFs now fall back to sharing a plain link.


packages/frontend/apps/ios/App/Shared/ShareInbox/SharePayloadBuilder.swift
```diff
@@ -77,6 +99,19 @@ enum SharePayloadBuilder {
...
+       let pdf = try? await loadRemotePDFText(
+         from: remoteURL,
+         session: remotePDFSession,
+         timeout: remotePDFTimeout
+       )
+    {
+      title = title == "Shared" ? pdf.title : title
+      text = pdf.text
+      url = nil
+    }
```
darkskygit, line 114: embed pdf as a attachment/preview block directly into doc, dont put any preview text extract from pdf into doc
prefer preview pdf directly based on ios's native preview control, but not extract the non-formatted text in pdf files directly
prefer only render the first page of pdf like a image and put in the share extension view, not required a interactive preview interface

packages/frontend/apps/ios/App/ShareExtension/SafariPageCapture.js
```diff
@@ -1,13 +1,267 @@
...
+  return fetchTranscript(response).then(function (transcript) {
+    return {
+      title: normalizeText(details.title) || document.title || '',
+      description: description,
+      content: description || pageContent(),
+      transcript: transcript,
+      thumbnailURL: thumbnail,
+    };
+  });
+}
```
darkskygit, line 232: should not parse youtube transcript / x / etc. in client, only use transcript data in link preview api

darkskygit (changes requested):

donqu1xotevincent (changes requested):
I agree with the direction in Darksky's review and recommend restructuring this PR before merge. The remaining issues are architectural and data-integrity/security concerns, not isolated guards.

### 1. Preserve PDFs as attachments instead of flattening them into text

`SharePayloadBuilder` currently extracts a local or remote PDF into plain text and clears the original `file`/`url`. This permanently loses the PDF binary, layout, and attachment identity before the manifest reaches the importer. The downstream Swift/TypeScript contracts also only model `url | text | image`, so the document cannot receive a PDF attachment/preview block.

Please:

- persist the original `application/pdf` data as a `ShareInboxAttachment`;
- extend `ShareInboxContent`, `PendingShareItem`, `ShareImportInput`, attachment resolution, and the importer with a PDF/file path;
- create a PDF attachment/preview block in the document;
- render only a first-page static thumbnail in the share extension;
- do not insert unformatted extracted PDF text into the document; and
- rewrite the current PDF tests, which presently assert the rejected `.text` behavior.

### 2. Keep provider-specific enrichment in the link-preview API

`SafariPageCapture.js` now parses YouTube's private page state, caption tracks, VTT/timed-text, and X-specific content. Swift then reconstructs a `ShareLinkPreview`, and `previewForImport()` returns `item.preview` before `SharePreviewRouteOwner.selectWorkspace()` or `load()` runs. In the auto-import/retry path, this can bypass the selected official or self-hosted link-preview endpoint entirely.

Please keep Safari preprocessing limited to generic URL/title/selected-text/page capture. YouTube/X metadata and transcript should come from the selected `/api/worker/link-preview` endpoint (`include: ['transcript']`) after workspace routing. A native preview may be used for temporary presentation, but it should not be persisted as authoritative enrichment or short-circuit endpoint selection. The tests should mock a valid API transcript and prove that the selected endpoint is called.

### 3. Remove the client-side remote PDF fetch

The remote `.pdf` path accepts any credential-free HTTP(S) host and uses `URLSession.shared`, which follows redirects. It does not reject loopback/private/link-local destinations or validate the final response URL. The current MIME check is also ineffective on this path because the original URL already has a `.pdf` extension.

Following the PDF attachment design above, the safest fix is to remove remote PDF download/extraction from the share extension and preserve the URL for the routed preview/import layer. If this fetch remains, it needs an ephemeral cookie-free/cache-free session, public-address validation before the request and after every redirect, strict final Content-Type/PDF-signature validation, and streamed size limits.

### 4. Define the multi-attachment behavior explicitly

The activation rule allows multiple supported attachments, but the builder only counts images. Each PDF overwrites shared state and clears `file`; two PDFs can produce the first title with the last body, while PDF + image behavior changes with provider order and may still report success.

Please either reject more than one binary attachment with a clear error, or implement true multi-attachment accumulation across the manifest, bridge, and importer.

### Tests and merge gate

Several new tests currently lock in the behavior Darksky rejected: PDF-to-text conversion and client-originated transcript/native-preview import. Those tests need to be rewritten around the attachment contract and API-owned enrichment.

CodeQL is also still failing on the client transcript parser. I do not think the reported `stripTags` path alone demonstrates an exploitable XSS because the entity decoder uses a detached `<textarea>` and downstream rendering is text-based. However, removing the client transcript parser as requested also removes this alert surface and is preferable to adding another regex sanitizer.

**Recommendation: request changes.** I would re-review after the PDF attachment/import contract is implemented, provider-specific enrichment is routed exclusively through the link-preview API, remote PDF fetching is removed or fully constrained, and mixed/multiple attachments have deterministic behavior with regression coverage.

PR #15547, opened 2026-08-29
feat(ios): add stable Share Extension imports

keepClamDown:
## Summary

The changes below are prepared locally and have not been committed or pushed.

- Save original URLs, user titles, selected text, images and PDFs through a versioned, file-backed Share Inbox. Stable document/attempt IDs, serialized imports and committed receipts make retries preserve user edits and recover native completion.
- Keep rich previews transient. Share previews request transcript excerpts, clip optional display fields and bound response/image downloads. Saving does not wait for preview requests or persist provider metadata, transcript snapshots or dedicated preview blobs.
- Restore the minimal App Group workspace-mode snapshot and `official`/`deferred` route. Official routes remain fixed for the pending share; deferred self-hosted previews use the selected server without a provider fallback.
- Let bookmark and YouTube details read the current workspace's Link Preview API by URL. Full transcripts, chapters and timestamps load on expansion without document writes. Reuse editor styles and the existing portal for canvas details; isolate endpoint/base/transcript caches and keep transcript results out of persistent cache.

## Validation

- Focused frontend suites: 128 tests passed; the 43 preview/details cases also passed after the final portal change.
- `ShareInboxSafetyTests`: 57 tests passed on iOS 26.5 simulator.
- `yarn typecheck` and full `yarn lint` passed.
- iOS web bundle, Capacitor asset copy, and signed simulator App build with embedded ShareExtension passed.
- Browser component checks: desktop, narrow viewport, light/dark themes, full transcript scrolling, canvas portal positioning and Escape dismissal.
- Simulator Safari share: rich preview, cancellation, native-to-app handoff, workspace selection and successful save. The pending manifest contained original content and routing only, and was removed after import.

Physical-device and network-capture matrices remain manual. The simulator YouTube iframe displayed player error 153; this is separate from the successful preview/import checks and video playback is not claimed as verified.

## Summary by CodeRabbit

* **New Features**
  * iOS sharing now supports PDF files alongside images, with previews and attachment validation.
  * Share imports offer improved workspace selection, offline handling, retry recovery, and clearer error messages.
  * Link previews can display expandable metadata and transcript details for bookmarks and YouTube embeds.
  * Share inbox entries now identify unsupported versions and provide safer attachment resolution.

* **Bug Fixes**
  * Improved cancellation, cleanup, and stale-request handling during sharing and importing.
  * Enhanced link-preview validation and protection against oversized or invalid responses.


packages/backend/server/src/__tests__/e2e/config/resolver.spec.ts
```diff
@@ -33,3 +34,31 @@ e2e('should enable local workspace feature by default', async t => {
...
+    const disabled = await app.gql({ query: serverConfigQuery });
+    t.false(
+      disabled.serverConfig.features.includes(
+        ServerFeature.SharePreviewBlobRefs
+      )
+    );
+
+    server.onConfigChangedBroadcast({
+      updates: { flags: { sharePreviewBlobRefs: true } },
+    });
```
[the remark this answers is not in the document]
keepClamDown: Accepted and fixed in `4dca541f79`. Both rollout state changes and assertions now run inside `try`, with `finally` always restoring `sharePreviewBlobRefs` to `false`. Verified by the focused config E2E suite: 4/4 passed.

packages/frontend/apps/ios/App/AppTests/ShareInboxSafetyTests.swift
```diff
@@ -75,15 +1203,15 @@ final class ShareInboxSafetyTests: XCTestCase {
...
+    let supportedPDFPayload: [String: Any] = [
       "extensionItems": [[
         "attachments": [[
           "registeredTypeIdentifiers": ["com.adobe.pdf", "public.movie"]
         ]]
       ]]
     ]
     XCTAssertTrue(predicate.evaluate(with: youtubePayload))
-    XCTAssertFalse(predicate.evaluate(with: unsupportedPayload))
+    XCTAssertTrue(predicate.evaluate(with: supportedPDFPayload))
```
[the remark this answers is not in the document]
keepClamDown: Accepted and fixed in `4dca541f79`. The activation test now retains the supported PDF case and adds a `public.movie`/`public.audio`-only payload that must evaluate to false. `ShareInboxSafetyTests` passes in Xcode.

packages/frontend/core/src/mobile/components/share-import-controller/index.tsx
```diff
@@ -335,20 +464,37 @@ export const ShareImportController = ({
...
     }
     return () => {
       active = false;
+      if (objectUrl) URL.revokeObjectURL(objectUrl);
+      if (attachmentRef.current?.itemId === item?.id) {
+        attachmentRef.current = undefined;
+      }
     };
-  }, [item?.content.kind, item?.id, provider]);
+  }, [item?.content.kind, item?.id, provider, updateLinkPreview]);
```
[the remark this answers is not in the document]
keepClamDown: Accepted and fixed in `4dca541f79`. Link-preview reset is now a separate effect, so workspace selection no longer tears down or re-resolves the image attachment. The regression switches workspace A to B and verifies one native `resolveAttachment` call, a stable object URL, and no premature revoke. Focused frontend tests: 213/213 passed.

packages/frontend/core/src/modules/import-clipper/services/import.ts
```diff
@@ -311,13 +712,130 @@ export class ImportClipperService extends Service {
...
+        props: { type: 'quote', text: selectedText },
+      });
+    }
+    if (input.content.url) {
+      nodes.push({
+        id: shareImportBlockIds(input.importAttemptId).sourceLink,
+        flavour: 'affine:bookmark',
+        props: {
+          url: input.content.url,
+          title: input.title.trim() || new URL(input.content.url).hostname,
```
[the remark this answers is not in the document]
keepClamDown: Accepted and fixed in `4dca541f79`. Added the shared `shareUrlTitle` guard and reused it for URL and non-URL block plans. An integration regression verifies malformed source URLs fall back to the raw string and still return the typed imported result. Focused frontend tests: 213/213 passed.

docs/superpowers/specs/2026-08-30-ios-share-rich-preview-design.md
```diff
@@ -0,0 +1,152 @@
...
+`${prefix}-transcript-heading`, `${prefix}-transcript-chapter-${originalIndex}`, and
+`${prefix}-transcript-segment-${originalIndex}`. Segments retain response order after empty
+entries are filtered, so IDs use the original response index. Chapters sort by
+`(startSeconds, originalIndex)` and are inserted before the first retained segment whose finite
+`startSeconds >= chapter.startSeconds`; unmatched chapters follow the final segment.
+
+Deduplication normalizes text by splitting Unicode whitespace, joining with one ASCII space,
+trimming, and lowercasing. A segment equal to the normalized selected text or description is
+removed. If the normalized concatenation of all retained segment text equals either value, the
+whole transcript is omitted. Timestamps use floor-to-whole-seconds formatting. Chapter titles,
```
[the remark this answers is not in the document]
keepClamDown: Accepted and fixed in `4dca541f79`. The implementation already joined retained segments with one ASCII space; the design contract now states that delimiter explicitly, and a `foo` + `bar` versus selected `foobar` regression proves the transcript remains.

packages/frontend/apps/ios/App/AppTests/ShareInboxSafetyTests.swift
```diff
@@ -10,7 +12,2161 @@ private final class SharePreviewURLProtocol: URLProtocol {
...
+    let requested = expectation(description: "preview requested")
+    SharePreviewURLProtocol.onStart = { protocolInstance, request in
+      XCTAssertEqual(request.url, ShareInboxConstants.officialLinkPreviewURL)
+      XCTAssertEqual(request.httpMethod, "POST")
+      XCTAssertEqual(request.value(forHTTPHeaderField: "Content-Type"), "application/json")
+      XCTAssertEqual(request.value(forHTTPHeaderField: "User-Agent"), "AFFiNE/0.27.0")
+      XCTAssertEqual(request.value(forHTTPHeaderField: "x-affine-version"), "0.27.0")
+      let body = try? XCTUnwrap(sharePreviewRequestBody(request)).flatMap {
+        try JSONSerialization.jsonObject(with: $0) as? [String: Any]
+      }
```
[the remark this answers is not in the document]
keepClamDown: The original expression does compile, as confirmed by the Xcode suite, because the optional `try?` result selects `Optional.flatMap`; however, the form is needlessly ambiguous. Fixed in `4dca541f79` by binding the unwrapped `Data` first and parsing it in a separate step. `ShareInboxSafetyTests` passes.

packages/frontend/core/src/mobile/components/share-import-controller/preview-route-owner.ts
```diff
@@ -1,17 +1,339 @@
...
+  duration: 7 * 24 * 60 * 60,
+} as const;
+
+const invalidField = Symbol('invalid-share-preview-field');
+const utf8Encoder = new TextEncoder();
+const swiftWhitespace = new Set([
+  0x0009, 0x000a, 0x000b, 0x000c, 0x000d, 0x0020, 0x0085, 0x00a0, 0x1680,
+  0x2000, 0x2001, 0x2002, 0x2003, 0x2004, 0x2005, 0x2006, 0x2007, 0x2008,
+  0x2009, 0x200a, 0x200b, 0x2028, 0x2029, 0x202f, 0x205f, 0x3000,
+]);
```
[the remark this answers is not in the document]
keepClamDown: Accepted and fixed in `4dca541f79`. Removed U+200B from the mirrored Swift whitespace set and inverted the existing cross-runtime regression so a persisted title beginning with ZERO WIDTH SPACE remains valid, matching `CharacterSet.whitespacesAndNewlines`. Focused frontend tests: 213/213 passed.

docs/ios-share-extension-feasibility.md
```diff
@@ -0,0 +1,328 @@
...
+
+## 执行步骤和提交计划
+
+| 阶段                   | 完成条件                                                                                                                                            | 验证                                                                                                                                            | 建议 commit 信息                                                   | 状态                  |
+| ---------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------ | --------------------- |
+| 0. 隐私与清单基线      | manifest v2 自定义迁移并稳定回填 `importAttemptId`；删除 Extension 网络；Worker 复用全局移动端 Origin；preview 绑定 item/workspace/generation       | Swift/TS/Worker 测试覆盖 v1、未来版本、Capacitor Origin、A→B 未返回即保存、本地/离线；MVP 请求体无 `include`                                    | `refactor(ios-share): make share inbox workspace routed`           | 已完成                |
```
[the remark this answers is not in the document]
keepClamDown: Accepted and fixed in `4dca541f79`. Phase 0 now scopes the network gate to deleting non-official or unapproved Extension requests while explicitly allowing the official preview/image-proxy endpoints, matching the approved enrichment sections.

keepClamDown:
Addressed all 9 CodeRabbit review threads in `4dca541f79` and `500250db9a`: rollout-flag cleanup, activation negative coverage, stable image attachment preview across workspace changes, malformed source-URL fallback, transcript delimiter contract, clearer Swift request-body parsing, Swift-compatible U+200B handling, scoped Extension networking, and a release checklist aligned with the shipped v1/v2-to-v3 migration.

Verification:
- focused frontend Vitest: 213/213
- backend config E2E: 4/4
- Xcode `ShareInboxSafetyTests`: passed
- repository TypeScript typecheck: passed
- oxfmt and `git diff --check`: passed

Each original thread has an inline technical reply and is resolved. Two fresh queries after the latest review cycle report no unresolved CodeRabbit conversations on head `500250db9ad4cde7058fbb2c12c767d2f9908acc`.

docs/ios-share-extension-feasibility.md
```diff
@@ -0,0 +1,328 @@
...
+| ---------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------ | --------------------- |
+| 0. 隐私与清单基线      | manifest v2 自定义迁移并稳定回填 `importAttemptId`；删除非官方或未批准的 Extension 网络请求，仅允许官方 preview/image-proxy 请求；Worker 复用全局移动端 Origin；preview 绑定 item/workspace/generation | Swift/TS/Worker 测试覆盖 v1、未来版本、Capacitor Origin、A→B 未返回即保存、本地/离线；MVP 请求体无 `include`                                    | `refactor(ios-share): make share inbox workspace routed`           | 已完成                |
```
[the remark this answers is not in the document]
keepClamDown: Accepted and fixed in `500250db9a`. The Phase 0 release checklist now names the shipped v1/v2-to-v3 migration, states that only v1 backfills `importAttemptId` while v2 preserves it, and requires v1/v2/v3 plus future-version coverage. This matches the detailed Inbox v3 contract and the implemented migration tests.

packages/frontend/core/src/modules/import-clipper/services/import.ts
```diff
@@ -141,112 +270,376 @@ export class ImportClipperService extends Service {
...
-        if (input.content.kind === 'url' && input.content.url) {
-          const embedOptions = createBlockStdScope(doc.blockSuiteDoc)
-            .get(EmbedOptionProvider)
-            .getEmbedBlockOptions(input.content.url);
-          this.addShareBlocks(
+        if (input.content.kind === 'pdf' && input.attachment) {
+          const std = this.createShareImportBlockStdScope(doc.blockSuiteDoc);
+          try {
+            const fileSizeLimit = std.get(FileSizeLimitProvider);
+            if (input.attachment.size > fileSizeLimit.maxFileSize) {
```
donqu1xotevincent: [P2] Run target admission before persisting the transaction

By the time this workspace-specific limit is checked, the `preparing` receipt has already been persisted and `createDoc({ skipInit: true })` has created the document record/skeleton. A PDF that passes the global 64 MiB limit but exceeds this workspace limit returns `attachment-too-large` here; the later `blobSync.set` failure has the same shape. The UI explicitly lets the user choose another workspace, and a subsequent success there completes/removes the native inbox item, leaving the first workspace with a blank document and a permanent `preparing` receipt. Please perform target admission before receipt/doc creation where possible, and otherwise safely roll back only an untouched record owned by this attempt. Add an A-fails/B-succeeds regression test.

packages/frontend/apps/ios/App/Shared/ShareInbox/SharePayloadBuilder.swift
```diff
@@ -172,86 +321,419 @@ enum SharePayloadBuilder {
...
+        guard let item else {
+          completion(.failure(ShareInboxError.invalidPayload))
+          return
+        }
+        do {
+          let file: SharePayloadFile
+          if let url = item as? URL, url.isFileURL {
+            file = try stageImage(from: url, suggestedName: url.lastPathComponent)
+          } else if let data = item as? Data {
+            file = try stageImage(data: data, suggestedName: suggestedName)
```
donqu1xotevincent: [P2] Bound image input before materializing it

The normal image-provider path still uses `loadItem`, so the provider may first materialize an arbitrarily large `Data` or a fully decoded `UIImage`; the 12 MiB check only happens later in `stageImage(data:)`, and the `UIImage` branch creates another full JPEG buffer. A high-resolution compressed HEIC/PNG can therefore exceed the Share Extension memory budget and be killed before any manifest is written. Please prefer `loadInPlaceFileRepresentation`/`loadFileRepresentation`, check file size and magic inside the coordinated read, then copy in chunks as the PDF path does. Any in-memory fallback should be provably bounded, and the test should exercise this real provider entry point rather than calling `stageImage(from:)` directly.

packages/backend/native/src/runtime/storage_runtime/blob_cleanup.rs
```diff
@@ -829,4 +992,173 @@ mod tests {
       .await
       .unwrap();
   }
+
+  #[tokio::test]
+  #[ignore = "requires DATABASE_URL and a migrated PostgreSQL database"]
```
donqu1xotevincent, line 997: [P2] Put the destructive cleanup scenarios on a CI path

Both new PostgreSQL tests for share-preview reference removal and parser-v1 stale/pending/failed fail-closed behavior are marked `#[ignore]` (the second is below at line 1079). The repository's Rust CI runs `cargo nextest run` without an ignored-test invocation, so these tests never execute even when a database is available. Because these are the end-to-end proofs that object deletion stays fail-closed, please add a required job with a migrated PostgreSQL/object-store fixture that explicitly runs them, or otherwise make them active in that environment.

packages/backend/native/src/runtime/storage_runtime/doc_blob_refs.rs
```diff
@@ -298,8 +299,36 @@ async fn purge_removed_doc_projections(
 }
 
 fn extract_refs(blob: Vec<u8>) -> RuntimeResult<Vec<ExtractedRef>> {
-  doc_loader::get_blob_refs_from_binary(blob)
-    .map_err(|err| RuntimeError::invalid_state(format!("Doc blob refs parse failed: {err}")))
+  let mut refs = doc_loader::get_blob_refs_from_binary(blob.clone())
```
donqu1xotevincent: [P2] Avoid decoding every snapshot twice during the v2 rebuild

`get_blob_refs_from_binary` already performs a full `ReadDoc::from_full_update_v1` parse internally; cloning the snapshot here and parsing it again on line 304 constructs the CRDT tree twice. Bumping `PARSER_VERSION` to 2 forces a workspace-wide reprojection before cleanup, so large deployments pay roughly double the decode CPU and peak allocations on the rollout path. Please extend the loader to collect bookmark refs in the same traversal (or return/reuse the parsed document), and add a representative large-snapshot benchmark/regression.

packages/backend/server/src/core/config/config.ts
```diff
@@ -81,4 +82,8 @@ defineModuleConfig('flags', {
     desc: 'Whether allow guest users to create demo workspaces.',
     default: true,
   },
+  sharePreviewBlobRefs: {
```
donqu1xotevincent, line 85: [P2] Commit the generated config artifacts

This is a public config descriptor, so `yarn affine server genconfig` adds it to both `.docker/selfhost/schema.json` and `packages/frontend/admin/src/config.json`; neither generated change is present on this head. The required `Check Git Status` job is failing for exactly this generation gate. Please run the workflow's full command (`yarn affine init && yarn affine gql build && yarn affine i18n build && yarn affine server genconfig`) and commit the resulting artifacts.

packages/backend/native/src/runtime/storage_runtime/blob_cleanup.rs
```diff
@@ -829,4 +992,173 @@ mod tests {
...
+        .plan_unreferenced_workspace_blobs(fixture.workspace_id.clone(), 0, 100)
+        .await
+        .map_err(|err| anyhow::anyhow!(err.to_string()))?;
+      assert_eq!(
+        (blocked_plan.candidates_marked, blocked_plan.protected_by_metadata),
+        (0, 1)
+      );
+
+      sqlx::query(
+        "UPDATE doc_blob_ref_projections p SET parser_version = $3, status = 'fresh', source_revision = s.updated_at FROM \
```
donqu1xotevincent: [P2] Apply the repository rustfmt result to this new query

The required `Lint Rust` job reports a diff at this exact string: current nightly rustfmt moves `FROM snapshots ...` onto the preceding continuation and rewraps the following predicate. Please run `cargo +nightly fmt --all` and commit the PR-owned change. The job also reports formatting drift in unrelated base files, but this newly added hunk should still be clean.

donqu1xotevincent (changes requested):
Requesting changes for the six inline P2 findings below. The main correctness issue is that an attachment failure in workspace A can leave a blank document with a permanent `preparing` receipt; if the user then succeeds in workspace B, the inbox item is removed and A is never recovered. The destructive blob-cleanup path also lacks active CI coverage and parser-v2 currently doubles full-snapshot decoding. Finally, the current head has two failing required checks: `Check Git Status` is missing generated config artifacts, and `Lint Rust` reports a PR-owned rustfmt diff in addition to baseline formatting drift.

packages/frontend/core/src/modules/import-clipper/services/import.ts
keepClamDown: Fixed in 8e62ab3211. Attachment admission and the initial blob write now complete before a preparing receipt or document is persisted. The share ingress limit is 64 MiB, already stricter than the current 2 GiB BlockStd limit, and is now applied to both image and PDF imports before the transaction starts. Blob-write and size-limit tests assert that no receipt or document is created, and a new workspace-A-fails/workspace-B-succeeds regression verifies A is not stranded.

packages/frontend/apps/ios/App/Shared/ShareInbox/SharePayloadBuilder.swift
keepClamDown: Fixed in 8e62ab3211. Image loading now selects the most specific registered image UTI and uses loadInPlaceFileRepresentation, falling back only to loadFileRepresentation. The staged file is size-checked before magic-byte validation and chunked copy; the unbounded loadItem Data/UIImage path was removed. Added real NSItemProvider regressions for an oversized specific file versus generic data and for a data-only provider synthesized through the file API. ShareInboxSafetyTests pass 79/79.

packages/backend/native/src/runtime/storage_runtime/blob_cleanup.rs
keepClamDown: Fixed in 8e62ab3211. The required Run native tests job now provisions an isolated pgvector PostgreSQL service and explicitly runs cargo test -p affine_server_native --release share_preview_blob_cleanup_ -- --ignored --nocapture. This executes both destructive cleanup scenarios instead of leaving them outside CI. The release test binary was compiled locally; execution is delegated to CI because no local PostgreSQL or Docker daemon is running.

packages/backend/native/src/runtime/storage_runtime/doc_blob_refs.rs
keepClamDown: Fixed in 8e62ab3211. Parser v2 now constructs one ReadDoc from the snapshot and extracts image, attachment, and bookmark refs in a single traversal; the second affine_doc_loader decode and snapshot clone are gone. Added an already-decoded projection regression plus a representative 512-reference large-snapshot regression. All doc_blob_refs tests and clippy pass.

packages/backend/server/src/core/config/config.ts
keepClamDown: Fixed in 8e62ab3211. Ran the same full generation chain as Check Git Status and committed the resulting .docker/selfhost/schema.json and packages/frontend/admin/src/config.json updates. Re-running the complete generation chain after the commit leaves the worktree clean.

packages/backend/native/src/runtime/storage_runtime/blob_cleanup.rs
keepClamDown: Fixed in 8e62ab3211. Applied the exact nightly rustfmt change reported for the PR-owned SQL string in blob_cleanup.rs. Stable cargo fmt --all -- --check and affine_server_native clippy with -D warnings pass; unrelated repository baseline formatting differences were left untouched.

keepClamDown:
@donqu1xotevincent Addressed all six requested changes in 8e62ab3211 and resolved the corresponding review threads.\n\nHighlights:\n- moved attachment admission and the initial blob write ahead of receipt/document persistence, with an A-fails/B-succeeds regression;\n- replaced unbounded image loadItem materialization with in-place/file representations and added real NSItemProvider coverage;\n- changed parser v2 to one snapshot decode and one ref traversal, including a 512-ref regression;\n- put both destructive cleanup scenarios on the required Rust CI path with an isolated PostgreSQL service;\n- committed generated config artifacts and applied the PR-owned nightly rustfmt diff.\n\nLocal verification: TypeScript typecheck, oxlint/oxfmt, 53 share-import tests, 79 ShareInboxSafetyTests, affine_server_native clippy -D warnings, doc_blob_refs tests, release integration-test compilation, and a clean post-commit generation rerun. Please take another look when convenient.

keepClamDown:
@donqu1xotevincent Follow-up on CI: the new head `8e62ab32119eee4581c32fc529e60c4084e3bbd0` has no executed failing job yet. Both `Build & Test` and `PR Title Lint` ended immediately as `action_required` with zero jobs, which indicates the fork workflows are waiting for maintainer approval. Could you approve the workflow run and re-review the resolved six P2 threads? I will monitor the resulting jobs and address any PR-owned failures.

darkskygit (approved):

PR #15577, opened 2026-09-08
fix(nbstore): prevent re-entrant realtime resubscriptions

### LINKED ISSUE #15576: [Bug]: Re-entrant realtime resubscribe can fan out subscriptions
## What happened?

A self-hosted AFFiNE server became unresponsive after one native desktop client reconnected. The server received approximately 99,000 `realtime:subscribe` requests and then exhausted its JavaScript heap. Other clients (including an MCP integration) subsequently timed out only because the backend was already unavailable; they were not the trigger.

The re-entrant resubscribe path in the current `canary` source appears sufficient to explain the fan-out:

1. `connect()` calls `resubscribeAll()` while `subscriptionsNeedResubscribe` is `true`.
2. `resubscribeAll()` emits `{ type: 'ready' }` for each re-established subscription.
3. A live-query consumer reacts to `ready` by calling `request()`, which calls `connect()` again.
4. The flag remains `true` until the outer loop finishes, so the nested `connect()` starts another `resubscribeAll()` over a map already being changed by the first loop.

Relevant current source:

- [`RealtimeManager.connect`](https://github.com/toeverything/AFFiNE/blob/canary/packages/common/nbstore/src/realtime/manager.ts#L235-L267)
- [`RealtimeManager.resubscribeAll`](https://github.com/toeverything/AFFiNE/blob/canary/packages/common/nbstore/src/realtime/manager.ts#L284-L315)
- [`RealtimeLiveQuery` revalidation on `ready`](https://github.com/toeverything/AFFiNE/blob/canary/packages/common/nbstore/src/realtime/live-query.ts)

I reproduced the mechanism in isolation from the exact `manager.ts` source checked on 2026-09-09. With 20 active subscriptions and a `ready` handler that triggers one request, the current ordering issued 3,840 subscribe calls before the harness stopped after 500 successful acknowledgements. Clearing the dirty flag before enumerating subscriptions reduced that harness to 20 subscribe calls. This is evidence of the re-entrancy mechanism, not a claim that it alone explains every server OOM.

### Expected behavior

One reconnect should issue at most one resubscription per active subscription. A `ready` callback that asks for data must not start a nested resubscription pass.

### Suggested direction

Make resubscription single-flight and mark the subscriptions as no longer pending before invoking any callback that can re-enter `connect()`. If a disconnect happens during the pass, mark it dirty again and let the next connection perform one additional pass. A regression test should cover a `ready` callback that calls `request()` while multiple subscriptions are being restored.

## Distribution version

macOS ARM 64 (Apple Silicon)

## App Version

Native client: `0.27.4`

## Are you self-hosting?

Yes.

## Self-hosting Version

The incident was observed with server `0.27.3`. I then upgraded the server to `0.27.4`; the reproduced source-level re-entrancy is still present in `canary` as of 2026-09-09. I cannot isolate whether the production recovery came from the release upgrade, the desktop restart, or both.

## Relevant log output

```text
single external native client reconnect
realtime:subscribe messages observed: approximately 99,000
affine-app outcome: JavaScript heap exhaustion and backend unresponsiveness
```

No user content, document identifiers, hosts, IP addresses, or credentials are included here.

## Anything else?

The server was stable while that client was disconnected. A narrow temporary block of the client's Socket.IO traffic prevented recurrence during diagnosis. The native client successfully synchronized again after the server update and a full desktop restart, so the exact production trigger needs maintainer investigation.

This report was prepared with AI assistance, but the source locations, measurements, and reproduction result were manually checked. I am intentionally not asserting the issue-template checkbox that says the submission was not AI-generated.

vadzhipov:
Fixes #15576.

## Summary

Prevent a `ready` callback from starting a nested realtime resubscription pass while a reconnect is already restoring subscriptions.

## Changes

- Mark subscriptions as no longer pending before emitting a re-established subscription's `ready` event.
- Keep one in-flight `resubscribeAll()` promise and return it to concurrent callers.
- Add a regression where two `ready` callbacks re-enter `request()` during a context-switch resubscription. The test expects only the two initial and two restored subscriptions.

## Validation

- `git diff --check`
- Isolated harness built from the exact `canary` `manager.ts`: the unpatched source produced more than the two expected ready-triggered revalidation requests; the patch completes with exactly four subscribe calls for two subscriptions.

The new Vitest regression is included for repository CI. I could not run the full AFFiNE Yarn workspace locally because this checkout is intentionally sparse and does not contain the complete dependency graph.

## Compatibility

No protocol or public API changes. Concurrent reconnect/request callers now share the same resubscription pass rather than issuing duplicate subscriptions.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved realtime subscription recovery when reconnects occur during an active recovery process.
  - Ensured subscriptions are restored after interrupted or overlapping recovery attempts.
  - Prevented unnecessary duplicate subscription requests during concurrent connection-ready events.
  - Improved reliability of live updates after connection interruptions.

- **Tests**
  - Added coverage for reconnects during subscription recovery.
  - Verified that follow-up recovery restores all subscriptions correctly.


packages/common/nbstore/src/realtime/manager.ts
```diff
@@ -276,42 +277,64 @@ export class RealtimeManager {
...
+    }
+    this.subscriptionsNeedResubscribe = true;
     this.resubscribeAll().catch(error => {
       this.lastError = normalizeError(error);
     });
   };
 
-  private async resubscribeAll() {
+  private resubscribeAll(): Promise<void> {
+    if (this.resubscribePromise) {
```
darkskygit, line 290: `resubscribePromise` can belong to the old socket after a reset. Scope it to the current connection generation and invalidate it on reset.

packages/common/nbstore/src/realtime/manager.ts
```diff
@@ -276,42 +277,64 @@ export class RealtimeManager {
...
-      }
-    }
-    this.subscriptionsNeedResubscribe = false;
+            this.subscriptions.delete(subscriptionId);
+            this.subscriptions.set(ack.data.subscriptionId, subscription);
+            subscription.onResubscribed(ack.data.subscriptionId);
+            subscription.subject$.next({
+              type: 'ready',
+            });
+          } catch (error) {
```
darkskygit, line 323: Socket.IO rejects pending acks on disconnect. This catch deletes the subscription, so the next pass cannot restore it. Keep transport failures pending.

packages/common/nbstore/src/realtime/__tests__/manager.spec.ts
```diff
@@ -383,6 +406,104 @@ test('subscribe registers server room again after reconnect', async () => {
...
+    authenticated: true,
+  });
+  const reconnectRequest = manager.request('notification.count.get', {});
+
+  await vi.waitFor(() =>
+    expect(
+      socket.emitted.filter(item => item.event === 'realtime:subscribe')
+    ).toHaveLength(4)
+  );
+  socket.connected = false;
```
darkskygit, line 491: This test still passes without the single-flight code. Reject the pending ack on disconnect and assert that only one pass runs.

